PR #2, opened 2024-04-04
Out 238 Design Kanban board for tasks

aatbip:
Ref: [OUT-238](https://linear.app/copilotplatforms/issue/OUT-238/design-kanban-board-for-tasks)

- Adds a advanced Droppable component for handling the drag and drop functionalities all over the app
- Adds a kanban board for tasks


src/app/canvas/Canvas.tsx
```diff
@@ -0,0 +1,128 @@
+'use client';
+
+import { useState, useCallback } from 'react';
+import update from 'immutability-helper';
+import { Stack } from '@mui/material';
+import { TaskCard } from '@/components/cards/TaskCard';
+import { TaskColumn } from '@/components/cards/TaskColumn';
+import { Droppable } from '@/hoc/Droppable';
+
+interface Task {
```
makeitraina, line 10: I recommend moving the type and mock data to a separate file. That file will be replaced by the logic that loads this data and that is also where the type would be defined

src/app/canvas/Canvas.tsx
```diff
@@ -0,0 +1,128 @@
...
+];
+
+const mockCompletedTask: Task[] = [
+  {
+    assignedTo: 'rock',
+    id: 7,
+  },
+];
+
+export const Canvas = () => {
```
makeitraina, line 56: why is this component called a Canvas? Is there a reason its named generically or is there more descriptive name we can give it.

src/app/canvas/Canvas.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [lists, setLists] = useState<Lists>({
+    1: mockTaskTodoData,
+    2: mockTaskInProgressData,
+    3: mockCompletedTask,
+  });
+
+  const moveCard = useCallback(
```
makeitraina, line 63: this a little confusing, what are we trying to do here exactly?

src/app/canvas/Canvas.tsx
```diff
@@ -0,0 +1,128 @@
...
+        <Droppable
+          key={listId}
+          accept={'taskCard'}
+          index={index}
+          id={Number(listId)}
+          onDropItem={(item) => onDropItem(Number(listId), item)}
+        >
+          <TaskColumn key={listId} columnName={`List ${listId}`} taskCount={String(listData.length)}>
+            {listData.map((task: Task, index: number) => (
+              <Droppable key={task.id} accept={'taskCard'} index={index} id={Number(listId)} moveCard={moveCard}>
```
makeitraina, line 119: we use the Droppable HOC here, but the TaskCard is actually a "Draggable" right? Its not that we drop something into TaskCard but instead we drag the TaskCard into a Droppable area/

Let me know if Im misunderstood the implementation, would be great to clarify a bit

makeitraina (commented):
This is a good start. Some questions about the naming of components and setup, but I like where this is going.

src/app/canvas/Canvas.tsx
```diff
@@ -0,0 +1,128 @@
...
+    3: mockCompletedTask,
+  });
+
+  const moveCard = useCallback(
+    (dragIndex: number, hoverIndex: number, sourceId: number, targetId: number) => {
+      const sourceList = lists[sourceId];
+      const targetList = lists[targetId];
+
+      if (sourceId === targetId) {
+        const updatedList = update(sourceList, {
```
makeitraina, line 69: what is this doing? is it rearrange within the column?

rrojan (changes requested):
Please fix merge conflicts. Please also pull the latest `main` and run `yarn prettier:fix` to use our latest prettier config to format files :raised_hands:

src/app/canvas/Canvas.tsx
aatbip: Sure. It is only for now I have kept all the mock data and type defs on the same page. The type defs will depend on the API response and while working on the API integration I should remove the mock data and add new type defs that should be compatible with the response from the API.

src/app/canvas/Canvas.tsx
aatbip: Canvas was just a name given to the component. Renamed the component and filename from `Canvas` to `TaskBoard` which sounds more descriptive IMO.

src/app/canvas/Canvas.tsx
aatbip: `sourceId` is the ID of the list from where the card is being moved (origin) and `targetId` is the ID of the list where the task is being moved (destination). It is checking if the task is getting moved in the same list or the another list.

rrojan (approved):

PR #4, opened 2024-04-08
OUT-262 | Create Database Migrations

rrojan:
### Ref:
- [OUT-262](https://linear.app/copilotplatforms/issue/OUT-262/create-database-migrations)

### Changes
- Add initial schema for Tasks app

### Testing
- Manually added records in new schema. Also for `OUT-240` which branches off this PR, objects are being created for these Prisma models.


prisma/migrations/20240408161437_init/migration.sql
```diff
@@ -0,0 +1,58 @@
...
+    "type" "StatusType" NOT NULL,
+    "name" TEXT NOT NULL,
+    "key" TEXT NOT NULL,
+    "color" TEXT,
+
+    CONSTRAINT "StatusSettings_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "Tasks" (
```
saugatmaharjan, line 25: @rrojan are we going to  use `PascalCase` for table name? CC: @sazanrjb .

if yes, then it should we make the column names `PascalCase` as well?

rrojan: @saugatmaharjan  The table names were suggested by the Copilot Team on their PRD [here](https://copilotplatforms.notion.site/Tasks-App-4e19e810bf0c4c8b96e9d583dc42c400#f10a8acfb3144225b5cbd93f7fa73a3c)

makeitraina: @rrojan the names suggested in the PRD are not table names, those were the resource names. You can use a different pattern for naming tables

prisma/migrations/20240408161437_init/migration.sql
```diff
@@ -0,0 +1,58 @@
...
+    "title" TEXT NOT NULL,
+    "body" TEXT,
+    "statusSettingId" UUID NOT NULL,
+    "createdBy" TEXT NOT NULL,
+
+    CONSTRAINT "Tasks_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "ViewSettings" (
```
makeitraina, line 38: you can leave out all of this stuff for ViewSettings for now. Let's keep our implementation simple and just focus on the basic functionality needed for m1.

prisma/schema.prisma
```diff
@@ -0,0 +1,51 @@
...
+enum StateType {
+  backlog
+  unstarted
+  started
+  completed
+  cancelled
+}
+
+enum AssigneeType {
+  iu
```
makeitraina, line 20: can we make this more explicit for `internalUser`. `iu` is just an abbreviation we use to make it less redundant but the code should be more clear

makeitraina (approved):
looks good

PR #9, opened 2024-04-12
OUT-240 | OUT-251 | OUT-257 | OUT-246 | Backend Architecture + Tasks & Status CRUD API

rrojan:
## Ref:
- [OUT-240](https://linear.app/copilotplatforms/issue/OUT-240)
- [OUT-251](https://linear.app/copilotplatforms/issue/OUT-251)
- [OUT-257](https://linear.app/copilotplatforms/issue/OUT-257)

## NOTES:
- Merge after Database Migrations PR https://github.com/copilot-platforms/tasks-app/pull/4
- [Link to postman collection](https://app.getpostman.com/join-team?invite_code=ae8a50274dffacdbf44e93c6a4b6f2dc&target_code=a0d6c2a2b4b9b94add3b90d8810783e0)

## Testing
- Tested all routes using IU token
- Tested all routes using client token


src/app/api/core/services/policies.service.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export class PoliciesService extends BaseService {
+  private defaultPolicies: Record<UserRole.Client, Record<Resource, UserAction[]>> = {
+    [UserRole.Client]: {
+      [Resource.Tasks]: ['read'],
+      [Resource.Status]: ['read'],
+    },
+  }
+
+  authorize(action: UserAction, resource: Resource): boolean | void {
```
makeitraina, line 18: I found this to be a little confusing. Can you explain what this is supposed to be doing and the scenarios? I think a code comment here would be really helpful

src/app/api/tasks/tasks.service.ts
```diff
@@ -0,0 +1,96 @@
...
+          ],
+        },
+      }
+    }
+
+    return filters
+  }
+  async getAllTasks() {
+    const user = this.user
+    new PoliciesService(user).authorize('read', Resource.Tasks)
```
makeitraina, line 42: what is this line doing?

src/app/api/tasks/tasks.service.ts
```diff
@@ -0,0 +1,96 @@
...
+type FilterByAssigneeId = {
+  assigneeId: string
+  assigneeType: AssigneeType
+}
+
+export class TasksService extends BaseService {
+  private buildReadFilters() {
+    const user = this.user
+
+    let filters = { where: { workspaceId: user.workspaceId, OR: undefined as unknown as FilterByAssigneeId[] } }
```
makeitraina, line 16: what is this `undefined as unknown as FilterByAssigneeId[]`? trying to do?

src/app/api/tasks/tasks.service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    if (user.clientId) {
+      filters = {
+        where: {
+          ...filters.where,
+          OR: [{ assigneeId: user.clientId as string, assigneeType: 'client' }],
```
makeitraina, line 22: is this saying `assignedID = clientID OR assigneeType = client`? Because I don't think that is correct.

makeitraina (commented):
I have a few questions would love to understand the approach.

Separately, I think this PR is trying to do too much. Its dense and hard to review because it covers so many flows. mostly, I call this out because it will take more time to review this PR in its current state.

Also, what testing have we done for this PR? I strongly suggest that every PR description should have a testing section that explains what was tested and a video to go along with it.

src/app/api/tasks/tasks.service.ts
rrojan: The SQL returned by this prisma query is more along the lines of this:

```sql
SELECT ...
WHERE <prev conditions> AND (
    (`assigneeId` = 'clientId' AND `assigneeType` = 'client')
  )
```

When OR  has multiple values inside the array, it performs an OR - but if it only has a single value in its filter array it acts like a normal `AND` statement, without `OR` actually being involved.

The reason I wrote it in a slightly more non-conventional way is that I could override it later in `if (user.clientId && user.companyId)`, it seemed cleaner and shorter to overwrite the filter this way.

src/app/api/tasks/tasks.service.ts
rrojan: @makeitraina I'm going with a resource-level authorization approach. 

Since all controllers use the service layer to access resources, each service method has a line on top using `PoliciesService#authorize` to check if user can perform the need action (read, write, update, etc) on the resource this particular service is associated with.

src/app/api/tasks/tasks.service.ts
rrojan: I wanted an `undefined` `OR` value for `where` so I could later inject it with `OR` condition if required. If `OR` condition is not required, the `undefined` OR value would cause no `OR`s to be generated in the final SQL query.

I admit it involves a bit of type gymnastics but i found it to be the shortest way to do this. I'm open to new solutions if required!

rrojan: I've added some method documentation here so we don't forget how this works in the future!

src/app/api/core/exceptions/api.ts
```diff
@@ -0,0 +1,12 @@
+class APIError extends Error {
+  status: number
+  message: string
+
+  constructor(status: number, message?: string) {
+    super(message)
```
sazanrjb, line 6: we can pass fallback message here as well?

src/app/api/core/models/User.model.ts
```diff
@@ -0,0 +1,24 @@
+import { UserAction, UserRole } from '@api/core/types/user'
```
sazanrjb, line 1: UserAction is unused

src/app/api/core/models/User.model.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Token } from '@/types/common'
+import { PoliciesService } from '@api/core/services/policies.service'
+import { Resource } from '@api/core/types/api'
```
sazanrjb, line 4:
```suggestion
```

src/app/api/core/models/User.model.ts
```diff
@@ -0,0 +1,24 @@
...
+
+class User implements Token {
```
sazanrjb, line 6:
```suggestion
class User {
```

src/app/api/core/services/policies.service.ts
```diff
@@ -0,0 +1,30 @@
...
+import { BaseService } from './base.service'
+import APIError from '@api/core/exceptions/api'
+import httpStatus from 'http-status'
+
+/**
+ * Resource-level policies control service for Users (IU + Client)
+ */
+export class PoliciesService extends BaseService {
+  private defaultPolicies: Record<UserRole.Client, Record<Resource, UserAction[]>> = {
+    [UserRole.Client]: {
```
sazanrjb, line 12: Can we check policies based on the resources and check role later on? that we if we need to check policies for internal users, then we wont have to add keys.

And also just checking actions (read, create) is not scalable. 

I think a bit scalable solution is:
- Create policies on resource label
- Create a Policy file for each resource
- Create methods with the name of actions required, in the Policy file

Eg:

```
// policies.ts
export const defaultPolicies: Record<UserRole.Client, Record<Resource, UserAction[]>> = {
      [Resource.Tasks]: TaskPolicy,
      [Resource.WorkflowState]: WorkflowStatePolicy,
  }
```

```
// PolicyGate.ts
export class PolicyGate {

  // ..

  authorize(action: PolicyAction, resource: Resource) {
    const policy = new defaultPolicies[resource]();
    // call method based on the action here
  }
}
```

```
// TaskPolicy.ts
export class TaskPolicy {
  read(resource: Resource, user: User) {
    if (user.role === UserRole.IU) {
      return true;
    }
    
    if (user is in the matched company OR task assignee is the client) {
     return true;
    }
    
    return false;
  }
}
```

Let me know what you think? I think this looks a bit clean and scalable

src/app/api/core/types/api.ts
```diff
@@ -0,0 +1,10 @@
+export enum Resource {
+  Tasks = 'Task',
+  WorkflowState = 'WorkflowStates',
```
sazanrjb, line 3: what is workflow state? also lets follow one convention, may be singular is better

src/app/api/core/utils/withErrorHandler.ts
```diff
@@ -0,0 +1,50 @@
+import { NextRequest, NextResponse } from 'next/server'
+import { ZodError, ZodIssue } from 'zod'
+import { CopilotApiError } from '@/types/CopilotApiError'
+import APIError from '../exceptions/api'
```
sazanrjb, line 4: absolute import

src/app/api/core/utils/withErrorHandler.ts
```diff
@@ -0,0 +1,50 @@
...
+  return async (req: NextRequest, params: any) => {
+    try {
+      return await handler(req, params)
+    } catch (error) {
+      console.error(error)
+
+      let status = 500
+      let message: string | ZodIssue[] = 'Something went wrong'
+
+      if (error instanceof ZodError) {
```
sazanrjb, line 36: zod error can be 422 as well?

src/app/api/tasks/tasks.service.ts
```diff
@@ -0,0 +1,103 @@
...
+        },
+      }
+    }
+
+    return filters
+  }
+
+  async getAllTasks() {
+    const user = this.user
+    new PoliciesService(user).authorize('read', Resource.Tasks)
```
sazanrjb, line 49: lets not hard code read. we have enum for actions?

sazanrjb: How about

```
const policyGate = new PolicyGate();
policyGate.authorize(Action.READ, Resources.Task, user);
```
?

sazanrjb (commented):
I am still reviewing

src/app/api/core/types/api.ts
rrojan: @sazanrjb I noticed this and fixed this in PR #10. Cherrypicking that change to this PR too.

src/app/api/core/utils/withErrorHandler.ts
rrojan: @sazanrjb this is just the default value for status. I have changed it to `422` in `ZodError` below now.

src/app/api/tasks/tasks.service.ts
rrojan: @sazanrjb I did that initially but it felt like 'read' felt more natural to read since you can read it like
`authorize` `read` of `resource task` for `user`


Changed it to enum again.

src/app/api/core/services/policies.service.ts
rrojan: @sazanrjb Your recommendation definitely makes sense and is cleaner but it also adds a lot of boilerplate for each resource. I think we should be good with the current approach at least for now... we can think up a solution or refactor to this when this requirement comes (if it does).

My idea was we could create a `Permissions` table and query permissions with `db.permissions.findFirst({ where: { resource })` instead of using `defaultPolicies` to authorize later on. This way we don't need individual `xxx.policy.ts` files and repeat ourselves.

src/app/api/core/utils/withErrorHandler.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @throws {APIError} Captures and handles APIError
+ */
+export const withErrorHandler = (handler: RequestHandler): RequestHandler => {
+  return async (req: NextRequest, params: any) => {
+    try {
+      return await handler(req, params)
+    } catch (error) {
+      console.error(error)
+
+      let status = 500
```
sazanrjb, line 33: let status: keyof httpStatus.HttpStatus = httpStatus.INTERNAL_SERVER_ERROR

src/app/api/core/utils/withErrorHandler.ts
```diff
@@ -0,0 +1,50 @@
...
+      let message: string | ZodIssue[] = 'Something went wrong'
+
+      if (error instanceof ZodError) {
+        status = 422
```
sazanrjb, line 37: status = httpStatus.UNPROCESSABLE_ENTITY

src/app/api/tasks/tasks.service.ts
sazanrjb: and why do we need to typecast it to assigneeId[]?

src/types/dto/tasks.dto.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from 'zod'
+
+export const AssigneeTypeSchema = z.enum(['internalUser', 'client', 'company']).optional()
+export type AssigneeType = z.infer<typeof AssigneeTypeSchema>
+
+export const CreateTaskRequestSchema = z.object({
+  assigneeId: z.string().optional(),
+  assigneeType: AssigneeTypeSchema,
+  title: z.string(),
+  body: z.string().optional(),
```
sazanrjb, line 10: is this for description?

src/types/dto/tasks.dto.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from 'zod'
+
+export const AssigneeTypeSchema = z.enum(['internalUser', 'client', 'company']).optional()
+export type AssigneeType = z.infer<typeof AssigneeTypeSchema>
+
+export const CreateTaskRequestSchema = z.object({
+  assigneeId: z.string().optional(),
```
sazanrjb, line 7: `.uuid()` ?

src/types/dto/tasks.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+  assigneeType: AssigneeTypeSchema,
+  title: z.string(),
+  body: z.string().optional(),
+  workflowStateId: z.string().uuid(),
+})
+export type CreateTaskRequest = z.infer<typeof CreateTaskRequestSchema>
+
+export const UpdateTaskRequestSchema = z.object({
+  assigneeId: z.string().optional(),
```
sazanrjb, line 16: uuid? or can it be string as well?

src/types/dto/tasks.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+export const CreateTaskRequestSchema = z.object({
+  assigneeId: z.string().optional(),
+  assigneeType: AssigneeTypeSchema,
+  title: z.string(),
+  body: z.string().optional(),
+  workflowStateId: z.string().uuid(),
+})
+export type CreateTaskRequest = z.infer<typeof CreateTaskRequestSchema>
+
+export const UpdateTaskRequestSchema = z.object({
```
sazanrjb, line 15: This looks same as Create schema just with optional, in Zod, we can achieve this using partial

CreateTaskRequestSchema.partial()

src/types/dto/workflowStates.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod'
+
+export const WorkflowStateTypeSchema = z.enum(['backlog', 'unstarted', 'started', 'completed', 'cancelled'])
```
welearnandshare, line 3: We may need the enum values later

```suggestion
export enum WorkflowState {
  BACKLOG = 'backlog',
  UNSTARTED = 'unstarted',
  STARTED = 'started',
  COMPLETED = 'completed',
  CANCELLED = 'cancelled'
}
export const WorkflowStateTypeSchema = z.nativeEnum(WorkflowState)
```

sazanrjb, line 3: We may need the enum values later

```suggestion
export enum WorkflowState {
  BACKLOG = 'backlog',
  UNSTARTED = 'unstarted',
  STARTED = 'started',
  COMPLETED = 'completed',
  CANCELLED = 'cancelled'
}
export const WorkflowStateTypeSchema = z.nativeEnum(WorkflowState)
```

makeitraina (commented):
more comments. Still not done reviewing

PR #10, opened 2024-04-15
OUT-247 | Create an API to show assignee list

rrojan:
## Ref:
- [OUT-247](https://linear.app/copilotplatforms/issue/OUT-247/create-an-api-to-show-assignee-list)

## Changes:
- [x] Add `Users` resource for API
- [x] Add GET endpoint to query grouped users fetched from Copilot API

## Testing
[Loom](https://www.loom.com/share/51ae49fe1909426991d6648cd6073eb3?sid=9c45cb82-5110-413c-8bb4-591159190af2)

## NOTES: 
- Merge after https://github.com/copilot-platforms/tasks-app/pull/12 , https://github.com/copilot-platforms/tasks-app/pull/13 and https://github.com/copilot-platforms/tasks-app/pull/14


rrojan:
@makeitraina This PR should only contain changes related to assignees list now

src/app/api/users/route.ts
```diff
@@ -0,0 +1,4 @@
+import { withErrorHandler } from '../core/utils/withErrorHandler'
```
sazanrjb, line 1: absolute import

src/app/api/users/users.service.ts
```diff
@@ -0,0 +1,30 @@
+import { CopilotAPI } from '@/utils/CopilotAPI'
+import { BaseService } from '@api/core/services/base.service'
+import User from '@api/core/models/User.model'
+import { PoliciesService } from '@api/core/services/policies.service'
+import { Resource } from '@api/core/types/api'
+import { UserAction } from '../core/types/user'
```
sazanrjb, line 6:
```suggestion
import { UserAction } from '@api/core/types/user'
```

src/app/api/users/users.service.ts
```diff
@@ -0,0 +1,30 @@
...
+    const user = this.user
+    new PoliciesService(user).authorize(UserAction.Read, Resource.Users)
+
+    const [ius, clients, companies] = await Promise.all([
+      this.copilot.getInternalUsers(),
+      this.copilot.getClients(),
+      this.copilot.getCompanies(),
+    ])
+
+    return { ius: ius.data, clients: clients.data, companies: companies.data }
```
sazanrjb, line 26: can we write a meaningful name please. ius seems confusing

src/types/common.ts
```diff
@@ -106,9 +106,23 @@ export const CustomFieldResponseSchema = z.object({
...
+  id: z.string(),
   givenName: z.string().optional(),
   familyName: z.string().optional(),
   companyId: z.string().uuid().optional(),
   customFields: z.record(z.string(), z.union([z.string(), z.array(z.string())]).nullish()).nullish(),
 })
 export type ClientRequest = z.infer<typeof ClientRequestSchema>
+
+export const InternalUsersSchema = z.object({
+  id: z.string(),
```
sazanrjb, line 118: is this uuid field?

src/types/common.ts
```diff
@@ -106,9 +106,23 @@ export const CustomFieldResponseSchema = z.object({
...
+  givenName: z.string(),
+  familyName: z.string(),
+  email: z.string(),
```
sazanrjb, line 121:
```suggestion
  email: z.string().email(),
```

src/types/common.ts
```diff
@@ -106,9 +106,23 @@ export const CustomFieldResponseSchema = z.object({
 export type CustomFieldResponse = z.infer<typeof CustomFieldResponseSchema>
 
 export const ClientRequestSchema = z.object({
+  id: z.string(),
```
sazanrjb, line 109: should this be uuid?

src/app/api/users/users.service.ts
rrojan: Changed. cc @aatbip Please accomodate this change in the API implementation PR

sazanrjb (approved):

src/app/api/users/users.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  private copilot: CopilotAPI
+
+  constructor(user: User) {
+    super(user)
+    this.copilot = new CopilotAPI(this.user.token)
+  }
+
+  async getGroupedUsers() {
+    const user = this.user
+    new PoliciesService(user).authorize(UserAction.Read, Resource.Users)
```
makeitraina, line 18: I don't see the purpose this line of code is serving. If you removed the line of code what would not work?

src/app/api/users/users.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    const [ius, clients, companies] = await Promise.all([
+      this.copilot.getInternalUsers(),
+      this.copilot.getClients(),
```
makeitraina, line 22: What if there are thousands of clients in a portal? Is this going to fetch through all pages? Who will this be handled?

makeitraina (approved):
Approving this for now, but we need to discuss scalability here.

Please create a linear ticket to handle pagination for listing clients and add that as a topic to discuss in our next sync before merging this PR

src/app/api/users/users.service.ts
rrojan: This is meant to be resource-level authorization for the service layer.

Let's say we query the `/api/users` endpoint as a client - then we can access a list of all the IUs, companies that this organization is working with, as well as all of their clients!

The PoliciesService validates that the current `user` is authorized to read (`UserAction.Read`) the users resource (`Resource.Users` in this case. @makeitraina

src/app/api/users/users.service.ts
rrojan: That is definitely an issue as we scale up. This endpoint is meant to fetch users for this view:

image

So if we're discussing pagination we should figure out how we will scale up this view dropdown when there are lots of users as well. How would we handle this in the design?

I will create a new issue in backlog for this for now

makeitraina: I think this design will still scale to filter assignees, but the search on the drop-down will have to be done through an api call rather than just filtering on the webapp. wdyt?

src/app/api/users/users.service.ts
rrojan: It's in the `policies.service.ts` file. We are using a hash lookup right now @makeitraina 

Screenshot from 2024-04-30 10-07-23

src/app/api/users/users.service.ts
rrojan: Makes sense. Currently we are doing a client-side filter for this which won't scale well.

I've made a linear issue for this, we will look over this in the future.

PR #12, opened 2024-04-18
OUT-279 | Backend architecture and setup

rrojan:
### Tasks

- [OUT-279 | Backend architecture and setup](https://linear.app/copilotplatforms/issue/OUT-279/backend-architecture-and-setup)
- [OUT-278 | Refactor Backend APIs](https://linear.app/copilotplatforms/issue/OUT-278/refactor-backend-apis)

### Changes

- [x] Model a Copilot user through the request token for cleanliness and reusability
- [x] Add `BaseService` that will act as a base / parent service for all other services that utilize current User and DB
- [x] Add skeleton for User authentication and authorization based on Copilot `getTokenPayload`
- [x] Add a custom APIError exception and hoc error boundary that work to return meaningful http responses instead of crashing the app like in our previous 3 apps

### Testing Criteria

- [x] This architecture was utilized in testing Tasks endpoint and WorkflowStates endpoint. 
- [x] LOOM: See loom videos utilizing this architecture in other PRs 

### Notes

Since this PR might seem overwhelming, here is a breakdown of everything in order:
- package.json and yarn lock: Bump down `prisma` by a patch version because the implementation for instantiating PrismaClient was throwing an obscure error.
image- 
- `api/core/exceptions` and `withErrorHandler`: We noticed that we were frequently running into app crashes because of edge cases in previous apps. These two combined should handle and provide a well-formed error response back to the client.
- `api/core/models/User.model.ts`: Models a copilot user based on their token
- `api/core/services/base`: Base service class that accesses current request user and db
- `api/core/utils/authenticate` and `api/core/services/policies` : Performs authorization and authentication for resource access.


src/app/api/core/services/policies.service.ts
```diff
@@ -0,0 +1,33 @@
+import { UserAction, UserRole } from '@api/core/types/user'
+import { Resource } from '@api/core/types/api'
+import { BaseService } from './base.service'
+import APIError from '@api/core/exceptions/api'
+import httpStatus from 'http-status'
+
+/**
+ * Resource-level policies control service for Users (IU + Client)
+ */
+export class PoliciesService extends BaseService {
```
makeitraina, line 10: i dont understand what purpose this files serves. Why is it part of this PR? what functionality does it acheive?

src/app/api/core/types/api.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Resource store made accessible by Tasks API
+ */
+export enum Resource {
```
makeitraina, line 4: is this relevant for this PR?

rrojan: We have defined resources in tables `Tasks` and `WorkflowStates` that are accessible through this API in the schema file

src/app/api/core/services/policies.service.ts
rrojan: Using a single policy gate-keeping service like this is to make sure we do role-based access control for resources in a clean and modular way.

makeitraina: Let me start by saying, I really like the approach and code itself that you've written here, in the right context this would probably get approved instantly. I feel this shouldn't be part of this PR. Here are my arguments for why:
1. this is not connected to any part of the code in these changes, so as a reviewer i have not context on this
2. the linear task has no mention of permission request system so there is no context about it why its needed an how. Instead I have to ask for clarification in code review
3. this flow is not tested as part of this PR so who knows if this is actually going to be use-case we need.

Essentially, in the context of this PR, this is just arbitrary code that was written to cover some future scenario

makeitraina (approved):

PR #20, opened 2024-04-25
OUT-313 | API - ViewSettings

rrojan:
### Tasks

- [OUT-313 | API - ViewSettings](https://linear.app/copilotplatforms/issue/OUT-313/api-viewsettings)

### Changes

- [x] Add schema for ViewSettings table
- [x] Add GET endpoint to fetch latest ViewSettings preferences for IU 
- [x] Add PATCH endpoint to update ViewSettings preferences
- [x] Add DTO for typesafe fetching from frontend 

### Testing Criteria

- [x] Tested GET endpoint for `/api/view-settings`
- [x] Tested PATCH endpoint for `/api/view-settings`
- [x] [LOOM](https://www.loom.com/share/129cae3e135f45a082670e478c232620?sid=082a10b2-0210-474d-b881-d3035555d0cf)


sazanrjb (approved):

src/types/dto/viewSettings.dto.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from 'zod'
+
+export const CreateViewSettingsSchema = z.object({
+  viewMode: z.enum(['board', 'list']).optional(),
```
sazanrjb, line 4: we have enum for ViewMode

src/app/api/view-settings/viewSettings.controller.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const updateViewSetting = async (req: NextRequest) => {
+  const user = await authenticate(req)
+
+  const data = CreateViewSettingsSchema.parse(await req.json())
+
+  const viewSettingsService = new ViewSettingsService(user)
+  const newViewSetting = await viewSettingsService.createOrUpdateViewSettings(data)
+
+  return NextResponse.json({ newViewSetting })
```
makeitraina, line 23: why do we pass this in using a `newViewSettings` object? IMO we can just pass in the object representing the view settings directly as the body.

src/app/api/view-settings/viewSettings.controller.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ViewSettingsService } from '@api/view-settings/viewSettings.service'
+import { CreateViewSettingsSchema } from '@/types/dto/viewSettings.dto'
+
+export const getViewSetting = async (req: NextRequest) => {
+  const user = await authenticate(req)
+
+  const viewSettingsService = new ViewSettingsService(user)
+  const viewSetting = await viewSettingsService.getViewSettingsForUser()
+
+  return NextResponse.json({ viewSetting })
```
makeitraina, line 12: why do we wrap the viewSetting in `{}`. Just pass this back directly

makeitraina (approved):
looks good, just two minor comment to consider and then you can go ahead and merge this.

src/app/api/view-settings/viewSettings.controller.ts
rrojan: This is supposed to be a response building design pattern so we could add more keys to the response if required without changing the response schema drastically @makeitraina 

E.g. I can easily accommodate changes on the API consumer if the response changed to `NextResponse.json({ viewSetting, newkey: value })` with this pattern - while I would have to do a lot more changes if I just returned `NextResponse.json(viewSetting)`

I see how it could be redundant for our use case though, so I will be removing it for now

cc @aatbip

src/app/api/view-settings/viewSettings.controller.ts
rrojan: Answered: https://github.com/copilot-platforms/tasks-app/pull/20#discussion_r1583197394

src/app/api/view-settings/viewSettings.controller.ts
sazanrjb: lets have convention of sending `data` in responses in that case

```
non paginated response: { data: {} }
paginated response: { data: {}, meta: {}, links: {} }
```

sazanrjb (approved):

PR #15, opened 2024-04-19
OUT-243 - API integration for dynamic workflow states and tasks

aatbip:
### Tasks

- [API Integration for dynamic workflow states and tasks](https://linear.app/copilotplatforms/issue/OUT-243/api-integration-for-dynamic-status-columns)

### Changes

- [x] Adds global error handling file in the root
- [x] Adds TaskResponse and WorkflowStateResponse types 
- [x] Adds HOC to handle client side state update
- [x] Adds util function to encode and decode string for URL params
- [x] Adds function in server component to fetch workflowState and tasks
- [x] Renders workflow state and tasks dynamically 
- [x] Refactors drag and drop to fit the responses coming from the API

### Testing Criteria

-  [Loom](https://www.loom.com/share/a5431c2a9a594fd38f2e0e67366a8546?sid=2e5e9fac-bf58-40a5-95a6-3bca0c5abc95)

### Notes

- This PR is dependent to #14. I have merged the latest of #14 with this PR.

PR #21, opened 2024-04-26
OUT 252 Implement Delete Task in UI

aatbip:
### Tasks

- [Implement Delete Task in UI](https://linear.app/copilotplatforms/issue/OUT-252/implement-delete-task-in-ui)

### Changes

- [x] Adds `ListBtn` and `MoreBtn` components
- [x] Adds `taskDetailsSlice` to handle states in details page
- [x] Adds `ConfirmDeleteUI` layout that can be used for delete confirmation throughout the app
- [x] Adds delete operation in the details page and integrated delete API

### Testing Criteria

- [LOOM](https://www.loom.com/share/9e72f5780d9b4ab989b687d6b98619ff?sid=43b36f71-0e70-4b78-b6a2-fa8212983313)

### Notes

- This branch is dependent on #18. 
- Merge this branch against main after #15 is merged.


src/app/detail/ui/MenuBox.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { ListBtn } from '@/components/buttons/ListBtn'
+import { MoreBtn } from '@/components/buttons/MoreBtn'
+import { Box, Popper } from '@mui/material'
+import { useState } from 'react'
+import { ClickAwayListener } from '@mui/base'
+import store from '@/redux/store'
+import { setShowConfirmDeleteModal } from '@/redux/features/taskDetailsSlice'
+
+export const MenuBox = () => {
```
makeitraina, line 11: is this supposed to be a generic component? It doesn't seem generic given that it has an explicit Delete menu item. The options should probably be passsed in as props right?

src/hooks/useHandleSelectorComponent.tsx
```diff
@@ -19,7 +19,20 @@ export const useHandleSelectorComponent = ({ item, type }: { item: unknown; type
 
     if (type === SelectorType.ASSIGNEE_SELECTOR) {
       store.dispatch(setCreateTaskFields({ targetField: 'assigneeId', value: (item as IAssigneeCombined)?.id }))
-      store.dispatch(setCreateTaskFields({ targetField: 'assigneeType', value: (item as IAssigneeCombined)?.type }))
+      const assigneeType = (item as IAssigneeCombined).type
```
makeitraina, line 22: why do we need this casting?

makeitraina (approved):
generally this looks good. The loom video with testing was also helpful.

I think there is broader question about revisting the redux architecture here, but we can have that conversation separately.

src/app/detail/ui/MenuBox.tsx
aatbip: This is not a generic component. Inside the `ui` folders of every pages, I am placing the client components with `use client` directive that are only useful for that particular page.

src/hooks/useHandleSelectorComponent.tsx
aatbip: MUI Autocomplete component expects `options` with type `unknown` and so for the `useHandleSelectorComponent` hook also I have given the type for the `item` as unknown. 
This type casting are needed to cast the type for the `unknown` `item` type.

aatbip:
Merging this branch for now. If anything is required to change, we will perform in separate PR. 
cc: @makeitraina @rrojan @sazanrjb

rrojan:
Please fix merge conflicts @aatbip

PR #23, opened 2024-04-29
OUT 299 UI & Integration - DnD for List view tasks

aatbip:
### Tasks

- [DnD for List View Tasks](https://linear.app/copilotplatforms/issue/OUT-299/ui-and-integration-dnd-for-list-view-tasks)

### Changes

- [x] Adds state in redux slice for managing toggle between list and board views
- [x] Adds view based rendering in the `TaskBoard` component
- [x] Adds DnD wrappers for the List view
- [x] Refactors the `Selector` component to use `Popper` 

### Testing Criteria

- [Loom](https://www.loom.com/share/0fdc4d64297c4bdea855ff1569047a42?sid=fcfe8227-355a-489a-8c03-cb1f4f65e915)

### Notes

- This PR is dependent on #22.


src/app/client/page.tsx
```diff
@@ -1,14 +1,13 @@
...
 import { Header } from '@/components/layouts/Header'
 
 export default function ClientPage() {
   return (
     <>
       <Header showCreateTaskButton={false} />
-      <TaskRow showConfigurableIcons={false} columnName="To do" taskCount={4}>
-        <ClientTaskCard />
-      </TaskRow>
+      {/* <TaskRow showConfigurableIcons={false} /> */}
```
rrojan, line 9: Remove commented code if not needed

src/app/page.tsx
```diff
@@ -63,9 +63,9 @@ export default async function Main({ searchParams }: { searchParams: { token: st
               'use server'
               await handleCreate(token, title, description, workflowStateId, assigneeId, assigneeType)
             }}
-            updateWorkflowStateIdOfTask={async (taskId, targetWorkflowStateId) => {
+            updateTask={async (taskId: string, payload: UpdateTaskRequest) => {
               'use server'
-              await updateWorkflowStateIdOfTask(token, taskId, targetWorkflowStateId)
+              await updateTask({ token, taskId, payload })
```
rrojan, line 68: let's organize actions related to specific pages in separate files instead of using them inline like this.

This provides a better separation of concerns between UI logic and actions

src/app/ui/TaskBoard.tsx
```diff
@@ -39,7 +40,7 @@ export const TaskBoard = ({
     store.dispatch(
       updateWorkflowStateIdByTaskId({ taskId: payload.taskId, targetWorkflowStateId: payload.targetWorkflowStateId }),
     )
-    updateWorkflowStateIdOfTask(payload.taskId, payload.targetWorkflowStateId)
+    updateTask(payload.taskId, { workflowStateId: payload.targetWorkflowStateId })
   }, [])
```
rrojan, line 44: Please fix lint warning regarding useEffect dependency

src/components/cards/ListViewTaskCard.tsx
```diff
@@ -21,21 +40,59 @@ export const ListViewTaskCard = ({ task }: { task: TaskResponse }) => {
...
+                    updateTask({
+                      payload: {
+                        assigneeType:
+                          assigneeType === 'ius'
+                            ? 'internalUser'
+                            : assigneeType === 'clients'
+                              ? 'client'
+                              : assigneeType === 'companies'
+                                ? 'company'
+                                : 'internalUser',
```
rrojan, line 73: Use an object to lookup the value for this instead of nesting multiple ternary operators

```ts
const assigneeTypes = {
  ius: 'internalUser',
  clients: 'client',
  companies: 'company'
}
```
```ts
payload: { assigneeType: assigneeTypes[assigneeType] || 'internalUser' }
```

rrojan, line 73: What happens if I unassigned a task? Will the payload have `assigneeId` as null but `assigneeType` as "internalUser"?

src/components/cards/ListViewTaskCard.tsx
```diff
@@ -2,15 +2,34 @@
 
 import { Avatar, Box, Stack, Typography } from '@mui/material'
 import { AppMargin, SizeofAppMargin } from '@/hoc/AppMargin'
-import { TaskResponse } from '@/types/dto/tasks.dto'
+import { TaskResponse, UpdateTaskRequest } from '@/types/dto/tasks.dto'
 import { useSelector } from 'react-redux'
 import { selectTaskBoard } from '@/redux/features/taskBoardSlice'
+import { SelectorType } from '../inputs/Selector'
+import { useHandleSelectorComponent } from '@/hooks/useHandleSelectorComponent'
```
rrojan, line 9: Please prefer absolute / aliased imports instead of relative

src/app/ui/TaskBoard.tsx
```diff
@@ -59,28 +60,63 @@ export const TaskBoard = ({
...
+              </DragDropHandler>
+            )
+          }
+          if (view === View.LIST_VIEW) {
+            return (
+              <DragDropHandler key={list.id} accept={'taskCard'} index={index} id={list.id} onDropItem={onDropItem}>
+                <TaskRow
+                  key={list.id}
+                  columnName={list.name}
+                  taskCount={calculateTaskCountBasedOnWorkflowStateId(list.id)}
```
rrojan, line 97: Isn't this function name a bit too long :stuck_out_tongue: 

Maybe something like `taskCountForWorkflowStateId` would be shorter

src/app/ui/TaskBoard.tsx
```diff
@@ -59,28 +60,63 @@ export const TaskBoard = ({
...
+                  })}
+                </TaskColumn>
+              </DragDropHandler>
+            )
+          }
+          if (view === View.LIST_VIEW) {
+            return (
+              <DragDropHandler key={list.id} accept={'taskCard'} index={index} id={list.id} onDropItem={onDropItem}>
+                <TaskRow
+                  key={list.id}
```
rrojan, line 95: This part is a bit hard to understand - also I think this is a reusable component across client and IU?

Please move this to separate components

src/types/interfaces.ts
```diff
@@ -3,6 +3,11 @@ export enum UserType {
   CLIENT_USER = 'cu',
 }
 
+export enum View {
+  LIST_VIEW,
+  BOARD_VIEW,
+}
+
```
rrojan, line 10: Does this task depend on the ViewSettings API? If so you can use the View enum from `@prisma/client` defined here. This way we don't implement this separately from the API and its generally more typesafe.

image

rrojan (changes requested):

src/app/client/page.tsx
aatbip: Client page is under construction. These commented code will get removed while working on the client page.

src/app/page.tsx
aatbip: Action is not being used inline here. `updateTask` action is already in a separate file. If this was what you meant to say by inline here. Otherwise, please correct me. 

For instance, `updateTask` action should be used from both `/page.tsx` to update the assignee and `/detail/page.tsx`, to update other parameters. For this reason, organizing actions related to specific pages in separate files may again repeat the same actions. 

I found that requests for same endpoint are being repeated if you observe in the `actions.ts` files, which is not necessary. So I was thinking about a new approach in organizing the actions. I think it  is better to organize the actions according to the API requests. For instance, for all the tasks related CRUD operations, keeping actions for POST, PATCH, and DELETE in a single `task-action.ts` file inside the `/action` folder.

src/app/ui/TaskBoard.tsx
aatbip: Added in bc28efa0b4fcd297a63bb056c9756da3bf55b9c7 commit

src/components/cards/ListViewTaskCard.tsx
aatbip: Updated in ff6350b62e855d120d47f943639badd245c0f85e commit.
However, there can be other multiple places where there is relative imports being used for imports from upto level 2 i.e. `..`. This should be tackled by configuring eslint and prettier in a separate PR.

src/app/ui/TaskBoard.tsx
aatbip: Updated in c2e4e869a82b95c44e00ca9f7d8441ad4edc270d commit

src/app/ui/TaskBoard.tsx
aatbip: This is not a reusable part. We can't make use of this in the client route. For ex: client route doesn't have DnD. Also, the components like `ListViewTaskCard` and `TaskRow` needs to be passed with different props to be used in the client route. 

I think abstracting this part into another component doesn't make this any better. It will increase prop drilling for no specific reason. For instance, `key`, `accept`, `index`, `id`, `onDropItem`, `onClick`, `task`, `columnName`, `taskCount`, `showConfigurableIcons`, `updateTask`, and `handleClick` should be drilled from the parent component. Here, I am just implementing the list view and board view which is part of the `TaskBoard`.

src/types/interfaces.ts
aatbip: Yeah, I am thinking about using the same type as of the API. However, since the API is not merged yet, I am unaware about the types being used. I will implement this during the API integration.

src/components/cards/ListViewTaskCard.tsx
aatbip: Updated in 8496e3d28ee8b5ef01c48848a2096fff85d747a7 commit

src/app/page.tsx
rrojan: If updateTask is already a server action, is there a reason why we are repeating 'use server' again? I'm assuming you have 'use server' at the top of your actions file already.

aatbip: It seems like 'use server' directive is necessary here otherwise I am getting error as
```
Error: Functions cannot be passed directly to Client Components unless you explicitly expose it by marking it with "use server". Or maybe you meant to call this function rather than return it.
  <... handleCreate={function handleCreate} updateTask=...>
```

rrojan (approved):
LGTM :trophy:

src/app/page.tsx
rrojan: You could try directly passing it into the funciton without the async function wrapper using three separate arguments instead of a single object

PR #25, opened 2024-04-30
Out 303 - UI - Functional Search bar

arpandhakal:
### Tasks

- [Out 303 - UI Functional Search bar](https://linear.app/copilotplatforms/issue/OUT-303/ui-functional-search-bar)

### Changes

- [x] Added FilterBar component 
- [x] Applied SearchBar component inside FilterBar component
- [x] created FilteredTasks state and updated it on taskboard slice actions
- [x] rendered FilteredTasks on taskboard 

### Testing Criteria

- [Loom](https://www.loom.com/share/9647ee1742a94226bedc7f577776ff1a)

### Notes

- Pulled recent changes from OUT - 243

PR #29, opened 2024-05-02
Out-302 UI - Filter by Assignee

arpandhakal:
### Tasks

- [UI - Filter by Assignee](https://linear.app/copilotplatforms/issue/OUT-302/ui-filter-by-assignee)

### Changes

- [x] Selector component modified : added extra placeholder and customOptions Prop.
- [x] Created UI for filter by assignee with reference to figma
- [x] Created a new reducer is taskboard slice to handle change in filter by assignee component.
- [x] Implemented filter by asignee logic with 3 cases : filter task by users, filter tasks which are unassigned and no filter.

### Testing Criteria

- [ LOOM](https://www.loom.com/share/29dc8747cd024825a280cf197e2f67ad)

### Notes

- Dependent on #25 and #15


src/components/layouts/FilterBar.tsx
```diff
@@ -18,7 +31,60 @@ export const FilterBar = () => {
...
+                        >
+                          <CrossIcon />
+                        </IconButton>
+                      </Stack>
+                    ) : (
+                      <> assignee</>
+                    )}
+                  </Stack>
+                </Typography>
+              }
```
aatbip, line 84: Create a separate component as `FilterButton` or similar to place this part of the code. This component will be responsible for the behaviour rendered by this code.

src/components/layouts/FilterBar.tsx
```diff
@@ -18,7 +31,60 @@ export const FilterBar = () => {
...
+              selectorType={SelectorType.ASSIGNEE_SELECTOR}
+              customOptions={{ id: '', name: 'No assignee', value: '' }}
+              buttonContent={
+                <Typography
+                  variant="bodySm"
+                  lineHeight="32px"
+                  fontWeight={500}
+                  fontSize="12px"
+                  sx={{ color: (theme) => theme.color.gray[600] }}
+                >
```
aatbip, line 54: Let's not use this `Typography` in a top level here. This `Typography` is rendered as `span` tag and technically we are rendering block elements inside the inline element which is not a preferred semantics. 
I think the node for `buttonContent` prop can be refactored for better readability as well. For ex: the way `<> Filter by </>` and `<> assignee </>` are rendered can be changed. Why not render both cases by breaking them down into smaller components that reside in the same file?

src/components/inputs/Selector.tsx
```diff
@@ -13,21 +13,46 @@ export enum SelectorType {
   STATUS_SELECTOR = 'statusSelector',
 }
 
+interface customOptions {
+  id: string
+  name: string
+  value?: string
+  //other custom types..
+}
```
aatbip, line 21: I think this way of extending `Selector` component to render `customOptions` will make it constraint for only one use case. Can we consider for more flexible use case? 
I think we should rely on passing `ReactNode` as `extraOptionRenderer`. If you need help with this then let's have sync up.

aatbip (changes requested):
Still checking on this. For now, this is what I am requesting to change.

src/components/inputs/Selector.tsx
```diff
@@ -13,21 +13,46 @@ export enum SelectorType {
   STATUS_SELECTOR = 'statusSelector',
 }
 
+interface customOptions {
```
saugatmaharjan, line 16: @arpandhakal let's capitalize interface names

src/components/inputs/Selector.tsx
```diff
@@ -13,21 +13,46 @@ export enum SelectorType {
...
+  value?: string
+  //other custom types..
+}
+
 interface Prop {
   getSelectedValue: (value: unknown) => void
   startIcon: ReactNode
   value: unknown
   selectorType: SelectorType.STATUS_SELECTOR | SelectorType.ASSIGNEE_SELECTOR
   options: unknown[]
```
saugatmaharjan, line 28: can't we utilize TS `Generics` here rather that using `unknown`, please try exploring `Generics`

src/components/inputs/Selector.tsx
```diff
@@ -13,21 +13,46 @@ export enum SelectorType {
...
+  customOptions,
+}: Prop) {
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
+
+  const handleClick = (event: React.MouseEvent<HTMLDivElement>) => {
+    setAnchorEl(anchorEl ? null : event.currentTarget)
+  }
+
+  const open = Boolean(anchorEl)
+  const id = open ? 'autocomplete-popper' : undefined
```
saugatmaharjan, line 51: use empty string rather than `undefined` for `html` attribute values

src/components/layouts/FilterBar.tsx
```diff
@@ -1,15 +1,28 @@
...
 import SearchBar from '@/components/searchBar'
+import Selector, { SelectorType } from '@/components/inputs/Selector'
+import { useHandleSelectorComponent } from '@/hooks/useHandleSelectorComponent'
+import { selectTaskBoard } from '@/redux/features/taskBoardSlice'
+import { useSelector } from 'react-redux'
+import { IAssigneeCombined } from '@/types/interfaces'
+import { CrossIcon, FilterByAsigneeIcon } from '@/icons'
 
-export const FilterBar = () => {
+export const FilterBar = ({}: {}) => {
```
saugatmaharjan, line 16: if there are no props expected let's remove empty object

src/redux/features/taskBoardSlice.tsx
```diff
@@ -44,6 +44,15 @@ const taskBoardSlice = createSlice({
           : state.tasks
       state.filteredTasks = filteredTasks
     },
+    setFilteredAsignee: (state, action) => {
```
saugatmaharjan, line 47: let's type the `action` as well, please refer to RTK [Typescript doc](https://redux-toolkit.js.org/usage/usage-with-typescript)

@arpandhakal @aatbip @rrojan

aatbip: @saugatmaharjan Added types to `action`s in c8792f7e4af6ff1c7630e4e7e40ec55049bb6a34 commit

cc: @arpandhakal @rrojan

src/components/inputs/Selector.tsx
```diff
@@ -37,64 +62,61 @@ export default function Selector({ getSelectedValue, startIcon, value, selectorT
...
-                <StatusSelectorRenderer props={props} option={option} />
-              )
+        <StyledAutocomplete
+          id="status-box"
+          onBlur={() => {
+            setAnchorEl(null)
+          }}
+          openOnFocus
+          autoHighlight
+          options={customOptions ? [customOptions, ...options] : options}
```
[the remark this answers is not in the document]
aatbip: `customOptions` is the optional option that is used if we want to render extra field in the Autocomplete that is not part of the `options`. I think this approach is necessary to meet the current functionality for the `Selector` component throughout this app. 
cc: @arpandhakal

src/components/inputs/Selector.tsx
aatbip: Changed in e58dca661e43a72f2a6a73a0ca916d88191512a5 commit

aatbip (dismissed):
I have pushed commits for the reviews by @saugatmaharjan . I think we can merge now after fixing the conflicts. 
cc: @arpandhakal

aatbip (approved):

PR #32, opened 2024-05-03
OUT-317 Add "Please provide a Valid Token" message instead of triggering an Error

arpandhakal:
### Tasks

- [Add "Please provide a Valid Token" message instead of triggering an Error](https://linear.app/copilotplatforms/issue/OUT-317/add-please-provide-a-valid-token-message-instead-of-triggering-an)

### Changes

- added a token error boundary 
- added invalidtoken component and invoked the component if token is empty is not valid 

### Testing Criteria

- [ LOOM](https://www.loom.com/share/3df6ef8e38e54efabfe413ed5aba6fcb)


src/components/invalidToken/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+        fontSize: 'clamp(20px, 4vw, 42px)',
+        fontFamily: 'monospace',
+      }}
+    >
+      Please provide a valid token!
+    </div>
+  )
+}
+
+export default InvalidToken
```
aatbip, line 18: Can we use the next.js error component to define the UI error boundary instead of custom component? With next.js error component, it will be helpful to catch other types of error as well. [See this reference.](https://nextjs.org/docs/app/api-reference/file-conventions/error)

aatbip (changes requested):

src/app/page.tsx
```diff
@@ -1,8 +1,16 @@
 import { DndWrapper } from '@/hoc/DndWrapper'
 import { TaskBoard } from './ui/TaskBoard'
 import { Header } from '@/components/layouts/Header'
+import { z } from 'zod'
+import InvalidToken from '@/components/invalidToken'
```
rrojan, line 5: Please use PascalCase while naming component folders
```suggestion
import { z } from 'zod'
import InvalidToken from '@/components/InvalidToken'
```

rrojan (changes requested):

src/app/error.tsx
```diff
@@ -1,9 +1,29 @@
...
+'use client'
 
-export default function GlobalError({ error }: { error: Error & { digest?: string } }) {
+import React from 'react'
+
+const getErrorMessage = (error: Error & { digest: string }) => {
+  if (error.message) {
+    return error.message
+  }
+  return 'Something went wrong'
```
rrojan, line 9:
```suggestion
const getErrorMessage = (error: Error & { digest: string }) =>
  error.message ?? 'Something went wrong'
```

rrojan: Can we also make sure this won't expose sensitive / technical information to this component?

This block catches all errors and displays their error message. All `Error` objects have a message so `if (message) -> "XYZ" || "Something went wrong"` really doesn't make sense. A non-technical user seeing something like "Fetch failed" or a long cryptic error message for a state update error would not be preferable - instead we would like to show "Something went wrong" in those cases.

rrojan (changes requested):

foleyatwork (approved):

aatbip (approved):
lgtm

rrojan (changes requested):
LGTM! :trophy:

rrojan (approved):
LGTM! :trophy:

aatbip (changes requested):
I have a few concerns regarding this implementation. 
1. The `error.tsx` file doesn't seem to be handling the error for the empty token case. In production, next js strips out the actual error thrown and includes only a generic message but only for the errors thrown from the server component. We can still make use of the next js error boundary to show errors thrown from the server component, using a intermediate client component. 
2. The `InvalidToken` component seems not so generic. Why don't we make this component as an intermediate component which is used to forward the error thrown from anywhere (including both server and client components) to the next js error boundary file?

foleyatwork (approved):

aatbip (approved):
lgtm

PR #34, opened 2024-05-06
Out-300 Support tasks filtering

arpandhakal:
### Tasks

- [Support tasks filtering](https://linear.app/copilotplatforms/issue/OUT-300/support-tasks-filtering)

### Changes

- [x] created FilterButton group and fillter buttons ui
- [x] integrated filter functionalities
- [x] passed a function to get payload of token in the filter component 
- [x] created new reducer in taskboardslice to handle filtering logic

### Testing Criteria

- [LOOM](https://www.loom.com/share/0d7c8cb32b38449aa29f9101e591e728)


yarn.lock
```diff
@@ -1394,6 +1394,11 @@
     react-is "^18.2.0"
 
 "@mui/x-date-pickers@^7.1.1":
+<<<<<<< HEAD
+  version "7.3.1"
+  resolved "https://registry.yarnpkg.com/@mui/x-date-pickers/-/x-date-pickers-7.3.1.tgz#7be42255d3b83afe894df2cf9782ccafd147db4e"
+  integrity sha512-ZIZC+/L5ch3+J6EWRWd/rzz1z/9KPZli8sO1g7t30uFgyjpRuMK4GuazRMqDDFq1KIUtk31qJV0sKKvUsVrKRw==
+=======
 <<<<<<< HEAD
```
rrojan, line 1402: I think you committed merge conflict markers as well :stuck_out_tongue:

yarn.lock
```diff
@@ -2219,6 +2225,11 @@ camelcase@^6.2.0, camelcase@^6.3.0:
   integrity sha512-Gmy6FhYlCY7uOElZUSbxo2UCDH8owEk996gkbrpsgGtrJLM3J7jGxl9Ic7Qwwj4ivOE5AWZWRMecDdF7hqGjFA==
 
 caniuse-lite@^1.0.30001579, caniuse-lite@^1.0.30001587, caniuse-lite@^1.0.30001599:
+<<<<<<< HEAD
+  version "1.0.30001614"
+  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001614.tgz#f894b4209376a0bf923d67d9c361d96b1dfebe39"
+  integrity sha512-jmZQ1VpmlRwHgdP1/uiKzgiAuGOfLEJsYFP4+GBou/QQ4U6IOJCB4NP1c+1p9RGLpwObcT94jA5/uO+F1vBbog==
+=======
 <<<<<<< HEAD
```
rrojan, line 2233: Please resolve merge conflict and remove these merge markers

yarn.lock
```diff
@@ -5493,3 +5516,4 @@ zod@^3.22.4:
   resolved "https://registry.yarnpkg.com/zod/-/zod-3.23.6.tgz#c08a977e2255dab1fdba933651584a05fcbf19e1"
   integrity sha512-RTHJlZhsRbuA8Hmp/iNL7jnfc4nZishjsanDAfEY1QpDQZCahUp3xDzl+zfweE9BklxMUcgBgS1b7Lvie/ZVwA==
 >>>>>>> main
+>>>>>>> 808fe12a4ae3a2040d6c555211b2b0e822c61171
```
rrojan, line 5519: Same

src/components/buttonsGroup/FilterButtonsGroup.tsx
```diff
@@ -0,0 +1,54 @@
...
+                sx={{
+                  color: (theme) => (index === activeButtonIndex ? theme.color.gray[600] : theme.color.gray[500]),
+                }}
+              >
+                {item.name}
+              </Typography>
+            }
+            handleClick={() => item.onClick(index)}
+            noBackground
+            enableBackground={index === activeButtonIndex ? true : false}
```
rrojan, line 46:
```suggestion
            enableBackground={index === activeButtonIndex}
```

src/components/layouts/FilterBar.tsx
```diff
@@ -14,19 +19,57 @@ import { IAssigneeCombined, View } from '@/types/interfaces'
...
+export const FilterBar = ({
+  updateViewModeSetting,
+  getTokenPayload,
+}: {
+  updateViewModeSetting: (mode: View) => void
+  getTokenPayload: () => Promise<Token>
+}) => {
   const [searchText, setSearchText] = useState('')
-
+  const [activeButtonIndex, setActiveButtonIndex] = useState<number>()
```
rrojan, line 33: Wouldn't it make more sense to add the default active index here? By default it should be the one for My Tasks

src/app/page.tsx
```diff
@@ -82,6 +90,10 @@ export default async function Main({ searchParams }: { searchParams: { token: st
               'use server'
               await updateViewModeSettings(token, mode)
             }}
+            getTokenPayload={async (): Promise<Token> => {
+              'use server'
+              return getTokenPayload(token)
+            }}
```
rrojan, line 96: We don't need to use a server action like this here. Page should already be a server component. Why don't we use `getTokenPayload` directly in the server component (outside the return) and use its values as props instead?

src/components/layouts/FilterBar.tsx
```diff
@@ -14,19 +19,57 @@ import { IAssigneeCombined, View } from '@/types/interfaces'
...
     item: assignee[0],
     type: SelectorType.ASSIGNEE_SELECTOR,
   })
 
+  const filterButtons = [
+    {
+      name: 'My tasks',
+      onClick: async (index: number) => {
+        const payload = await getTokenPayload()
+        store.dispatch(setFilteredAsignee({ id: payload.internalUserId }))
```
rrojan, line 47: It would be much cleaner and easier to understand if we passed the internalUserId as prop to this component directly.

Use `getTokenPayload` in the Page server component and pass `internalUserId` as prop instead.

src/redux/features/taskBoardSlice.tsx
```diff
@@ -56,6 +56,13 @@ const taskBoardSlice = createSlice({
           : state.tasks.filter((task) => !task.assigneeId)
       state.filteredTasks = filteredTasks
     },
+    setFilteredTaskByType: (state, action) => {
+      const assigneeType: string[] = action.payload?.assigneeType
```
rrojan, line 60: The type for assignee type should be made stricter using `AssigneeType` from `@prisma/client`

rrojan: `(AssigneeType | 'all')[]` in your case *

src/components/buttons/SecondaryBtn.tsx
```diff
@@ -8,22 +8,24 @@ export const SecondaryBtn = ({
...
       startIcon={startIcon ? startIcon : null}
       sx={(theme) => ({
         textTransform: 'none',
-        border: enableBackground ? 'none' : `1px solid ${theme.color.borders.border}`,
+        border: enableBackground || noBackground ? 'none' : `1px solid ${theme.color.borders.border}`,
         bgcolor: enableBackground ? theme.color.gray[150] : '',
         '&:hover': {
-          border: enableBackground ? 'none' : `1px solid ${theme.color.borders.border}`,
+          border: enableBackground || noBackground ? 'none' : `1px solid ${theme.color.borders.border}`,
           bgcolor: theme.color.gray[150],
```
aatbip, line 29: This comparison is bit hard to understand because if I don't want background then my props would be `enableBackground={false}` and `noBackground={true}`, which is confusing to implement the behaviour needed.
Can we change the `noBackground` to `outlined`?

src/components/buttonsGroup/FilterButtonsGroup.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <Box
+      sx={(theme) => ({
+        border: `1px solid ${theme.color.borders.border}`,
+        borderRadius: 1,
+        display: 'flex',
+        gap: '12px',
+        alignItems: 'center',
+        padding: '4px',
+      })}
+    >
```
aatbip, line 27: Let's use `<Stack/>` if we want flex behaviour.

src/components/layouts/FilterBar.tsx
aatbip: Lets store the `payload` in the redux store rather than passing as a prop to the component it could be required to use the `payload` in other places in the future.

aatbip (changes requested):

src/components/layouts/FilterBar.tsx
arpandhakal: Wouldn't a default for all task be more relevant?

src/redux/features/authDetailsSlice.ts
```diff
@@ -0,0 +1,32 @@
+import { Token } from '@/types/common'
+import { createSlice } from '@reduxjs/toolkit'
+import { RootState } from '@/redux/store'
+
+interface IInitialState {
+  tokenPayload: Token | {}
```
aatbip, line 6: Instead of `{}`, let's type this as `undefined` for undefined cases. The return type of `getTokenPayload()` method is `Token | null`, so you might want to consider this as well. 
`undefined` type is more suitable than `{}` for the following reasons:
- Semantic Clarity: By using undefined, you explicitly communicate that the tokenPayload property may not have a value initially. This aligns better with the intent of the code and makes it clearer for other developers who might work with the code in the future.
- Type Safety: Using undefined allows TypeScript to enforce stricter type checks. If you attempt to access properties or methods on tokenPayload without first checking if it's defined, TypeScript will warn you, helping to catch potential bugs at compile-time.
- Avoids Ambiguity: When you use {}, it could imply an empty object ({}), which might lead to confusion regarding whether the property is expected to be an object with properties or just an empty placeholder.

src/redux/features/authDetailsSlice.ts
```diff
@@ -0,0 +1,32 @@
...
+  internalUserId: string
+}
+
+const initialState: IInitialState = {
+  tokenPayload: {},
```
aatbip, line 11: `tokenPayload: undefined` initially.

src/hoc/ClientSideStateUpdate.tsx
```diff
@@ -49,6 +53,10 @@ export const ClientSideStateUpdate = ({
     if (viewSettings) {
       store.dispatch(setViewSettings(viewSettings))
     }
+    if (tokenPayload) {
+      store.dispatch(setTokenPayload(tokenPayload))
+      tokenPayload.internalUserId && store.dispatch(setInternalUserId(tokenPayload.internalUserId))
```
aatbip, line 58: Since we are already setting the `tokenPayload` in the store, why is it necessary to set `internalUserId` separately? I think the `tokenPayload` itself would be enough.

src/redux/features/authDetailsSlice.ts
```diff
@@ -0,0 +1,32 @@
...
+  internalUserId: '',
```
aatbip, line 12: Having `internalUserId` separately is not necessary as we already have `tokenPayload` in this store which consists of the `internalUserId`.

src/redux/features/authDetailsSlice.ts
```diff
@@ -0,0 +1,32 @@
...
+const authDetailsSlice = createSlice({
+  name: 'authDetails',
+  initialState,
+  reducers: {
+    setTokenPayload: (state, action) => {
+      state.tokenPayload = action.payload
+    },
+    setInternalUserId: (state, action) => {
+      state.internalUserId = action.payload
+    },
```
aatbip, line 24: Please give types to the `action` as we have started typing the actions now after suggestion from SM dai. You can have take reference from other slices.

aatbip (changes requested):

src/redux/features/authDetailsSlice.ts
aatbip: Let's not make copies of same data more often because we have to update both of these to maintain consistency if we need updated `tokenPayload` again. Someone may update `tokenPayload` and forget to update `internalUserId`.

src/components/buttonsGroup/FilterButtonsGroup.tsx
```diff
@@ -0,0 +1,55 @@
...
+        gap: '12px',
+        alignItems: 'center',
+        padding: '4px',
+      })}
+      direction={'row'}
+    >
+      {filterButtons.map((item, index) => {
+        return (
+          <SecondaryBtn
+            key={index}
```
foleyatwork, line 32: Is there something other than `index` we can use for a key? Using indexes as keys is typically not recommended.

foleyatwork (approved):

aatbip (approved):
lgtm

src/app/page.tsx
```diff
@@ -59,11 +67,12 @@ export default async function Main({ searchParams }: { searchParams: { token: st
     throw new Error('Please pass the token!')
   }
 
-  const [workflowStates, tasks, assignee, viewSettings] = await Promise.all([
+  const [workflowStates, tasks, assignee, viewSettings, tokenPayload] = await Promise.all([
     await getAllWorkflowStates(token),
     await getAllTasks(token),
     addTypeToAssignee(await getAssigneeList(token)),
     await getViewSettings(token),
+    await getTokenPayload(token),
```
saugatmaharjan, line 75: I think we don't need `await` here @arpandhakal

rrojan:
I think the comments haven't been resolved yet right @arpandhakal

src/app/page.tsx
```diff
@@ -33,6 +35,12 @@ async function getAllTasks(token: string): Promise<TaskResponse[]> {
   return data.tasks
 }
 
+async function getTokenPayload(token: string): Promise<Token> {
+  const copilotClient = new CopilotAPI(token)
+  const payload = TokenSchema.safeParse(await copilotClient.getTokenPayload())
+  return payload.data as Token
```
saugatmaharjan, line 41: don't we need to handle failures? like invalid token and safeParse fails?

src/components/layouts/FilterBar.tsx
```diff
@@ -4,29 +4,77 @@ import { Avatar, Box, IconButton, Stack, Typography } from '@mui/material'
...
     type: SelectorType.ASSIGNEE_SELECTOR,
   })
 
+  useFilter(filterOptions)
+
+  const filterButtons = [
+    {
+      name: 'My tasks',
+      onClick: async (index: number) => {
+        handleFilterOptionsChange(FilterOptions.TYPE, tokenPayload?.internalUserId as string)
```
foleyatwork, line 48: This is an unfortunate typecast, it's very clear `tokenPayload?.internalUserId` _can_ be undefined. But we're telling the compiler it's definitely a string with the `as string`. I recommend doing a runtime check for the `internalUserId` rather than casting the type.

src/hoc/ClientSideStateUpdate.tsx
```diff
@@ -49,6 +53,9 @@ export const ClientSideStateUpdate = ({
     if (viewSettings) {
       store.dispatch(setViewSettings(viewSettings))
     }
+    if (tokenPayload) {
+      store.dispatch(setTokenPayload(tokenPayload))
+    }
   }, [workflowStates, tasks, token, assignee, viewSettings])
```
foleyatwork, line 59: Any reason tokenPayload is not in the dependency array?

src/hooks/useFilter.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return filteredTasks as TaskResponse[]
+}
+function filterByKeyword(filteredTasks: TaskResponse[], filterValue: string) {
+  const keyword = (filterValue as string).toLowerCase()
+  filteredTasks =
+    filterValue !== ''
+      ? filteredTasks.filter(
+          (task) => task.title?.toLowerCase().includes(keyword) || task.body?.toLowerCase().includes(keyword),
+        )
+      : filteredTasks
```
foleyatwork, line 31: This ternary's a little tough to read, can we split it out into a function or something?

foleyatwork (approved):
The code looks good. There's _a lot_ of typecasting in this PR, that's a huge source of bugs in our codebase. I recommend staying away from typecasting whenever possible, we prefer using type narrowing or type guards to make sure the runtime matches the static type analysis.

src/components/layouts/FilterBar.tsx
rrojan: @arpandhakal You can use the IUToken schema in `types/common.ts` to parse this to make sure this token payload is one for an internal user

aatbip (approved):
lgtm

rrojan (approved):
LGTM :trophy:

PR #36, opened 2024-05-10
OUT-358 | API - Soft delete records in database

rrojan:
### Tasks

- [OUT-358 | API - Soft delete records in database](https://linear.app/copilotplatforms/issue/OUT-358/api-soft-delete-records-in-database)

### Changes

- [x] Implement 3 prisma extensions for `softDelete`, `softDeleteMany` and `filterSoftDeleted`
- [x] Extend the global prisma client export with these three client extensions 
- [x] Add deletedAt column in all tables to implement softdelete

### Testing Criteria

- [x] Tested CRUD for all resources and records were found to be soft deleted
- [x] [LOOM](https://www.loom.com/share/bb9246961f7f447882f0ac86883ea539?sid=604afde8-2680-4eda-b7c5-4eb8911814b9) 

### Notes

- Prisma Client Extensions [docs](https://www.prisma.io/docs/orm/prisma-client/client-extensions)
- `$use` is used everywhere online - even Prisma doc's middlewares example uses it, but it is deprecated - use client extensions instead.

### DIFF
--- a/src/lib/prismaExtensions.ts
+++ b/src/lib/prismaExtensions.ts
@@ -1,0 +1,64 @@
+import { Prisma } from '@prisma/client'
+
+/**
+ * Prisma extension for soft delete
+ */
+export const softDelete = Prisma.defineExtension({
+  name: 'softDelete',
+  model: {
+    $allModels: {
+      async delete<M, A>(this: M, args: Prisma.Args<M, 'delete'>['where']): Promise<Prisma.Result<M, A, 'update'>> {
+        const context = Prisma.getExtensionContext(this)
+
+        return (context as any).update({
+          where: { ...args.where },
+          data: {
+            deletedAt: new Date(),
+          },
+        })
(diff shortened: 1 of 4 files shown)


src/lib/prismaExtensions.ts
```diff
@@ -0,0 +1,64 @@
...
+        })
+      },
+    },
+  },
+})
+
+/**
+ * Prisma extension for filtering out soft deleted values from all relevant operations
+ */
+export const filterSoftDeleted = Prisma.defineExtension({
```
sazanrjb, line 51: does this run automatically in each query? or do we have to call this in queries? @rrojan

rrojan: This runs automatically in all queries @sazanrjb

sazanrjb (approved):
We should open a way to fetch deleted records as well, as we might need that

PR #37, opened 2024-05-13
Out-325 Attachments endpoint

arpandhakal:
### Tasks

- [Attachments endpoint](https://linear.app/copilotplatforms/issue/OUT-325/api-attachments-endpoint)

### Changes

- [x] Created attachments schema where each tasks has one to many relationship with attachments.
- [x] Created Create, Get and Delete endpoints for attachments.
- [x] For attachment creation, used handleUpload module from vercel/blob to use the vercel storage for attachments and retrieve url to the databse.

### Testing Criteria

- Created a dummy form in client side and used upload module from vercel/blob to handle the uploads using ngrok.
- [LOOM](https://www.loom.com/share/a01c460eb8464ebe9f0f5ee217a6a214)


prisma/schema.prisma
```diff
@@ -85,3 +87,13 @@ model ViewSetting {
   @@index([userId, workspaceId], name: "IX_ViewSettings_userId_workspaceId")
   @@map("ViewSettings")
 }
+
+model Attachment {
+  id          String   @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  task        Task     @relation(fields :[taskId], references : [id])
+  taskId      String   @db.Uuid
+  url         String   
+  createdById String   @db.Uuid
```
rrojan, line 96: Please add `createdAt` timestamp as well

prisma/schema.prisma
```diff
@@ -85,3 +87,13 @@ model ViewSetting {
...
+
+  @@map("attachments")
```
rrojan, line 98: We are following the convention of using Pascal case for table names
```suggestion

  @@map("Attachments")
```

src/app/api/attachments/attachments.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+  const user = await authenticate(request)
+  const body = (await request.json()) as HandleUploadBody
+  const attachmentsService = new AttachmentsService(user)
+  const fileUploadResponse = await handleUpload({
+    body,
+    request,
+    onBeforeGenerateToken: async (pathname: string, clientPayload: string | null) => {
+      if (!user.internalUserId) {
+        throw new Error('Not authorized')
+      }
```
rrojan, line 19: Throw `APIError` instead. Else we will return a `500` server side error http status even if this is supposed to be a `401`

src/app/api/attachments/attachments.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+      return {
+        tokenPayload: JSON.stringify({
+          data: clientPayload,
+        }),
+      }
+    },
+    onUploadCompleted: async ({ blob, tokenPayload }) => {
+      const data: CreateAttachmentRequest = {
+        taskId: tokenPayload && JSON.parse(JSON.parse(tokenPayload).data).taskId,
```
rrojan, line 28: Please use strict types and parsing while using `JSON.parse`. You never know what kind of data you are getting.

Please also use optional chaining if necessary before `.data`

src/app/api/attachments/attachments.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+  const attachments = await attachmentsService.getAttachments(id)
+  return NextResponse.json({ attachments })
+}
+
+export const deleteAttachment = async (req: NextRequest, { params: { id } }: IdParams) => {
+  const user = await authenticate(req)
+  const attachmentsService = new AttachmentsService(user)
+  await attachmentsService.deleteAttachment(id)
+  return new NextResponse(null, { status: httpStatus.NO_CONTENT })
+}
```
rrojan, line 50: By my understanding we're treating using the same named`id` for  taskId in`getAttachments`  and for `id` in `deleteAttachment`. This is inconsistent behaviour - please use taskId as a search param instead.

src/app/api/core/services/policies.service.ts
```diff
@@ -15,6 +15,7 @@ export class PoliciesService extends BaseService {
       [Resource.WorkflowState]: [UserAction.Read],
       [Resource.ViewSetting]: [],
       [Resource.Users]: [],
+      [Resource.Attachments]: [],
```
rrojan, line 18: Clients are authorized to read attachments. Please add `UserAction.Read` else the client view of tasks app will be completely broken :bow:

prisma/schema.prisma
rrojan, line 96: Please add a `workspaceId` field to this table to check whether the user's workspace owns the attachments or not. If we don't filter this way, users can easily read, delete, etc the attachments of another workspace,  which is a security concern.

So basically on any CRUD operation you have to first check if the `workspaceId` of the attachment matches the workspaceId extracted from token payload or not

src/app/api/attachments/attachments.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+import authenticate from '@api/core/utils/authenticate'
+import { CreateAttachmentRequest, CreateAttachmentRequestSchema } from '@/types/dto/attachments.dto'
+import { AttachmentsService } from '@api/attachments/attachments.service'
+import httpStatus from 'http-status'
+import { handleUpload, type HandleUploadBody } from '@vercel/blob/client'
+import { IdParams } from '@api/core/types/api'
+
+export const createAttachment = async (request: NextRequest) => {
+  const user = await authenticate(request)
+  const body = (await request.json()) as HandleUploadBody
```
rrojan, line 11: Is it possible to avoid using a type cast here. The request body could be anything and we could be vulnerable to injecting manual data. If possible parse it using a schema instead so we throw an error if the request body does not match our expected schema.

src/app/api/attachments/attachments.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    return attachments
+  }
+
+  async createAttachments(data: CreateAttachmentRequest) {
+    const policyGate = new PoliciesService(this.user)
+    policyGate.authorize(UserAction.Create, Resource.Attachments)
+    const newAttachment = await this.db.attachment.create({
+      data: {
+        ...data,
+        createdById: this.user.internalUserId as string,
```
rrojan, line 27: Do not use type cast, instead you can do `z.string().parse(this.user.internalUserId` to be safer

src/app/api/attachments/attachments.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+    onUploadCompleted: async ({ blob, tokenPayload }) => {
+      const data: CreateAttachmentRequest = {
+        taskId: tokenPayload && JSON.parse(JSON.parse(tokenPayload).data).taskId,
+        url: blob.url,
+      }
+      await attachmentsService.createAttachments(data)
+    },
+  })
+
+  return NextResponse.json(fileUploadResponse)
```
rrojan, line 35: Also return http 201 CREATED

rrojan (changes requested):

src/types/dto/attachments.dto.ts
```diff
@@ -0,0 +1,25 @@
+import { boolean, z } from 'zod'
+
+export const CreateAttachmentRequestSchema = z.object({
+  taskId: z.string().uuid(),
+  filePath: z.string(),
+  fileSize: z.number(),
+  fileType: z.string(),
```
rrojan, line 7: It might be better to make fileType an enum that only accepts a restricted list of file types. E.g. we don't want users uploading `exe` files.

We can include common document types like pdf, etc, media like jpeg, png, etc - basically anything we have a corresponding icon for

rrojan (changes requested):

rrojan:
Please resolve merge conflicts

rrojan (approved):
LGTM :medal_military:

PR #41, opened 2024-05-20
OUT 363 Add responsiveness to the task app layout

aatbip:
### Tasks

- [Add responsiveness to the task app layout](https://linear.app/copilotplatforms/issue/OUT-363/add-responsiveness-to-the-task-app-layout)

### Changes

- [x] Adds responsiveness to throughout the app


src/app/detail/ui/ToggleController.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Box, useMediaQuery } from '@mui/material'
+import { ReactNode, useEffect } from 'react'
+import { useSelector } from 'react-redux'
+
+export const ToggleController = ({ children }: { children: ReactNode }) => {
+  const { showSidebar } = useSelector(selectTaskDetails)
+  const matches = useMediaQuery('(max-width:600px)')
+
+  useEffect(() => {
+    if (matches && !matches) {
```
saugatmaharjan, line 14: Is this condition intentional @aatbip

aatbip: @saugatmaharjan It was a hack but I have added `nonMobile` to avoid confusion in the commit below.

saugatmaharjan (approved):

src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -14,6 +14,10 @@ import { ClientSideStateUpdate } from '@/hoc/ClientSideStateUpdate'
 import { deleteTask, updateAssignee, updateTaskDetail } from './actions'
 import { updateTask, updateWorkflowStateIdOfTask } from '@/app/actions'
 import { MenuBoxContainer } from '../../ui/MenuBoxContainer'
+import { MenuIcon } from '@/icons'
+import { ToggleBtn } from '@/components/buttons/ToggleBtn'
+import { ToggleButtonContainer } from '../../ui/ToggleButtonContainer'
+import { ToggleController } from '../../ui/ToggleController'
```
Comment (rrojan) [inline src/app/detail/[task_id]/[user_type]/page.tsx:20]: Please use aliased imports

src/app/detail/ui/Sidebar.tsx
```diff
@@ -111,17 +134,19 @@ export const Sidebar = ({
...
-          />
+          <Box>
+            <DatePickerComponent
+              getDate={(date) => {
+                const isoDate = formatDate(date)
+                updateTask({
+                  dueDate: isoDate,
+                })
+              }}
+              dateValue={dueDate ? isoToReadableDate(dueDate) : undefined}
```
rrojan, line 147:
```suggestion
              }}
              dateValue={!dueDate ?? isoToReadableDate(dueDate) }
```

This is optional as we would also have to handle condition for `null`

src/components/layouts/FilterBar.tsx
```diff
@@ -79,85 +79,170 @@ export const FilterBar = ({ updateViewModeSetting }: { updateViewModeSetting: (m
...
+            }}
+            buttonContent={
+              <FilterByAssigneeBtn
+                assigneeValue={assigneeValue}
+                updateAssigneeValue={updateAssigneeValue}
+                handleClick={handleFilterOptionsChange}
+              />
+            }
+          />
+
```
rrojan, line 238: This is really confusing, we have a bit of a prop hell going on here. Is this possible to clean up? Maybe change one prop to children

src/hoc/AppMargin.tsx
```diff
@@ -9,14 +9,14 @@ export enum SizeofAppMargin {
...
   }
 
   if (size === SizeofAppMargin.MEDIUM) {
-    return <Box sx={{ padding: `${py ? py : '0px'} 28px` }}>{children}</Box>
+    return <Box sx={{ padding: { xs: py ? py : '20px', sm: `${py ? py : '0px'} 28px` } }}>{children}</Box>
   }
 
   if (size === SizeofAppMargin.SMALL) {
-    return <Box sx={{ padding: `${py ? py : '0px'} 25px` }}>{children}</Box>
+    return <Box sx={{ padding: { xs: py ? py : '20px', sm: `${py ? py : '0px'} 25px` } }}>{children}</Box>
```
rrojan, line 20: Can we make this reusable but assign proper type for typesafety?

src/hoc/DndWrapper.tsx
```diff
@@ -1,9 +1,22 @@
...
+import { TouchBackend } from 'react-dnd-touch-backend'
 
 export const DndWrapper = ({ children }: { children: ReactNode }) => {
-  return <DndProvider backend={HTML5Backend}>{children}</DndProvider>
+  const matches = useMediaQuery('(max-width:600px)')
+
+  const options = {
+    scrollAngleRanges: [{ start: 300 }, { end: 60 }, { start: 120, end: 240 }],
+  }
+  console.log(matches)
```
rrojan, line 15: Please remove console log

rrojan (changes requested):
Can you please add little right padding to text field - currently the text touches the border on right while there is a clean padding on the left
image


It also seems we can only drag and drop when dragging on the top-most part of the card only.
Is it possible to make it so that touching any part of the card makes it possible to drag it? Right now text gets highlighted with has a weird user experience

a

src/app/detail/ui/Sidebar.tsx
aatbip: The above suggestion doesn't work with the type of dateValue for the `DatePickerComponent`. 
`null` cases has already been handled.

src/components/layouts/FilterBar.tsx
aatbip: `Selector` component accepts the `buttonContent` prop. `FilterByAssigneeBtn` itself is a reusable component which accepts the above props. I don't think there is a prop hell going on here since we are only drilling prop from parent to child component (level 1 drill).

src/hoc/AppMargin.tsx
aatbip: Where can we add more types here? I think `AppMargin` component is already typed. Please let me know your solution.

aatbip:
> Can you please add little right padding to text field - currently the text touches the border on right while there is a clean padding on the left image
> 
> It also seems we can only drag and drop when dragging on the top-most part of the card only. Is it possible to make it so that touching any part of the card makes it possible to drag it? Right now text gets highlighted with has a weird user experience
> 
> a

Regarding the DnD on mobile, we are currently utilizing `MobileBackend` offered by the ReactDnD which doesn't seem to be so much powerful and currently we are not able to handle a lot of cases. We have to think about writing custom solutions or migrating to a different library. 

We took decision to using ReactDnD because initially we didn't know if mobile responsiveness was needed for the Task app. However, we were provided with the mobile design later and by that time we already implemented a base DnD support using ReactDnD. 

I don't know if we want to invest time in this sprint for upgrading the ReactDnD. May be we can wait for Copilot team's feedback for the mobile DnD support and work on it later. 
cc: @potluck @makeitraina @foleyatwork @rrojan @saugatmaharjan

rrojan (approved):
LGTM :trophy: 

Thanks for clearing out all confusions related to changes

PR #45, opened 2024-05-22
OUT 390 Get rid of column action elements

aatbip:
### Tasks

- [Get rid of column action elements](https://linear.app/copilotplatforms/issue/OUT-390/get-rid-of-column-action-elements)

### Changes

- [x] Removes action items from task column and row


rrojan (approved):
LGTM :trophy:

PR #48, opened 2024-05-27
Out-398 When I am filtering on "My tasks", I should not be able to Filter by assignee

arpandhakal:
### Tasks

- [When I am filtering on "My tasks", I should not be able to Filter by assignee](https://linear.app/copilotplatforms/issue/OUT-398/when-i-am-filtering-on-my-tasks-i-should-not-be-able-to-filter-by)

### Changes

- [x] Removed filter by Assignee button if my task filter is active 

### Testing Criteria

- [ LOOM](https://www.loom.com/share/041abd2944434982a53c66117cd226ba)

### DIFF
--- a/src/components/layouts/FilterBar.tsx
+++ b/src/components/layouts/FilterBar.tsx
@@ -100,7 +100,6 @@ export const FilterBar = ({ updateViewModeSetting }: { updateViewModeSetting: (m
                     }}
                     startIcon={<FilterByAsigneeIcon />}
                     options={assignee}
-                    disabled={filterOptions[FilterOptions.TYPE] == tokenPayload?.internalUserId}
                     placeholder="Assignee"
                     value={assigneeValue}
                     selectorType={SelectorType.ASSIGNEE_SELECTOR}
--- a/src/components/layouts/FilterBar.tsx
+++ b/src/components/layouts/FilterBar.tsx
@@ -69,7 +69,6 @@ export const FilterBar = ({ updateViewModeSetting }: { updateViewModeSetting: (m
     },
   ]
   const assigneeValue = _assigneeValue as IAssigneeCombined
-
   return (
     <Box>
       <Box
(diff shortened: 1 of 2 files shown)


src/components/layouts/FilterBar.tsx
```diff
@@ -80,66 +79,74 @@ export const FilterBar = ({ updateViewModeSetting }: { updateViewModeSetting: (m
...
-                    <FilterByAssigneeBtn
-                      assigneeValue={assigneeValue}
-                      updateAssigneeValue={updateAssigneeValue}
-                      handleClick={handleFilterOptionsChange}
-                    />
-                  }
-                />
+                <FilterButtonGroup filterButtons={filterButtons} activeButtonIndex={activeButtonIndex} />
               </Box>
+              {filterOptions[FilterOptions.TYPE] !== tokenPayload?.internalUserId && (
```
aatbip, line 89: This comparison looks confusing. I know we are trying to find if a specific filterOption is active or not. But why are we comparing with `tokenPayload.internalUserId`? 

```
 filterOptions: {
    [FilterOptions.ASSIGNEE]: '',
    [FilterOptions.KEYWORD]: '',
    [FilterOptions.TYPE]: '',
  },
```
Why can't we just detect if any filter is active just by checking if `filterOptions[key] !== ''`?

aatbip (changes requested):

src/components/layouts/FilterBar.tsx
arpandhakal: We need to check specific part of FilterOptions.TYPE here. FilterOptions.TYPE contains the button group containing My tasks, My team tasks, All tasks and Client tasks button. In the above code, I applied a check if filterOptions.TYPE contains the logged in iu's id. It could have been non empty string for client's task and my team's tasks too.

aatbip (approved):
lgtm

PR #49, opened 2024-05-27
OUT-401 If I'm filtering by a user, when I click "New Task", assignee should default to the user I was filtering on

arpandhakal:
### Tasks

- [If I'm filtering by a user, when I click "New Task", assignee should default to the user I was filtering on](https://linear.app/copilotplatforms/issue/OUT-401/if-im-filtering-by-a-user-when-i-click-new-task-assignee-should)

### Changes

- [x] used the filterOptions slice to find if the assignee list with the user being filtered on 
- [x] if assignee filter is on, in New Task component, the default assignee will be the user being filtered. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/e65a7de000e8425ba769073dcfbea624)


aatbip (approved):
lgtm

PR #50, opened 2024-05-27
OUT 400 Include a preview of the task description in the client list view

aatbip:
### Tasks

- [Include a preview of the task description in the client list view](https://linear.app/copilotplatforms/issue/OUT-400/include-a-preview-of-the-task-description-in-the-client-list-view)

### Changes

- [x] Version bumped Tapwrite to make use of the `readonly` property
- [x] Adds a description in client preview using Tapwrite 
- [x] Adds `readonly` in the details page for clients


src/components/cards/ClientTaskCard.tsx
```diff
@@ -32,7 +33,7 @@ export const ClientTaskCard = ({
         <Stack direction="row" columnGap={8} alignItems="center" justifyContent="space-between">
           <Stack sx={{ width: '100%', cursor: 'pointer' }} direction="column" onClick={() => handleRouteChange()}>
             <Typography variant="sm">{task?.title}</Typography>
-            {/* <Typography variant="bodySm">{task?.body}</Typography> */}
+            <Tapwrite content={task?.body || ''} getContent={() => {}} readonly />
```
rrojan, line 36: Why is `getContent` an empty function? Is there a way we could make getContent prop optional in Tapwrite instead?

rrojan (changes requested):

src/components/cards/ClientTaskCard.tsx
aatbip: Yes sure. We can do that but it needs some work in Tapwrite. I will do it in the next PR? @rrojan

PR #53, opened 2024-05-28
OUT-377 Get rid of Archived Status

arpandhakal:
### Tasks

- [Get rid of Archived Status](https://linear.app/copilotplatforms/issue/OUT-377/get-rid-of-archived-status)

### Changes

- [x] removed 'archived' workflow status from the default workflow status


rrojan (approved):
LGTM :trophy:

PR #54, opened 2024-05-28
OUT-378 Flip order of statuses

arpandhakal:
### Tasks

- [Flip order of statuses](https://linear.app/copilotplatforms/issue/OUT-378/flip-order-of-statuses)

### Changes

- [x] applied order by query to sort 'type' property in ascending order to achieve the requirement from query level.

### Testing Criteria

- [LOOM](https://www.loom.com/share/f9cb93ad70624efdac915d25bd34af4b)

### DIFF
--- a/src/app/api/workflow-states/workflowStates.service.ts
+++ b/src/app/api/workflow-states/workflowStates.service.ts
@@ -12,7 +12,7 @@ class WorkflowStatesService extends BaseService {
     policyGate.authorize(UserAction.Read, Resource.Tasks)
 
     const filters = { where: { workspaceId: this.user.workspaceId } }
-    let workflowStates = await this.db.workflowState.findMany(filters)
+    let workflowStates = await this.db.workflowState.findMany({ ...filters, orderBy: { type: 'asc' } })
 
     if (!workflowStates.length) {
       await this.setDefaultWorkflowStates()
@@ -53,7 +53,6 @@ class WorkflowStatesService extends BaseService {
         { type: 'unstarted', name: 'Todo', key: 'todo', workspaceId: this.user.workspaceId },
         { type: 'started', name: 'In Progress', key: 'inProgress', workspaceId: this.user.workspaceId },
         { type: 'completed', name: 'Done', key: 'completed', workspaceId: this.user.workspaceId },
-        { type: 'backlog', name: 'Archived', key: 'archived', workspaceId: this.user.workspaceId },
       ],
     })
   }


src/app/api/workflow-states/workflowStates.service.ts
```diff
@@ -11,14 +11,33 @@ class WorkflowStatesService extends BaseService {
...
+    SELECT * FROM "WorkflowStates"
+    WHERE "workspaceId" = ${this.user.workspaceId}
+    ORDER BY 
+      CASE 
+        WHEN "key" = 'todo' THEN 1
+        WHEN "key" = 'inProgress' THEN 2
+        WHEN "key" = 'completed' THEN 3
+        ELSE 4
+      END;
+  `
```
rrojan, line 24: Key can be anything based on the name of the workflow state.. If i named 'todo' something like "Pending", this logic would all fall apart.

Please order using the Workflowstate type instead 
image

src/app/api/workflow-states/workflowStates.service.ts
```diff
@@ -11,14 +11,33 @@ class WorkflowStatesService extends BaseService {
...
+      SELECT * FROM "WorkflowStates"
+      WHERE "workspaceId" = ${this.user.workspaceId}
+      ORDER BY 
+        CASE 
+          WHEN "key" = 'todo' THEN 1
+          WHEN "key" = 'inProgress' THEN 2
+          WHEN "key" = 'completed' THEN 3
+          ELSE 4
+        END;
+    `
```
rrojan, line 40: Same thing about workflow state type. Also please make this query reusable since this is used in more than one place.

rrojan (changes requested):

src/app/api/workflow-states/workflowStates.service.ts
```diff
@@ -11,14 +11,33 @@ class WorkflowStatesService extends BaseService {
...
+    let workflowStates: WorkflowStateResponse[] = await this.db.$queryRaw`
+    SELECT * FROM "WorkflowStates"
+    WHERE "workspaceId" = ${this.user.workspaceId}
+    ORDER BY 
+      CASE 
+        WHEN "key" = 'todo' THEN 1
+        WHEN "key" = 'inProgress' THEN 2
+        WHEN "key" = 'completed' THEN 3
+        ELSE 4
+      END;
```
rrojan, line 23: Since we are using a raw query and have soft delete implemented prisma extensions will not work. You also have to add a case `WHERE "deletedAt" IS NULL`

rrojan (changes requested):

src/app/api/workflow-states/workflowStates.service.ts
rrojan: Can you please look into if Prisma implements case when natively? This way we can prevent writing a raw query.

src/app/api/workflow-states/workflowStates.service.ts
```diff
@@ -11,14 +11,33 @@ class WorkflowStatesService extends BaseService {
     const policyGate = new PoliciesService(this.user)
     policyGate.authorize(UserAction.Read, Resource.Tasks)
 
-    const filters = { where: { workspaceId: this.user.workspaceId } }
-    let workflowStates = await this.db.workflowState.findMany(filters)
+    let workflowStates: WorkflowStateResponse[] = await this.db.$queryRaw`
+    SELECT * FROM "WorkflowStates"
```
rrojan, line 15: Type will be `WorkflowState[]` instead, where that type comes from the schema defined in `@prisma/client`

rrojan (changes requested):

aatbip (approved):
lgtm

rrojan (approved):
LGTM :trophy:

PR #57, opened 2024-05-29
OUT 399 When IU clicks on Due Date value, they should see the calendar

aatbip:
### Tasks

- [When IU clicks on Due Date value, they should see the calendar](https://linear.app/copilotplatforms/issue/OUT-399/when-iu-clicks-on-due-date-value-they-should-see-the-calendar)

### Changes

- [x] Improves datapicker component's UI


src/components/inputs/DatePickerComponent.tsx
```diff
@@ -18,40 +19,78 @@ export const DatePickerComponent = ({ getDate, dateValue }: Prop) => {
...
   }
 
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null)
+
+  const handleClick = (event: React.MouseEvent<HTMLElement>) => {
+    setAnchorEl(anchorEl ? null : event.currentTarget)
+  }
+
+  const open = Boolean(anchorEl)
+  const id = open ? 'calender-element' : undefined
```
saugatmaharjan, line 29: can't we use `null` instead of `undefined` @aatbip

aatbip: I am using `undefined` because of [this](https://mui.com/material-ui/react-popper/) recommendation.

aatbip: @saugatmaharjan I am not able to use `null` here. Getting type error since `id` is used as attribute for `Popper` component which should have type `string | undefined`.

arpandhakal (approved):
lgtm

src/components/inputs/DatePickerComponent.tsx
saugatmaharjan: Ok

PR #63, opened 2024-05-31
Out- 330 UI - Activity component

arpandhakal:
### Tasks

- [UI - Activity component](https://linear.app/copilotplatforms/issue/OUT-330/ui-activity-component)

### Changes

- [x] designed TaskActivityComponent which renders ActivityLog component right now and Comment Component in the future.
- [x] designed Activity Log component.
- [x]  Added vertical line, bold typography and wrapper for vertical line using emotion mui.
- [x] created getTimeDifference function to calculate time for ui of activitylog.

### Testing Criteria

- [LOOM](https://www.loom.com/share/6d61bd95cb734263ad76fc0930059576)
 
### Notes

- Dependent on [OUT-369 API - Activity Log Schema](https://linear.app/copilotplatforms/issue/OUT-369/api-activity-log-schema) for activity schema. Right now I am using any type for activity data.


src/app/detail/ui/ActivityLog.tsx
```diff
@@ -0,0 +1,35 @@
+import ActivityDescription from '@/components/activity/ActivityDescription'
+import { ActivityType } from '@/types/interfaces'
+
+import { Avatar, Box, Stack } from '@mui/material'
+import { BoldTypography, StyledTypography, VerticalLine, WrapperStack } from '@/app/detail/ui/styledComponent'
+import { getTimeDifference } from '@/utils/getTimeDifference'
+
+interface Prop {
+  log: any
```
foleyatwork, line 9: We can do better than any... do we not have an interface for log?

src/app/detail/ui/TaskActivityComponent.tsx
```diff
@@ -0,0 +1,33 @@
...
+  activities: any
+}
+
+export const TaskActivityComponent = ({ activities }: Prop) => {
+  return (
+    <>
+      <Stack direction="column" alignItems="left" p="10px 5px" rowGap={5}>
+        <Typography variant="xl">Activity</Typography>
+        <Stack direction="column" alignItems="left" p="10px 5px" rowGap={4}>
+          {activities.map((item: any, index: number) => {
```
foleyatwork, line 16: Let's type item, rather than using any.

src/app/detail/ui/TaskActivityComponent.tsx
```diff
@@ -0,0 +1,33 @@
...
+              >
+                <ActivityLog log={item} isLast={index == activities.length - 1} />
+              </Box>
+            )
+          })}
+        </Stack>
+      </Stack>
+    </>
+  )
+}
```
aatbip, line 33: This component shouldn't have any state or events. I believe it is just about rendering the static content. Then why use `'use client'` and bring this on the client side to render? We can render this in the server side as activity log is a static content. 

We also don't need separate components such as `ActivityLog.tsx` and `TaskActivityComponent.tsx` just to render the list of activity logs. This is creating unnecessary modularization which is confusing. We can directly make use of single `server` component to handle the whole activity log (not comment) related rendering.

src/app/detail/ui/ActivityLog.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <Avatar alt="user" src={log?.iconImageUrl || log?.avatarImageUrl} sx={{ width: '20px', height: '20px' }} />
+        <Stack direction="row" columnGap={1}>
+          <BoldTypography>{log.details.initiator}</BoldTypography>
+          <ActivityDescription activityType={log.activityType} args={renderArgs} />
+        </Stack>
+        <StyledTypography> {getTimeDifference(log.createdAt)}</StyledTypography>
+      </Stack>
+    </WrapperStack>
+  )
+}
```
aatbip, line 35: I don't think we need this component. Currently, we have `<TaskActivityComponent/>` which is a client component with `use client` directive on the top. Then `<ActivityLog/>` which is a child and also rendered in the client. Let's not render such static items in the client side. Please check out compositional pattern in next js docs: 
https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns

src/components/activity/ActivityDescription.tsx
```diff
@@ -0,0 +1,36 @@
...
+type ActivityDescriptionProps = {
+  activityType: ActivityType
+  args?: any[]
+}
+
+export default function ActivityDescription({ activityType, args = [] }: ActivityDescriptionProps) {
+  const description = activityDescription[activityType]
+  const renderDescription = typeof description === 'function' ? description(...args) : description
+  return <> {renderDescription}</>
+}
```
aatbip, line 36: I don't think we need this as well. Including this, we are having 3 different modules just to handle static rendering of the activity logs. Instead of creating component to render different activity types (which will bring this on the client side rendering), why not create a simpler object that has the style definition and text definition. I think we can handle this in much simpler way with few lines of code.

aatbip (changes requested):
I think your approach is good but we have a lot of modularization which is not needed and most of all we are rendering everything on the client side (related to activity which are static contents). Please make changes addressing the comments below. Thanks!

src/app/detail/ui/ActivityLog.tsx
aatbip: @foleyatwork I have created schema and types for the logs in [this](https://github.com/copilot-platforms/tasks-app/pull/68) PR. After it gets merged, @arpandhakal will be utilizing the types to get rid of all `any`.

aatbip (approved):
lgtm

PR #72, opened 2024-06-07
OUT-416 File attachment not working on task creation

arpandhakal:
### Tasks

- [File attachment not working on task creation](https://linear.app/copilotplatforms/issue/OUT-416/file-attachment-not-working-on-task-creation)

### Changes

- [x] supports deletion of attachments from supabase on task-create modal close and on attachments delete
- [x] created a new api route for bulk upload of attachments supporting transactions in its corresponding service.
- [x] refactored attachmentsCard component.
- [x] added a new state of attachments in createTaskSlice and removeOneAttachment method.  
- [x] reorganized the flow of attachments 

### Testing Criteria

- [LOOM](https://www.loom.com/share/153be73736354674963d7042e409ee16)


src/app/ui/NewTaskForm.tsx
```diff
@@ -219,23 +236,59 @@ const NewTaskFormInputs = () => {
           onChange={(e) => store.dispatch(setCreateTaskFields({ targetField: 'description', value: e.target.value }))}
         />
       </Stack>
+      <Stack direction="row" columnGap={2} m="16px 0px">
+        {attachments?.map((el, key) => {
+          return (
+            <Box key={key}>
```
rrojan, line 242: It's generally a good practice to not use index as a key, especially when indexes can change when attachments are deleted.

Please use attachment's id instead

src/app/page.tsx
```diff
@@ -110,12 +117,22 @@ export default async function Main({ searchParams }: { searchParams: { token: st
...
+              return response
             }}
             updateTask={async (taskId, payload) => {
               'use server'
               await updateTask({ token, taskId, payload })
             }}
+            getSignedUrlUpload={async (fileName: string) => {
+              'use server'
+              const data = await getSignedUrlUpload(token, fileName)
+              return data
```
rrojan, line 130: Not a big issue but it can be returned directly without an in-between var :bow:

src/types/dto/attachments.dto.ts
```diff
@@ -2,7 +2,7 @@ import { boolean, z } from 'zod'
 import { FileTypes } from '@/types/interfaces'
 
 export const CreateAttachmentRequestSchema = z.object({
-  taskId: z.string().uuid(),
+  taskId: z.string(),
```
rrojan, line 5: Is there a reason .uuid() was removed here?

rrojan (changes requested):

src/types/dto/attachments.dto.ts
arpandhakal: createAttachmentRequestSchema is used in createTaskSlice for storing attachments on task creation. Initially while creating attachments on task creation, they dont have a taskId so it will be an empty string. Later on, while posting the attachments on db, taskId is appended from the created task. So, uuid was removed. Should i create another schema for posting attachments without task id for posting attachments on task creation?

rrojan (dismissed):
Everything looks good, please fix the merge conflicts and ping for re-approval :trophy:

rrojan (approved):
LGTM :trophy:

PR #73, opened 2024-06-07
OUT 430 I shouldn't be able to create a task with an empty title

aatbip:
### Tasks

- [I shouldn't be able to create a task with an empty title](https://linear.app/copilotplatforms/issue/OUT-430/i-shouldnt-be-able-to-create-a-task-with-an-empty-title)

### Changes

- [x] Adds verification for `title` field while creating task


src/app/ui/TaskBoard.tsx
```diff
@@ -129,9 +129,11 @@ export const TaskBoard = ({
         >
           <NewTaskForm
             handleCreate={() => {
-              store.dispatch(setShowModal())
-              store.dispatch(clearCreateTaskFields())
-              handleCreate({ title, body: description, workflowStateId, assigneeType, assigneeId })
+              if (title) {
+                store.dispatch(setShowModal())
+                store.dispatch(clearCreateTaskFields())
+                handleCreate({ title, body: description, workflowStateId, assigneeType, assigneeId })
```
rrojan, line 135: Please also add `title: z.string().min(1)` logic to the corresponding CreateTaskRequestSchema and parse using it

rrojan (changes requested):

src/app/ui/TaskBoard.tsx
aatbip: made changes in the commit below

rrojan (approved):
LGTM :trophy:

PR #77, opened 2024-06-10
OUT-410 When I create a task with no assignee, it doesn't show up on the page immediately

arpandhakal:
### Tasks

- [When I create a task with no assignee, it doesn't show up on the page immediately](https://linear.app/copilotplatforms/issue/OUT-410/when-i-create-a-task-with-no-assignee-it-doesnt-show-up-on-the-page)
- [When an assigned user doesn't have a profile image, show their initial](https://linear.app/copilotplatforms/issue/OUT-421/when-an-assigned-user-doesnt-have-a-profile-image-show-their-initial)

### Changes

- [x] created AvatarWithInitials component and used it on relevant places.
- [x] removed initials for no assignee
- [x] added async on handlecreate of tasks which might be the possible solution for tasks not re fetching. (Didnt face the first problem "When I create a task with no assignee, it doesn't show up on the page immediately" while testing).

### Testing Criteria

- [LOOM](https://www.loom.com/share/400baef777e440c282fe32b2c3dc8af5)


src/components/inputs/Selector.tsx
```diff
@@ -214,7 +215,8 @@ const AssigneeSelectorRenderer = ({ props, option }: { props: HTMLAttributes<HTM
       }}
     >
       <Stack direction="row" alignItems="center" columnGap={3}>
-        <Avatar
+        <AvatarWithInitials
+          altName={(option as IAssigneeCombined)?.givenName || (option as IAssigneeCombined)?.familyName}
           alt="user"
           src={(option as IAssigneeCombined).avatarImageUrl || (option as IAssigneeCombined).iconImageUrl}
```
rrojan, line 221:
```suggestion
const assignee = option as IAssigneeCombined

return (
        ...
        <AvatarWithInitials
          altName={assignee?.givenName || assignee?.familyName}
          alt="user"
          src={assignee.avatarImageUrl || assignee.iconImageUrl}
          ...
)
```

Also, I think we have a schema for combined assignee used in the backend `types` folder somewhere as well, which could be used instead of the `option as IAssigneeCombined` type casting, but this should be fine for now too.

rrojan (changes requested):

rrojan (changes requested):

rrojan (dismissed):
LGTM :trophy:

rrojan (approved):

rrojan (dismissed):
LGTM :trophy:

src/app/detail/ui/Comments.tsx
```diff
@@ -15,7 +16,12 @@ export const Comments = ({ comment, createComment, deleteComment, task_id }: Pro
   return (
     <Stack direction="row" columnGap={2} position="relative">
       <VerticalLine />
-      <Avatar alt="user" src={''} sx={{ width: '25px', height: '25px', marginTop: '5px' }} />
+      <AvatarWithInitials
+        altName={comment?.initiator?.givenName}
+        alt="user"
+        src={''}
+        sx={{ width: '25px', height: '25px', marginTop: '5px' }}
+      />
```
aatbip, line 24: I think this component name is misleading. `AvatarWithInitials` can be anything. Can you please specify what are the initials? Better to change the component name IMO.

src/utils/getNameForAvatar.ts
```diff
@@ -0,0 +1,6 @@
+import { IAssigneeCombined } from '@/types/interfaces'
+
+export const getNameForAvatar = (user: IAssigneeCombined): string => {
+  const name = user?.familyName || user?.givenName == 'No assignee' ? '' : user?.familyName || user?.givenName
+  return name ?? ''
+}
```
aatbip, line 6: We are using multiple `Avatar` components in this app. This particular utils `getNameForAvatar` is misleading IMO. Can we please specify what kind of name does it actually help returning?

aatbip (changes requested):

aatbip (dismissed):
lgtm

aatbip (approved):

PR #79, opened 2024-06-10
OUT-414 UI Bug when I click on the assignee dropdown and then hit escape

aatbip:
### Tasks

- [UI Bug when I click on the assignee dropdown and then hit escape](https://linear.app/copilotplatforms/issue/OUT-414/ui-bug-when-i-click-on-the-assignee-dropdown-and-then-hit-escape)

### Changes

- [x] `Selector` component's autocomplete dropdown closes completely after hitting ESC key twice

### DIFF
--- a/src/components/inputs/Selector.tsx
+++ b/src/components/inputs/Selector.tsx
@@ -3,7 +3,7 @@ import { SecondaryBtn } from '@/components/buttons/SecondaryBtn'
 import { StyledAutocomplete } from '@/components/inputs/Autocomplete'
 import { statusIcons } from '@/utils/iconMatcher'
 import { useFocusableInput } from '@/hooks/useFocusableInput'
-import { HTMLAttributes, ReactNode, useState } from 'react'
+import { HTMLAttributes, ReactNode, useEffect, useState } from 'react'
 import { StyledTextField } from './TextField'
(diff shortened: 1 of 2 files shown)


src/components/inputs/Selector.tsx
```diff
@@ -71,6 +71,20 @@ export default function Selector({
...
+      }
+    }
+
+    window.addEventListener('keydown', closePopper)
+
+    return () => {
+      window.removeEventListener('keydown', closePopper)
+    }
+  }, [])
+
```
rrojan, line 87: Would it make more sense to make this a custom hook instead? I think this logic is repeated for new task modal as well so.

```suggestion
import { useEffect } from 'react';

const useCloseOnEscape = (closeFunction: () => void) => {
  useEffect(() => {
    const closePopper = (e: KeyboardEvent) => {
      if (e.key === 'Escape') {
        closeFunction();
      }
    };

    window.addEventListener('keydown', closePopper);

    return () => {
      window.removeEventListener('keydown', closePopper);
    };
  }, [closeFunction]);
};

export default useCloseOnEscape;


...

const Selector = (...) => {
     useCloseOnEscape(() => setAnchorEl(null));
     
     return <... />
}

```

rrojan (changes requested):

PR #82, opened 2024-06-11
OUT-429 Client-side UI shows "no assignee"

arpandhakal:
### Tasks

- [Client-side UI shows "no assignee"](https://linear.app/copilotplatforms/issue/OUT-429/client-side-ui-shows-no-assignee)
- [Client-side Mark done not working](https://linear.app/copilotplatforms/issue/OUT-380/client-side-mark-done-not-working)

### Changes
- [x] Added UserAction.Read on User resources for preventing 401 error while fetching users from client view. User list is required to display asignee name.
- [x] Added UserAction.Update on Tasks resources for preventing 401 error while updating tasks ie handling mark done.  

### Testing Criteria

- [LOOM](https://www.loom.com/share/daef1e575f424700921d7332c07aef51)

### Notes

- Handling mark done from client side triggers a notification on task workflowstate changed. On workflowstate changed, the recipent id is taken as createdById of the task. And recipentId is passed to getClient function of copilot sdk because assigneeType will be client. 
Right now tasks are created by IUs, so using getClient() to search creator's id (ius) throws an error. Need some clarity here.

### DIFF
--- a/src/app/api/core/services/policies.service.ts
+++ b/src/app/api/core/services/policies.service.ts
@@ -10,7 +10,7 @@ import httpStatus from 'http-status'
 export class PoliciesService extends BaseService {
   private defaultPolicies: Record<UserRole.Client, Record<Resource, UserAction[]>> = {
     [UserRole.Client]: {
-      [Resource.Tasks]: [UserAction.Read],
+      [Resource.Tasks]: [UserAction.Read, UserAction.Update],
       [Resource.TaskTemplates]: [],
       [Resource.WorkflowState]: [UserAction.Read],
       [Resource.ViewSetting]: [],
--- a/src/app/api/core/services/policies.service.ts
+++ b/src/app/api/core/services/policies.service.ts
@@ -14,7 +14,7 @@ export class PoliciesService extends BaseService {
       [Resource.TaskTemplates]: [],
       [Resource.WorkflowState]: [UserAction.Read],
       [Resource.ViewSetting]: [],
-      [Resource.Users]: [],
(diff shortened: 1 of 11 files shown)


src/app/api/core/services/policies.service.ts
```diff
@@ -10,11 +10,11 @@ import httpStatus from 'http-status'
 export class PoliciesService extends BaseService {
   private defaultPolicies: Record<UserRole.Client, Record<Resource, UserAction[]>> = {
     [UserRole.Client]: {
-      [Resource.Tasks]: [UserAction.Read],
+      [Resource.Tasks]: [UserAction.Read, UserAction.Update],
```
rrojan, line 13: We shouldn't allow client user to update tasks directly since would also allow them to edit title, description, etc.

For this particular issue it would be better not add `UserAction.Update` to the permissions, rather find a workaround for it in the service layer

src/app/api/core/services/policies.service.ts
```diff
@@ -10,11 +10,11 @@ import httpStatus from 'http-status'
...
       [Resource.TaskTemplates]: [],
       [Resource.WorkflowState]: [UserAction.Read],
       [Resource.ViewSetting]: [],
-      [Resource.Users]: [],
+      [Resource.Users]: [UserAction.Read],
```
rrojan, line 17: Same here. This would allow users to view data of all of the users associated with a workspace, including user data from other organizations

rrojan (changes requested):

rrojan (approved):
LGTM 🏆

PR #102, opened 2024-06-25
OUT-428 UI Issue in Client View

arpandhakal:
### Tasks

- [UI Issue in Client View](https://linear.app/copilotplatforms/issue/OUT-428/ui-issue-in-client-view)

### Changes

- [x] added a function to extract texts from html contents of tapwrite.
- [x] displayed text similar to figma ui in client view task description.
- [x] trucated the text with max characters 456. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/34f027c463b54356b2c80428b783149b)

### Notes

- [ ] The no assignee anomaly in task board in client view is because of result returned from CopilotSDK.getClients(). It is because companyId is different in some cases for same assignee from CopilotSDK.getClients() and CopilotSDK.getTokenPayload().


rrojan (approved):
LGTM :trophy:

PR #106, opened 2024-06-27
OUT-473 | Task workflowstate change error in initial phases of application

rrojan:
### Tasks

- [Task workflowstate change error in initial phases of application](https://linear.app/copilotplatforms/issue/OUT-473/task-workflowstate-change-error-in-initial-phases-of-application)

### Changes

- [x] Add token to dependency array of useCallback

### Notes

- None


aatbip (changes requested):
The `token` state imported from `taskBoardSlice` willl never be undefined if `token` is passed in the search params. In this case, `token` was undefined inside the `useCallback` hook and wasn't being parsed properly because of the empty dependency array in the `useCallback` hook. 

Note: When a component renders, React creates a closure for the component, capturing all the variables and states within that render. The onDropItem callback function is created once using the value of `token` from the initial render. The behaviour of `token` being undefined within the callback suggests that `token` might be getting updated asynchronously or in a way that isn't immediately reflected inside the closure. Including `token` in the dependency array to ensures that the callback function is recreated with the current value of token.

aatbip (approved):

PR #107, opened 2024-06-27
OUT 354 Load Testing for Tasks App

aatbip:
### Tasks

- [Load Testing for Tasks App](https://linear.app/copilotplatforms/issue/OUT-354/load-testing-for-tasks-app)

### Changes

- [x] Adds task seeder using faker.js for fake data

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "lint-staged": "npx lint-staged",
     "postinstall": "prisma generate",
     "test": "jest",
-    "seed:task": "npx ts-node ./src/seeder/task-seeder.ts"
+    "seed:task": "npx ts-node -P tsconfig.seeder.json ./src/seeder/task-seeder.ts"
   },
   "dependencies": {
     "@emotion/react": "^11.11.3",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "prettier:fix": "prettier --write .",
     "lint-staged": "npx lint-staged",
     "postinstall": "prisma generate",
-    "test": "jest"
+    "test": "jest",
+    "seed:task": "npx ts-node ./src/seeder/task-seeder.ts"
   },
   "dependencies": {
     "@emotion/react": "^11.11.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -23,7 +23,8 @@
     "prettier:fix": "prettier --write .",
     "lint-staged": "npx lint-staged",
     "postinstall": "prisma generate",
-    "test": "jest"
+    "test": "jest",
+    "seed:task": "npx ts-node -P tsconfig.seeder.json ./src/seeder/task-seeder.ts"
```
rrojan, line 27: Is there a better way to do this without creating a separate tsconfig?

src/seeder/task-seeder.ts
```diff
@@ -0,0 +1,55 @@
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(payload),
+    })
+  } catch (e) {
+    console.log('Error:', e)
+  }
+}
```
rrojan, line 30: This is okay, but we might want to insert data into the db directly using prisma instead of hitting the endpoint. This will make it a lot slower.

src/seeder/task-seeder.ts
```diff
@@ -0,0 +1,55 @@
...
+async function seedTasks(count: number, workflowStateId: string) {
+  for (let i = 0; i < count; i++) {
+    const title = faker.fakerEN.lorem.sentence()
+    console.log(`Seeding ${i + 1} of ${count} for state ID ${workflowStateId}`)
+    await createTask({ title, workflowStateId: workflowStateId })
```
rrojan, line 35: This will seed data one by one for hundreds of tasks - since we have a very generous connection limit for dbs, we should attempt to run multiple insertions in parallel here.

src/seeder/task-seeder.ts
```diff
@@ -0,0 +1,55 @@
+import { CreateTaskRequest } from './../types/dto/tasks.dto'
+import * as faker from '@faker-js/faker'
+
+import { WorkflowStateResponse } from './../types/dto/workflowStates.dto'
+
+const apiUrl = 'http://localhost:3000'
+
+const token = ''
```
rrojan, line 8: If the token is empty, won't this stop us from extracting the value of workspaceId, etc?

rrojan (changes requested):

PR #109, opened 2024-06-28
OUT-479 Task card overflow issue

arpandhakal:
- also introduced divider in CU view

### Tasks

- [Task card overflow issue](https://linear.app/copilotplatforms/issue/OUT-479/task-card-overflow-issue)

### Changes

- [x] Fixed task card overflow in board view by truncating text which are more than 22 char for assignee Name.
- [x] Introduces divider in CU task view with 1px padding and color aligning to figma design.

### Testing Criteria

image

### DIFF
--- a/src/components/cards/ClientTaskCard.tsx
+++ b/src/components/cards/ClientTaskCard.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { Avatar, Box, Stack, Typography } from '@mui/material'
+import { Avatar, Box, Divider, Stack, Typography } from '@mui/material'
 import { SecondaryBtn } from '../buttons/SecondaryBtn'
 import { TaskResponse } from '@/types/dto/tasks.dto'
 import { useSelector } from 'react-redux'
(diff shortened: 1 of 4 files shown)


src/components/cards/TaskCard.tsx
```diff
@@ -37,8 +39,11 @@ export const TaskCard = ({ task }: { task: TaskResponse }) => {
           <Typography variant="sm" fontSize="12px" sx={{ color: (theme) => theme.color.gray[500] }}>
             {(currentAssignee as IAssigneeCombined).name === 'No assignee'
               ? 'No assignee'
-              : (currentAssignee as IAssigneeCombined)?.name ||
-                `${(currentAssignee as IAssigneeCombined)?.givenName ?? ''} ${(currentAssignee as IAssigneeCombined)?.familyName ?? ''}`.trim()}
+              : truncateText(
+                  (currentAssignee as IAssigneeCombined)?.name ||
+                    `${(currentAssignee as IAssigneeCombined)?.givenName ?? ''} ${(currentAssignee as IAssigneeCombined)?.familyName ?? ''}`.trim(),
+                  TruncateMaxNumber.TASK_CARD_NAME,
+                )}
```
aatbip, line 46: I think this computation is unnecessary overload to the app. I think it is better to create an `hoc` that supports trucating with the styling below. We can achieve this using simple css styling - 
```css
width: 146px; //the size of visible element
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```

src/components/cards/ClientTaskCard.tsx
```diff
@@ -35,89 +35,92 @@ export const ClientTaskCard = ({
...
+                      whiteSpace: 'nowrap',
+                      overflow: 'hidden',
+                    }}
+                  >
+                    {truncateText(
+                      (currentAssignee as IAssigneeCombined)?.name ||
+                        `${(currentAssignee as IAssigneeCombined)?.givenName ?? ''} ${(currentAssignee as IAssigneeCombined)?.familyName ?? ''}`.trim() ||
+                        'No Assignee',
+                      TruncateMaxNumber.SELECTOR,
+                    )}
```
aatbip, line 100: I think we are doing two things for the same purpose here. We are giving the above styling in `sx` prop which is exactly responsible for the truncate styling. We don't need to use `truncateText` function in this case.

src/types/constants.ts
```diff
@@ -2,4 +2,5 @@ export enum TruncateMaxNumber {
   ATTACHMENT = 15,
   SELECTOR = 16,
   CLIENT_TASK_DESCRIPTION = 256,
+  TASK_CARD_NAME = 22,
```
aatbip, line 5: I think all of these can be removed and simply use the css styling for truncation. We can create an `hoc` for re-usability. 

I think using such numbers for truncation will bring inconsistency and won't be usable in all kinds of design. It's better to let CSS handle such things for various screen sizes. Styling related jobs should be given to CSS as much as possible for easy life.

aatbip (changes requested):

rrojan (approved):
LGTM :trophy:

rrojan (approved):
LGTM again

rrojan (approved):

aatbip (approved):

PR #110, opened 2024-07-01
OUT-490,OUT-489,OUT-488 In the list view, hide section if they have 0 associated tasks

arpandhakal:
### Tasks

- [In the list view, hide section if they have 0 associated tasks](https://linear.app/copilotplatforms/issue/OUT-490/in-the-list-view-hide-section-if-they-have-0-associated-tasks)
- [Assignee dropdown styling](https://linear.app/copilotplatforms/issue/OUT-489/assignee-dropdown-styling) [OUT-489]
- [Date formatting for due dates](https://linear.app/copilotplatforms/issue/OUT-488/date-formatting-for-due-dates) [OUT-488]

### Changes

- [x] fixed : In the list view, workflow state is hidden if task is 0.
- [x] fixed Date formatting for due dates.
- [x] fixed Assignee dropdown styling.
- [x]  removed date-fns and used day-js for time and date calculation functions.
- [x] introduced groupRender in selector. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/4cee72cbcff34c27b61b98e077c4a0d1)


src/components/utils/DueDateLayout.tsx
```diff
@@ -1,21 +1,28 @@
...
     formattedDueDate = `Today`
-  } else if (isTomorrow(date)) {
+  } else if (parsedDate.isTomorrow()) {
     formattedDueDate = `Tomorrow`
   } else {
-    isPast = now > new Date(date)
-    formattedDueDate = format(date, 'dd, MMM, yyyy')
+    isPast = now.isAfter(parsedDate)
+    formattedDueDate = parsedDate.format('MMMM D, YYYY')
   }
```
aatbip, line 27: Let's not create a `utils` inside components. IMO, there shouldn't exist utility components. Components that performs logical operations before rendering should be higher order components and should be placed inside `hoc` folder. The date related code that you've written here should be inside an `hoc` which should pass the computed value to their child.

src/components/utils/DueDateLayout.tsx
```diff
@@ -1,21 +1,28 @@
...
+import isTomorrow from 'dayjs/plugin/isTomorrow'
+
+dayjs.extend(isToday)
+dayjs.extend(isTomorrow)
 
 interface DueDateLayoutProp {
   date: string | Date
 }
 
 export const DueDateLayout = ({ date }: DueDateLayoutProp) => {
```
aatbip, line 13: I think it is better to memoize this component. Please refer to `memo`.

src/components/utils/DueDateLayout.tsx
```diff
@@ -1,21 +1,28 @@
...
     formattedDueDate = `Today`
-  } else if (isTomorrow(date)) {
+  } else if (parsedDate.isTomorrow()) {
     formattedDueDate = `Tomorrow`
   } else {
-    isPast = now > new Date(date)
-    formattedDueDate = format(date, 'dd, MMM, yyyy')
+    isPast = now.isAfter(parsedDate)
+    formattedDueDate = parsedDate.format('MMMM D, YYYY')
   }
```
aatbip, line 26: Its better to utilize `useCallback` hook for such computations. This should be not be computed in every render unless the date changes.

src/utils/getTimeDifference.ts
```diff
@@ -1,3 +1,6 @@
-import { formatDistance } from 'date-fns'
+import dayjs from 'dayjs'
+import relativeTime from 'dayjs/plugin/relativeTime'
 
-export const getTimeDifference = (createdAt: string): string => formatDistance(createdAt, new Date())
+dayjs.extend(relativeTime)
+
+export const getTimeDifference = (createdAt: string): string => dayjs().to(dayjs(createdAt))
```
aatbip, line 6: This utility function is hard to comprehend. I think it is better to write some comments along with some example of both input and output for this function. 

I also believe that this function is more suited inside the `hoc` that I suggest to create. We don't need utilities for such computations which can simply reside inside the `hoc` or hooks.

aatbip (changes requested):

src/components/utils/DueDateLayout.tsx
arpandhakal: I think memorizing is not necessary since we're using useCallback hook.

src/components/utils/DueDateLayout.tsx
arpandhakal: I think it is not feasible to create a higher order component, however higher order functions works good. But we cannot use hooks for computations in higher order functions. What are your thoughts on this?

src/app/ui/TaskBoard.tsx
```diff
@@ -113,27 +112,29 @@ export const TaskBoard = ({
...
+                          updateTask={({ payload }) => {
+                            updateTask({ token: z.string().parse(token), taskId: task.id, payload })
+                          }}
+                          handleClick={() => advancedFeatureFlag && router.push(`/detail/${task.id}/iu?token=${token}`)}
+                        />
+                      </DragDropHandler>
+                    )
+                  })}
+                </TaskRow>
+              )}
```
rrojan, line 137: Instead is it possible to do something like this:

```suggestion
                <TaskRow
                  key={list.id}
                  columnName={list.name}
                  taskCount={taskCountForWorkflowStateId(list.id)}
                  showConfigurableIcons={false}
                  display={!!filterTaskWithWorkflowStateId(list.id).length}
                >
                  {filterTaskWithWorkflowStateId(list.id).map((task, index) => {
                    return (
                      <DragDropHandler key={task.id} accept={'taskCard'} index={index} id={task.id || ''} draggable>
                        <ListViewTaskCard
                          task={task}
                          key={task.id}
                          updateTask={({ payload }) => {
                            updateTask({ token: z.string().parse(token), taskId: task.id, payload })
                          }}
                          handleClick={() => advancedFeatureFlag && router.push(`/detail/${task.id}/iu?token=${token}`)}
                        />
                      </DragDropHandler>
                    )
                  })}
                </TaskRow>
```

yarn.lock
```diff
@@ -2951,7 +2951,7 @@ data-view-byte-offset@^1.0.0:
     es-errors "^1.3.0"
     is-data-view "^1.0.1"
 
-date-fns@*, date-fns@^3.6.0:
+date-fns@*:
```
rrojan, line 2954: Please regenerate `yarn.lock` so that date functions package is removed as a dependency

src/components/inputs/Selector.tsx
```diff
@@ -156,6 +156,23 @@ export default function Selector({
...
+                  <Box key={params.key} component="li">
+                    <Stack direction="row" alignItems="center" columnGap={2}>
+                      <Typography variant={'sm'} sx={{ color: (theme) => theme.color.gray[500], marginLeft: '10px' }} p={1}>
+                        {params.group}
+                      </Typography>
+                    </Stack>
+                    {params.children}
+                  </Box>
+                )
+              : params.children && <Box key={params.key}> {params.children}</Box>
```
rrojan, line 174:
```suggestion
           if (!params.children) return <></>
           
            return hasNoAssignee
              ? <Box key={params.key}> {params.children}</Box> 
              :(
                  <Box key={params.key} component="li">
                    <Stack direction="row" alignItems="center" columnGap={2}>
                      <Typography variant={'sm'} sx={{ color: (theme) => theme.color.gray[500], marginLeft: '10px' }} p={1}>
                        {params.group}
                      </Typography>
                    </Stack>
                    {params.children}
                  </Box>
                )
```

rrojan (changes requested):

rrojan (approved):
LGTM :trophy:

aatbip (approved):
lgtm

PR #121, opened 2024-07-03
OUT-510 Client should not be allowed to change due dates in the UI

aatbip:
### Tasks

- [Client should not be allowed to change due dates in the UI](https://linear.app/copilotplatforms/issue/OUT-510/client-should-not-be-allowed-to-change-due-dates-in-the-ui)

### Changes

- [x] Clients not allowed to modify/add due dates in the details page

### DIFF
--- a/src/app/detail/ui/Sidebar.tsx
+++ b/src/app/detail/ui/Sidebar.tsx
@@ -160,6 +160,7 @@ export const Sidebar = ({
                 })
               }}
               dateValue={dueDate ? isoToReadableDate(dueDate) : undefined}
+              disabled={disabled}
             />
           </Box>
         </Stack>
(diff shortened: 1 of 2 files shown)


rrojan (changes requested):
If client cannot make changes to the due date, it would make more sense to not make the cursor a pointer
image

aatbip:
> If client cannot make changes to the due date, it would make more sense to not make the cursor a pointer image

Fixed in commit f5043739aa005c1ec7db5123c7fa6fc5b0e399e3 for all the buttons in Sidebar

rrojan (dismissed):
LGTM :trophy:

rrojan (approved):

PR #122, opened 2024-07-03
OUT-509 No assignee shown for task assigned to client

aatbip:
### Tasks

- [No assignee shown for task assigned to client](https://linear.app/copilotplatforms/issue/OUT-509/no-assignee-shown-for-task-assigned-to-client)

### Changes

- [x] Adds read policy to authorize CU to access groupedBy users

### DIFF
--- a/src/app/api/core/services/policies.service.ts
+++ b/src/app/api/core/services/policies.service.ts
@@ -14,7 +14,7 @@ export class PoliciesService extends BaseService {
       [Resource.TaskTemplates]: [],
       [Resource.WorkflowState]: [UserAction.Read],
       [Resource.ViewSetting]: [],
-      [Resource.Users]: [UserAction.Read],
+      [Resource.Users]: [],
       [Resource.Attachments]: [UserAction.Read],
       [Resource.Comment]: [UserAction.Read],
     },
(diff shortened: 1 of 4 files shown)


rrojan (changes requested):
When i assign a task to the company that the client belongs to, it still shows no assignee in the list view

image

This behavior is not reproduced in description view

src/app/api/core/services/policies.service.ts
```diff
@@ -14,7 +14,7 @@ export class PoliciesService extends BaseService {
       [Resource.TaskTemplates]: [],
       [Resource.WorkflowState]: [UserAction.Read],
       [Resource.ViewSetting]: [],
-      [Resource.Users]: [],
+      [Resource.Users]: [UserAction.Read],
```
rrojan, line 17: This would make client users be able to fetch a complete list of users from our API which would be a security concern. It is recommended to skip the policy gate and go with a custom authorization code for this if necessary

I believe @arpandhakal has done a task similar to this before

aatbip: Fixed in commit below

rrojan (approved):
LGTM :trophy:

PR #128, opened 2024-07-04
OUT-524 List view is not aligned when label is longer than usual

arpandhakal:
### Tasks

- [List view is not aligned when label is longer than usual](https://linear.app/copilotplatforms/issue/OUT-524/list-view-is-not-aligned-when-label-is-longer-than-usual)

### Changes

- [x] added a fixed width for task label
- [x] changes made for responsive design in list view 

### Testing Criteria

- [LOOM](https://www.loom.com/share/e89f485bf5434608bab041132a088acb)


src/components/cards/ListViewTaskCard.tsx
```diff
@@ -52,42 +52,33 @@ export const ListViewTaskCard = ({
...
-                columnGap="20px"
-                sx={{
-                  minWidth: {
-                    xs: 'none',
-                    sm: '300px',
-                    md: '350px',
-                  },
-                }}
-              >
+              <Stack direction="row" alignItems="center" justifyContent={'space-between'} columnGap={1}>
```
aatbip, line 66: The way we are adding width here to define the element size is wrong here. Instead of width, we should add gaps for the flex containers. For example, see the image below. 
Screenshot from 2024-07-04 17-39-09

The title is going into next line for the size below sm i.e. 600px. If you see the Figma then we can clearly see the width for title as 135px minimum and the gap is 16px which doesn't match the implementation in this PR. 
In our implementation gap is around 33-34px - 
Screenshot from 2024-07-04 17-45-53

Gap in Figma - 
Screenshot from 2024-07-04 17-44-20

Width in Figma -
Screenshot from 2024-07-04 17-45-08

Never give a fix width for such elements. Let's give floating width and adjust the spacing using gaps.

src/components/cards/ListViewTaskCard.tsx
```diff
@@ -52,42 +52,33 @@ export const ListViewTaskCard = ({
...
                 <Box
                   sx={{
-                    display: {
-                      xs: 'none',
-                      sm: 'flex',
-                    },
+                    display: 'flex',
                     minWidth: 'fit-content',
-                    width: '200px',
+                    width: { md: '200px', sm: '100px', xs: '100px' },
```
aatbip, line 71: Same thing here. If we give 100px to xs then upto 600px width it will be 100px itself, which is not correct.

aatbip (changes requested):

aatbip (changes requested):
I looks good. The layout of ListViewTaskCard is awesome. However, there is one issue which is shown in the video below. Below 600px i.e. sm, the parent width is too high that it becomes scrollable for no reason. It doesn't fit to the screen which is a critical issue. Please kindly fix this. 

Video - 
[issue.webm](https://github.com/copilot-platforms/tasks-app/assets/38468429/74ebad26-a091-4fa1-af11-206233158d64)

aatbip (dismissed):
Awesome work!

src/components/layouts/FilterBar.tsx
```diff
@@ -210,7 +210,7 @@ export const FilterBar = ({
           direction="row"
           justifyContent="space-between"
           alignItems="center"
-          sx={{ display: { sm: 'flex', sd: 'none' }, mb: { xs: '12px', md: '0px' } }}
+          sx={{ display: { sm: 'flex', sd: 'none' }, mb: { xs: '12px', md: '0px' }, minWidth: '340px' }}
```
aatbip, line 213: This minWidth is giving unexpected issues. See below image - 

Screenshot from 2024-07-09 14-18-29

The filter bar doesn't have side margins below 350px.

aatbip (changes requested):

src/components/layouts/FilterBar.tsx
aatbip: Also, we are able to see border right on the top layout when <350px width.

PR #129, opened 2024-07-05
OUT-535 Command K should work in the Tasks App

aatbip:
### Tasks

- [Command K should work in the Tasks App](https://linear.app/copilotplatforms/issue/OUT-535/command-k-should-work-in-the-tasks-app)

### Changes

- [x] Adds an HOC that enables interruption for cmd+k to trigger browser's search bar


rrojan (approved):
LGTM, need to test in dashboard

PR #133, opened 2024-07-05
OUT-528 Simplifications for editor

aatbip:
### Changes

- [x] Updates Tapwrite version to remove extra editor formatters


rrojan (approved):
LGTM :trophy: 

Also it would be great if you could link the Tapwrite commits / PR along with this PR in cases of version bump since we're maintaining that as well

PR #136, opened 2024-07-08
OUT-534 Better escape-ing logic

arpandhakal:
### Tasks

- [Better escape-ing logic](https://linear.app/copilotplatforms/issue/OUT-534/better-escape-ing-logic)

### Changes

- [x] applied escaping logic in components to stop propagating.
- [x] created handleKeyDown functions in searchBar, and selector component

### Testing Criteria

- [LOOM](https://www.loom.com/share/2cfbe98f8edb4b6db7cd84a17a20e0c6)

### DIFF
--- a/src/components/inputs/Selector-WorkflowState.tsx
+++ b/src/components/inputs/Selector-WorkflowState.tsx
@@ -2,7 +2,7 @@ import { WorkflowStateResponse } from '@/types/dto/workflowStates.dto'
 import { SecondaryBtn } from '../buttons/SecondaryBtn'
 import { statusIcons } from '@/utils/iconMatcher'
 import { Box, ClickAwayListener, Popper, Stack, Typography } from '@mui/material'
-import { ReactNode, useEffect, useState } from 'react'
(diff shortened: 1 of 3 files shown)


src/components/inputs/Selector-WorkflowState.tsx
```diff
@@ -29,9 +29,30 @@ export const WorkflowStateSelector = ({
...
+  const handleKeyDown = useCallback(
+    (event: React.KeyboardEvent<HTMLDivElement>) => {
+      event.stopPropagation()
+      if (event.key === 'Escape') {
+        setAnchorEl(null)
+      }
+    },
+    [setAnchorEl],
+  )
+
```
aatbip, line 41: We are not passing `handleKeyDown` as a prop from parent component to child that relies on referential equality to avoid unnecessary re-renders. So I think `useCallback` here is not necessary. Is there any benefit we get with this or we are using it for some specific reason?

aatbip (changes requested):

aatbip (approved):
lgtm

PR #137, opened 2024-07-08
OUT-456 | Incorrect notification count shown for Tasks

rrojan:
### Tasks

- [OUT-456 | Incorrect notification count shown for Tasks](https://linear.app/copilotplatforms/issue/OUT-456/incorrect-notification-count-shown-for-tasks)
- [OUT-554 | When task is assigned to a company and then unassigned, the update is unsuccessful](https://linear.app/copilotplatforms/issue/OUT-554/when-task-is-assigned-to-a-company-and-then-unassigned-the-update-is)

### Changes

- [x] Add ClientNotifications table to store all the notifications that have been triggered through our Tasks app
- [x] Implement NotificationsService to add a new ClientNotification row whenever task is created with client as assignee, or a task is reassigned from one IU / client to a client
- [x] When a task is reassigned to another user and prev assignee is a client, mark notification as read for that client
- [x] When task is unassigned from a client,  mark that client's notification as read 
- [x] When a task is deleted and it belongs to a client user, mark that notification as read as well
- [x] When a client user completes a task, mark that notification as read    
- [x] Fix edge case where company task fails to update when unassigned 

### Testing Criteria

- [x] [LOOM 1](https://www.loom.com/share/35ffd2f5bb804074bf460ab4c2549e3e?sid=8463086d-15b7-4ca9-a873-d7941fee0d79) (Task creation, completion, deletion, etc)
- [x] [LOOM 2](https://www.loom.com/share/29bef263882d470e9c65aafa2ee2b488?sid=23133eb6-ac94-4343-bc7b-5d8f31d95026) (Task assignment, reassignment, unassignment)


src/app/api/notification/notification.service.ts
```diff
@@ -49,16 +49,71 @@ export class NotificationService extends BaseService {
...
+  }
+
+  /**
+   * Marks a client notification in Copilot Notifications service as read
+   * @param id Notification ID for object as exists in Copilot
+   */
+  async markClientNotificationAsRead(task: Task) {
+    const copilot = new CopilotAPI(this.user.token)
+    try {
+      const relatedNotification = await this.db.clientNotification.findFirst({
```
sazanrjb, line 84: can't there be multiple client notifications related with the same task? 

```
[
  {
    clientId: 123,
    taskId: 321,
    notificationId: xxx,
  },
  {
    clientId: 123,
    taskId: 321,
    notificationId: xxx,
  },
]
```

it will mark the first notification as read?

src/app/api/notification/notification.service.ts
```diff
@@ -49,16 +49,71 @@ export class NotificationService extends BaseService {
...
+        where: {
+          clientId: z.string().parse(task.assigneeId),
+          taskId: task.id,
+        },
+      })
+      if (!relatedNotification) {
+        console.error(`Failed to delete client notification for task id ${task.id}`)
```
sazanrjb, line 91:
```suggestion
        console.error(`Failed to delete client notification for task id ${task.id} because the notification for client ${task.assigneeId} was not found`)
```

src/app/api/notification/notification.service.ts
```diff
@@ -49,16 +49,71 @@ export class NotificationService extends BaseService {
...
+        return
+      }
+
+      await copilot.markNotificationAsRead(relatedNotification.notificationId)
+      await this.db.clientNotification.delete({ where: { id: relatedNotification?.id } })
+    } catch (e: unknown) {
```
sazanrjb, line 97: also log an actual error trace

src/app/api/notification/notification.service.ts
```diff
@@ -73,11 +128,11 @@ export class NotificationService extends BaseService {
     const getAssignedTo = async (): Promise<CopilotUser | CompanyResponse> => {
       switch (task.assigneeType) {
         case AssigneeType.internalUser:
-          return copilot.getInternalUser(senderId)
+          return await copilot.getInternalUser(senderId)
         case AssigneeType.client:
-          return copilot.getClient(recipientId)
+          return await copilot.getClient(z.string().parse(task.assigneeId))
```
sazanrjb, line 133:
```suggestion
          return await copilot.getClient(z.string().uuid().parse(task.assigneeId))
``` ?

File: src/app/api/tasks/tasks.service.ts [hunk 5]
```diff
@@ -277,6 +257,114 @@ export class TasksService extends BaseService {
...
+      console.error('Notification failed to trigger for task:', task)
+    }
+    if (task.assigneeType === AssigneeType.client) {
+      await notificationService.addToClientNotifications(task, NotificationCreatedResponseSchema.parse(notification))
+    }
+  }
+
+  private sendCompanyTaskNotifications = async (task: Task, notificationService: NotificationService) => {
+    const copilot = new CopilotAPI(this.user.token)
+    const { recipientIds } = await notificationService.getNotificationParties(
```
sazanrjb, line 282: will this always return client recipients?

src/app/api/notification/notification.service.ts
rrojan: Yes, there are notifications triggered for other events as well like you said. However, this condition is avoided because the table only logs notifications that occur when the task is created.

src/app/api/notification/notification.service.ts
rrojan: Screenshot from 2024-07-09 17-49-30
It has been logged too

src/app/api/notification/notification.service.ts
sazanrjb: hmm ok.. in that case table name is not very clear.. may be add comments / rename.. as in future other devs might use this table for other purposes

prisma/migrations/20240708131554_add_client_notifications_table/migration.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE INDEX "IX_ClientNotifications_notificationId" ON "ClientNotifications"("notificationId");
+
+-- CreateIndex
+CREATE INDEX "IX_ClientNotifications_clientId_taskId_deletedAt" ON "ClientNotifications"("clientId", "taskId", "deletedAt");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "ClientNotifications_notificationId_key" ON "ClientNotifications"("notificationId");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "ClientNotifications_clientId_taskId_deletedAt_key" ON "ClientNotifications"("clientId", "taskId", "deletedAt");
```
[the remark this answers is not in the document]
rrojan: it prevents more than one client <-> task notification from being created since we're just logging the task create notifications in our ClientNotifications table.

Another added bonus and performance improvement is for this query. Prisma extension handles filtering out soft deleted (deletedAt IS NULL) here so the index can optimize our query:
image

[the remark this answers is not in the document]
sazanrjb: if composite unique key of clientId, taskId and deletedAt is created

lets multiple rows with same clientId and taskId are possible with different deletedAt, no?

may be one unique key of clientId and taskId should be created.. if deleted at needs index, we can separately create that?

sazanrjb (approved):

PR #145, opened 2024-07-10
OUT-552 input shifting down when opening assignee dropdown in new task dialog

arpandhakal:
### Tasks

- [input shifting down when opening assignee dropdown in new task dialog](https://linear.app/copilotplatforms/issue/OUT-552/input-shifting-down-when-opening-assignee-dropdown-in-new-task-dialog)

### Changes

- [x] changed assignee dropdown height with respect to sm and xs
- [x] button fixes 

### Testing Criteria

- [LOOM](https://www.loom.com/share/2568b2b1a3024d1297a94855b539e847)

### DIFF
--- a/src/components/inputs/Selector.tsx
+++ b/src/components/inputs/Selector.tsx
@@ -35,6 +35,7 @@ interface Prop {
   disableOutline?: boolean
   padding?: string
   buttonWidth?: string
+  buttonHeight?: string
   responsiveNoHide?: boolean
 }
 
@@ -52,6 +53,7 @@ export default function Selector({
   disableOutline,
   padding,
   buttonWidth,
+  buttonHeight,
   responsiveNoHide,
 }: Prop) {
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null)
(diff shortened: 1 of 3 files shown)


src/components/inputs/Selector.tsx
```diff
@@ -355,7 +361,7 @@ const SelectorButton = ({
             fontSize: '13px',
           },
         },
-        height: '32px',
+        height: height ? height : '32px',
```
rrojan, line 364:
```suggestion
        height: height ?? '32px',
```

src/components/inputs/Selector.tsx
```diff
@@ -169,6 +172,7 @@ export default function Selector({
           }}
           openOnFocus
           onKeyDown={handleKeyDown}
+          ListboxProps={{ sx: { maxHeight: { xs: '175px', sm: '299px' } } }}
```
rrojan, line 175: On the design it seems to be 291px... can you please confirm how this value was reached

If the figma value causes any visual inconsistencies for our use case we should use this value, else we can be on the safe side and decrease it by a few pixels :stuck_out_tongue: 

Ref: https://www.figma.com/design/Z9ONck5HGIqNDszxj5YM0G/Custom-%26-Marketplace-Apps?node-id=1554-11774&m=dev

rrojan (changes requested):

rrojan (approved):
LGTM :trophy: 

Work of art!

PR #148, opened 2024-07-10
OUT-557 No assignee shown as input in the textfield of autocomplete when the task is unassigned.

arpandhakal:
### Tasks

- [No assignee shown as input in the textfield of autocomplete when the task is unassigned.](https://linear.app/copilotplatforms/issue/OUT-557/no-assignee-shown-as-input-in-the-textfield-of-autocomplete-when-the)

### Changes

- [x] passed selector value as null to selector component if assignee value == no assignee

### Testing Criteria

- [LOOM](https://www.loom.com/share/d384bb03d05f4bd0aa19a6a88ffb24ca)


aatbip (approved):
work of an art!

PR #149, opened 2024-07-10
OUT-512 Unexpected flickering

aatbip:
### Changes

- [x] Updated revalidation techniques for all the requests
- [x] Removed unwanted revalidation

Note: Optimistic update wasn't a valid option in our case because the data we want to revalidate and re-render depends upon the response from the API especially in the case of `label`. In this PR, I have tried my best to remove unexpected flickering caused due to multiple revalidations by removing the unwanted revalidation for all the cases. For ex: in `updateViewModeSettings` 3 different revalidations were performed which was unnecessary. Another ex: when new task is created, now instead of revalidation using tag, we are updating the redux store to re-render for the first time. For now, revalidation is only used in the task detail page to revalidate the main route after redirection. This shouldn't cause flickering because it revalidates the main route `/` using revalidatePath which updates the main route instead of the details page route. 

This PR has also improved the performance of the initial task creation IMO.


src/app/detail/ui/Sidebar.tsx
```diff
@@ -20,6 +20,7 @@ import ExtraOptionRendererAssignee from '@/components/inputs/ExtraOptionRenderer
 import { WorkflowStateSelector } from '@/components/inputs/Selector-WorkflowState'
 import { CopilotAvatar } from '@/components/atoms/CopilotAvatar'
 import { AssigneePlaceholder } from '@/icons'
+import { updateTaskDetail } from '../[task_id]/[user_type]/actions'
```
arpandhakal, line 23: use alias import.

arpandhakal (approved):

src/app/detail/ui/Sidebar.tsx
aatbip: fixed in commit below

arpandhakal (approved):

PR #150, opened 2024-07-10
OUT-452 | Improve Assignee Selector

rrojan:
### Tasks

- [OUT-452 | Improve Assignee Selector](https://linear.app/copilotplatforms/issue/OUT-452/improve-assignee-selector)

### Changes

- [x] Implement async type as you search filtered results fetching for all assignee selectors
- [x] Optimize fetching logic using incremental fetching in background
- [x] Add loading state while first batch of search results is returned. Rest of the batches of data will update quietly.  

### Testing Criteria

- [x] [LOOM](https://www.loom.com/share/2264c52c09524836812a2cef6b092e3b?sid=abae6adc-3301-4fb5-9423-b8ff9c8a7eba)


src/app/api/users/users.service.ts
```diff
@@ -33,14 +32,31 @@ class UsersService extends BaseService {
...
 
     return { internalUsers: ius.data, clients: clientsWithCompanyData, companies: filteredCompanies }
   }
 
+  /**
+   * Service to filter users starting with a given keyword
+   * @param keyword
+   * @returns
+   */
+  async getFilteredUsersStartingWith(keyword: string) {
```
makeitraina, line 49: this returns IU, CU, and companies, so calling the func `getFilteredUsersStartingWith` was a little unclear to me at first, and only after jumping up and down in the code did I realize what this was doing.

src/app/api/users/users.service.ts
```diff
@@ -33,14 +32,31 @@ class UsersService extends BaseService {
...
+    const filterByKeyword = (users: FilterableUser[]) => filterUsersByKeyword(users, keyword)
+
+    const { internalUsers, clients, companies } = await this.getGroupedUsers(100_000)
```
makeitraina, line 52: what is `100_000`. That's not a number right? Im unsure how this is working.

A more realistic limit might be 500.

src/app/api/users/users.service.ts
```diff
@@ -6,24 +6,23 @@ import { Resource } from '@api/core/types/api'
...
-  async getGroupedUsers() {
+  async getGroupedUsers(limit: number = this.DEFAULT_USERS_LIMIT) {
     const user = this.user
     new PoliciesService(user).authorize(UserAction.Read, Resource.Users)
 
-    const listArgs: CopilotListArgs = {
-      // We want the complete list of users / companies for now
-      limit: 10000,
-    }
+    const listArgs: CopilotListArgs = { limit }
```
makeitraina, line 25: using the limit to fetch EVERYTHING in one request is still going to cause some problems, the network might take a long time to load and you might have page size limits. What I was suggesting is that we use a large limit 1000 and make multiple calls to the backend to fetch all clients. Its slow, but at least it will always work.

src/utils/users.ts
```diff
@@ -0,0 +1,39 @@
...
+  setActiveDebounceTimeoutId: Dispatch<SetStateAction<NodeJS.Timeout | null>>,
+  setAssigneeState: Dispatch<SetStateAction<IAssigneeCombined[]>>,
+  token: string,
+  newInputValue: string,
+) => {
+  if (activeDebounceTimeoutId) {
+    clearTimeout(activeDebounceTimeoutId)
+  }
+  const newTimeoutId = setTimeout(async () => {
+    const newAssignees = await getAssigneeList(z.string().parse(token), newInputValue)
```
makeitraina, line 35: I think, this logic here is where we fetch fetch subsequent pages. That way, as we load each page of clients we can set the assignee state so users start seeing search results on the first page and as next pages complete, we add the newly found items to the state so that more items are shown. wdyt about this approach?

makeitraina (commented):
this is on the right track. I've left some comments for feedback

src/app/api/users/users.service.ts
rrojan: This is ES2021 syntactic sugar for separating larger numbers so they are easier to read. I don't see much JS/TS devs using this but from my time doing Ruby I find it to be really elegant for dev readability @makeitraina 

```js
const regularMiliion = 1000000
const readableMiliion = 1_000_000
```

Ref: https://www.30secondsofcode.org/js/s/numeric-separator/#:~:text=Numeric%20separators%20are%20a%20lesser,to%20make%20them%20more%20readable.

makeitraina: Very interesting! Good to know. This is resolved

src/app/api/users/users.service.ts
rrojan: Please look into the pagination issue mentioned in Slack

src/utils/users.ts
```diff
@@ -0,0 +1,60 @@
...
+  const newTimeoutId = setTimeout(async () => {
+    setLoading(true)
+    const newAssignees = await getAssigneeList(z.string().parse(token), newInputValue, 2000, '0') // 2000 is hardcoded for now
+    setAssigneeState(addTypeToAssignee(newAssignees))
+    setLoading(false)
+
+    // TODO: Incremental refetching is being blocked by a fault(?) in Copilot API pagination
+    // https://www.loom.com/share/3e9598dcf0de4180a255c5ce1f1173f0?sid=b6d7bb71-248e-4c36-a045-7cb5f31466be
+
+    // First we fetch the first `CHUNK_SIZE` number of results for each of IU / client / company
```
sazanrjb, line 44: are these necessary for the future?

PR #152, opened 2024-07-11
OUT-559 React minified error on conditional empty dashboard rendering

arpandhakal:
### Tasks

- [React minified error on conditional empty dashboard rendering](https://linear.app/copilotplatforms/issue/OUT-559/react-minified-error-on-conditional-empty-dashboard-rendering)

### Changes

- [x] changed position of dndwrapper to taskboard instead of rendering it in page.tsx

### DIFF
--- a/src/app/actions.ts
+++ b/src/app/actions.ts
@@ -3,13 +3,18 @@ import { apiUrl } from '@/config'
 import { CreateAttachmentRequest } from '@/types/dto/attachments.dto'
 import { CreateTaskRequest, UpdateTaskRequest } from '@/types/dto/tasks.dto'
 import { CreateViewSettingsDTO } from '@/types/dto/viewSettings.dto'
+import { revalidateTag } from 'next/cache'
 
-export const handleCreate = async (token: string, payload: CreateTaskRequest) => {
+export const handleCreate = async (token: string, payload: CreateTaskRequest, revalidate?: boolean) => {
   try {
     const response = await fetch(`${apiUrl}/api/tasks?token=${token}`, {
       method: 'POST',
       body: JSON.stringify(payload),
     })
+
+    if (revalidate) {
+      revalidateTag('getTasks')
+    }
     return await response.json()
   } catch (e: unknown) {
     console.error('Something went wrong while creating task!', e)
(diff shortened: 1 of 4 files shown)


src/app/actions.ts
```diff
@@ -3,13 +3,18 @@ import { apiUrl } from '@/config'
...
+export const handleCreate = async (token: string, payload: CreateTaskRequest, revalidate?: boolean) => {
   try {
     const response = await fetch(`${apiUrl}/api/tasks?token=${token}`, {
       method: 'POST',
       body: JSON.stringify(payload),
     })
+
+    if (revalidate) {
+      revalidateTag('getTasks')
+    }
```
aatbip, line 17: Why do we need revalidation here?

src/app/page.tsx
```diff
@@ -100,15 +101,29 @@ export default async function Main({ searchParams }: { searchParams: { token: st
...
+                return await getSignedUrlUpload(token, fileName)
+              }}
+              handleCreateMultipleAttachments={async (attachments: CreateAttachmentRequest[]) => {
+                'use server'
+                await createMultipleAttachments(token, attachments)
+              }}
+            />
+          </DndWrapper>
+        ) : (
+          <DashboardEmptyState
```
aatbip, line 125: I think we don't need to do this much just to solve the conditional rendering error.

src/components/layouts/EmptyState/DashboardEmptyState.tsx
```diff
@@ -54,6 +75,47 @@ const DashboardEmptyState = ({ userType }: { userType: UserType }) => {
...
+                  title,
+                  body: description,
+                  workflowStateId,
+                  assigneeType,
+                  assigneeId,
+                  dueDate,
+                }),
+                true, //revalidation true for DashboadEmptyState only.
+              )
+              store.dispatch(appendTask(createdTask))
```
aatbip, line 105: I think it should be possible to use the response to update the tasks state in the store rather than revalidating somewhere and somewhere not revalidating.

aatbip (changes requested):
I think these code changes doesn't justify the reason of the proposed solution for the problem we are facing. 

If you have any idea why this sort of changes should resolve the error then please kindly mention them in the comments so that we can come to a best conclusion. I think we should try to understand the reason to the problem and change just that.

My point is, rather than improving the code some of these changes are kind of making the code rather unpredictable and bulky IMO. Let me know your thoughts.

PR #153, opened 2024-07-11
OUT-559 Fix of React minified error

aatbip:
### Changes

- [x] Fixes React minified error which says`Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.` 

Note: The NewTaskForm modal which is opened using `showModal` state variable was inserted into two different components that were rendered logically. This modal continuously re-renders in both the components which resulted in maximum update depth exceeded error. By using two different instances of modal in both logically rendering components, this issue should resolve.


arpandhakal (approved):
LGTM!!

PR #157, opened 2024-07-12
OUT-556 | Add Sentry integration to Tasks App

rrojan:
### Tasks

- [OUT-556 | Review in errors in Sentry](https://linear.app/copilotplatforms/issue/OUT-556/review-in-errors-in-sentry)

### Changes

- [x] Add sentry package for NextJS
- [x] Configure Sentry to work with Next 14 server components, api routes and client components

### Testing Criteria

- [x] Tested for clint components, server components, API routes and server actions
- [x] Screenshots for caught errors:
image

### Notes

- Also configured Vercel to sync release deployments and source map uploads on every push so we can debug issues better (hopefully)
- Due to our failing `lint` action, I also wrote a workflow step to patch `stylish.js` as well. This should fix all future issues for us related to lint fails.


sazanrjb (approved):
adding library code may break in future.. so lets find a better solution in the future

PR #159, opened 2024-07-15
OUT-581 Task does not have the right code

aatbip:
### Changes

- [x] Implements delete during update task (previously unassigned) and delete task

### DIFF
--- a/src/app/api/label-mapping/label-mapping.service.ts
+++ b/src/app/api/label-mapping/label-mapping.service.ts
@@ -163,4 +163,17 @@ export class LabelMappingService extends BaseService {
       },
     })
   }
+
+  async deleteLabel(label: string) {
+    const currentLabel = await this.db.label.findFirst({
+      where: {
+        label: label,
+      },
+    })
+    await this.db.label.delete({
+      where: {
+        id: currentLabel?.id,
+      },
+    })
+  }
 }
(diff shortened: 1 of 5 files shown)


src/app/api/label-mapping/label-mapping.service.ts
```diff
@@ -163,4 +163,17 @@ export class LabelMappingService extends BaseService {
       },
     })
   }
+
+  async deleteLabel(label: string) {
+    const currentLabel = await this.db.label.findFirst({
+      where: {
+        label: label,
+      },
+    })
```
rrojan, line 172:
```suggestion
    const currentLabel = await this.db.label.findFirst({ where: { label } })
```

src/app/api/tasks/tasks.service.ts
```diff
@@ -220,6 +222,9 @@ export class TasksService extends BaseService {
       const notificationsService = new NotificationService(this.user)
       await notificationsService.markClientNotificationAsRead(task)
     }
+    //delete the associated label
+    const labelMappingService = new LabelMappingService(this.user)
+    await labelMappingService.deleteLabel(task?.label as string)
```
rrojan, line 227: Why is task possibly undefined here? Is there a condition where this is possible and has it been accounted for?

rrojan (changes requested):

src/app/api/tasks/tasks.service.ts
aatbip: changed made in the commit below

src/app/api/label-mapping/label-mapping.service.ts
aatbip: changes had been applied in the commit below

rrojan (approved):
LGTM :trophy:

PR #161, opened 2024-07-15
OUT-584 Disable re-assignment in task details

aatbip:
### Changes

- [x] Turned off re-assignment in the task details if the task is already assigned


rrojan (changes requested):
According to the latest messages by @marlonmisra and @makeitraina , we should hide the No Assignee option alltogether in all views.

Ref: https://linear.app/copilotplatforms/issue/OUT-584/disable-re-assignment-in-task-details#comment-3bed4b22

rrojan (approved):
LGTM :trophy:

PR #162, opened 2024-07-16
OUT-582 | Due date seems wrong when you make it today

rrojan:
### Tasks

- [Due date seems wrong when you make it today](https://linear.app/copilotplatforms/issue/OUT-582/due-date-seems-wrong-when-you-make-it-today)

### Changes

- [x] Change db, backend and frontend to use `Date` instead of `Datetime` / timestamp to avoid all timezone issues in the future
- [x] Migrate preview and prod db data to use Date instead of Datetime
- [x] Change dueDate prisma schema type to use `String` instead of `Datetime` since Prisma doesn't seem to support PostGres `Date` type natively. These dates are parsed at all DTO levels with a regex to avoid faulty dates from being saved.

### Testing Criteria

- [x] [LOOM](https://www.loom.com/share/74ff1d328ef54226b9d5aa398c9fd989?sid=296bd603-5de1-4f06-9658-5bfccdff272d)

## Note:
Use this SQL command to migrate your data to Date string
```sql
UPDATE "Tasks"
SET "dueDate" = SUBSTR("dueDate", 1, 10)
WHERE "dueDate" IS NOT NULL;
```

### DIFF
--- a/prisma/migrations/20240716021900_change_tasks_to_accept_date_string/migration.sql
+++ b/prisma/migrations/20240716021900_change_tasks_to_accept_date_string/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Tasks" ALTER COLUMN "dueDate" SET DATA TYPE TEXT;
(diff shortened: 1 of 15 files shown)


arpandhakal (approved):
lgtm!

PR #163, opened 2024-07-16
OUT-579 Tasks that clients mark as done disappear

aatbip:
### Changes

- [x] Adds workspaceId parameter in the `findFirst` db call in the `completeTask` API


rrojan (approved):
LGTM :trophy:

PR #167, opened 2024-07-17
OUT 574 No realtime updates on IU or Client side

aatbip:
### Changes

- [x] Adds supabase realtime feature

**Reason for adding migration - https://supabase.com/partners/integrations/prisma#troubleshooting**
Every time a migration is run through prisma, it resets the database default grant. For that reason, we have to write a migration to re-establish the missing grant.

**Note**
Please make sure the database publication for the Task model is turned on from the Supabase. It usually gets off automatically after migrations.


src/app/detail/[task_id]/[user_type]/actions.ts
```diff
@@ -4,7 +4,7 @@ import { advancedFeatureFlag, apiUrl } from '@/config'
 import { CreateAttachmentRequest } from '@/types/dto/attachments.dto'
 import { CreateComment } from '@/types/dto/comment.dto'
 import { UpdateTaskRequest } from '@/types/dto/tasks.dto'
-import { revalidatePath, revalidateTag } from 'next/cache'
+import { revalidateTag } from 'next/cache'
```
Comment (makeitraina) [inline src/app/detail/[task_id]/[user_type]/actions.ts:7]: remind me why we removed revalidatePath?

src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -102,12 +124,14 @@ export default async function TaskDetailPage({
...
 
-  const [task, assignee, attachments, activities, tokenPayload] = await Promise.all([
+  const [workflowStates, task, assignee, attachments, activities, tokenPayload, tasks] = await Promise.all([
+    getAllWorkflowStates(token),
     getOneTask(token, task_id),
     addTypeToAssignee(await getAssigneeList(token, params.user_type)),
     getAttachments(token, task_id),
     getActivities(token, task_id),
     copilotClient.getTokenPayload(),
+    getAllTasks(token),
```
Comment (makeitraina) [inline src/app/detail/[task_id]/[user_type]/page.tsx:134]: how come we do getAllTasks on details page?

src/hoc/RealTime.tsx
```diff
@@ -0,0 +1,53 @@
...
+        const newTaskArr = [...tasks.filter((task) => task.id !== updatedTask.id), updatedTask]
+        store.dispatch(setTasks(newTaskArr))
+      }
+    }
+  }
+
+  useEffect(() => {
+    const channel = supabase
+      .channel('realtime tasks')
+      .on('postgres_changes', { event: '*', schema: 'public', table: 'Tasks' }, handleTaskRealTimeUpdates)
```
makeitraina, line 44: interesting, how are we making sure that only the relevant tasks get updated in real time. E.g. a client should not get a real-time update for a task they don't have access to.

src/hoc/RealTime.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (payload.eventType === 'INSERT') {
+      store.dispatch(setTasks([...tasks, payload.new]))
+    }
+    if (payload.eventType === 'UPDATE') {
+      const updatedTask = payload.new
+      if (updatedTask.deletedAt) {
+        const newTaskArr = tasks.filter((el) => el.id !== updatedTask.id)
+        store.dispatch(setTasks(newTaskArr))
+        if (pathname.includes('detail')) {
+          router.push(`/?token=${token}`)
```
makeitraina, line 32: can you explain why we are doing this?  adding a code comment here would be helpful

src/hooks/useHandleSelectorComponent.tsx
```diff
@@ -15,12 +15,12 @@ export const useHandleSelectorComponent = ({ item, type }: { item: unknown; type
   }
 
   useEffect(() => {
-    if (type === SelectorType.STATUS_SELECTOR) {
+    if (type === SelectorType.STATUS_SELECTOR && item) {
```
makeitraina, line 18: why add this check for `item`? can you add code comment here.

makeitraina (commented):
Nice to see us using supabase. It seems like a reasonable effort but will be worth it. I just did an initial pass with some questions/feedback but this is looking like its on the right track

sazanrjb (dismissed):

sazanrjb:
Approved the PR by mistake :facepalm: Please ignore it

src/app/detail/[task_id]/[user_type]/actions.ts
Comment (aatbip) [reply src/app/detail/[task_id]/[user_type]/actions.ts:7]: We have removed `revalidatePath` from here because from now onwards we are depending on the `RealTime` HOC for updating page after mutations by updating the `tasks` state in the redux store. For the real time functionality, after any database events we could have refetched the data, but this approach would take time to fetch and it is our requirement to not have any loading states in the app while doing activities like dragging/dropping and task creation. For that reason, when we started with real time functionality, we decided to remove revalidation and instead update the state in the redux.

src/app/detail/[task_id]/[user_type]/page.tsx
Comment (aatbip) [reply src/app/detail/[task_id]/[user_type]/page.tsx:134]: We are doing `getAllTasks` here for the real time functionality of the tasks details page. When a task title is updated, supabase event is listened in the `RealTime` hoc where we also update the `tasks` store with the updated task. After that, we update the `tasks` state in the redux store which is filtered out with the `task_id` in the details page to show the updated `title` in real time. In contrast, we need `tasks` state which we update from `getAllTasks`. 
We have to fetch `getAllTasks` here because next js doesn't load the page of `/` route and doesn't fetch tasks and update the `tasks` state if users access the `/details` route.

src/hoc/RealTime.tsx
aatbip: It only shows the tasks that belongs to respective clients and respective workspace because it does filtering on the client side based on the workspace and clients.

src/hoc/RealTime.tsx
aatbip: We are doing this to redirect user to the `/` route if they are in `/details` route when the task is deleted.

src/hooks/useHandleSelectorComponent.tsx
aatbip: `item` can be null and we don't want this block to run if it is null. That's why we are making a check here.

src/app/detail/ui/Sidebar.tsx
```diff
@@ -15,61 +15,74 @@ import { UpdateTaskRequest } from '@/types/dto/tasks.dto'
...
     type: SelectorType.ASSIGNEE_SELECTOR,
   })
 
   const statusValue = _statusValue as WorkflowStateResponse //typecasting
   const assigneeValue = _assigneeValue as IAssigneeCombined //typecasting
 
+  useEffect(() => {
+    if (tasks && workflowStates) {
+      const currentTask = tasks.find((el) => el.id === task_id)
+      const currentWorkflowState = workflowStates.find((el) => el?.id === currentTask?.workflowStateId)
```
rrojan, line 75: This should be unnecessary since we are including `workflowState` in the tasks API response right? Unless you are doing some filtering to remove the workflowState key, after you find the current task the `workflowState` key will contain the necessary info and another `.find` should be unnecessary.

src/app/detail/ui/Sidebar.tsx
```diff
@@ -15,61 +15,74 @@ import { UpdateTaskRequest } from '@/types/dto/tasks.dto'
...
+      const currentAssigneeId = currentTask?.assigneeId
+      updateStatusValue(currentWorkflowState)
+      updateAssigneeValue(currentAssigneeId ? assignee.find((el) => el.id === currentAssigneeId) : NoAssignee)
+      setDueDate(currentTask?.dueDate)
+    }
+  }, [tasks])
```
rrojan, line 81: Please address the missing dependencies warning here

src/app/detail/ui/Sidebar.tsx
```diff
@@ -184,10 +197,13 @@ export const Sidebar = ({
           </StyledText>
           <DatePickerComponent
             getDate={(date) => {
-              const dueDate = DateStringSchema.parse(formatDate(date))
-              updateTask({ dueDate })
+              const isoDate = formatDate(date as string)
+              setDueDate(date as string)
```
rrojan, line 201: Is there a reason why we are we removing a schema parse here and type casting instead?
The role of DateStringSchema is to make sure the date is in YYYY-MM-DD format. the `*date` functions could break if it isn't used.

src/app/detail/ui/Sidebar.tsx
```diff
@@ -184,10 +197,13 @@ export const Sidebar = ({
...
+              updateTask({
+                dueDate: isoDate,
+              })
             }}
-            dateValue={dueDate ? createDateFromFormattedDateString(dueDate) : undefined}
+            dateValue={dueDate ? (dueDate as Date) : undefined}
```
rrojan, line 206: I think this is reverting previous changes from OUT-582. Is this intentional?

src/app/detail/ui/TaskEditor.tsx
```diff
@@ -67,6 +63,10 @@ export const TaskEditor = ({
       }
     }
   }
+  useEffect(() => {
+    setUpdateTitle(tasks.find((el) => el.id === task_id)?.title || '')
+    setUpdateDetail(tasks.find((el) => el.id === task_id)?.body ?? '')
```
rrojan, line 68:
```suggestion
   const currentTask = tasks.find((el) => el.id === task_id)
    setUpdateTitle(currentTask?.title || '')
    setUpdateDetail(currentTask?.body ?? '')
```

src/components/inputs/DatePickerComponent.tsx
```diff
@@ -8,14 +9,14 @@ import { SecondaryBtn } from '../buttons/SecondaryBtn'
 import { useState } from 'react'
 
 interface Prop {
-  getDate: (value: string) => void
+  getDate: (value: unknown) => void
```
rrojan, line 12: Looking at its usage, this seems very weakly typed.

src/hoc/RealTime.tsx
rrojan: Can you confirm that filtering on client side based on workspace actually prevents this?

Let's say workspace A has tasks 1, 2, 3 and B has 4, 5, 6. Are we sure that changes to 1 / 2 / 3 wouldn't cause our channel to update data for workspace B?

rrojan (changes requested):

src/app/detail/ui/Sidebar.tsx
aatbip: This is necessary here because the tasks response doesn't provide the complete information of the workflowState which are required here.

src/app/detail/ui/TaskEditor.tsx
aatbip: applied in the commit below

src/components/inputs/DatePickerComponent.tsx
aatbip: fixed in the commit below

src/hoc/RealTime.tsx
aatbip: I have already confirmed it in the above comment. Adding a link to loom demo - 
[LOOM](https://www.loom.com/share/f13edbc5e18c49f683221ec92afd3c5d?sid=88eb2c27-a6bc-4146-b159-52ac67e7c438)

rrojan:
LGTM so far, please fix the merge conflicts on details page

rrojan (dismissed):
LGTM, great job with this one @aatbip  :trophy:

rrojan (approved):

PR #168, opened 2024-07-18
OUT-578 Unclear when task update has saved

arpandhakal:
### Tasks

- [Unclear when task update has saved](https://linear.app/copilotplatforms/issue/OUT-578/unclear-when-task-update-has-saved)

### Changes

- [x] created a debounce hook
- [x] applied debouce with 500ms delay for title and description update 

### Testing Criteria

- [LOOM](https://www.loom.com/share/cef8c216886c4e8cb73705d9d2510c49)

### DIFF
--- a/src/app/detail/ui/TaskEditor.tsx
+++ b/src/app/detail/ui/TaskEditor.tsx
@@ -19,6 +19,7 @@ import { ISignedUrlUpload, UserType } from '@/types/interfaces'
 import { WorkflowStateResponse } from '@/types/dto/workflowStates.dto'
 import { advancedFeatureFlag } from '@/config'
 import { Tapwrite } from 'tapwrite'
+import { useDebounce } from '@/hooks/useDebounce'
 
 interface Prop {
   title: string
@@ -67,6 +68,12 @@ export const TaskEditor = ({
     }
   }
 
+  const _taskUpdateDebounced = async (title: string, details: string) => {
+    updateTaskDetail(title, details)
+  }
+
+  const taskUpdateDebounced = useDebounce(_taskUpdateDebounced)
+
   return (
     <>
       <StyledTextField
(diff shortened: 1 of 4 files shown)


src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,30 @@
...
+export function useDebounce<Func extends SomeFunction>(func: Func, delay = 300) {
+  const timer = useRef<Timer>()
+  useEffect(() => {
+    return () => {
+      if (!timer.current) return
+      clearTimeout(timer.current)
+    }
+  }, [])
+
+  const debouncedFunction = ((...args) => {
```
rrojan, line 21:
```suggestion
  const debouncedFunction = ((...args: Parameters<Func>) => {
```

This makes the args strongly typed

src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,30 @@
...
+type SomeFunction = (...args: any[]) => void
+/**
+ *
+ * @param func The original, non debounced function (You can pass any number of args to it)
+ * @param delay The delay (in ms) for the function to return
+ * @returns The debounced function, which will run only if the debounced function has not been called in the last (delay) ms
+ */
+
+export function useDebounce<Func extends SomeFunction>(func: Func, delay = 300) {
+  const timer = useRef<Timer>()
```
rrojan, line 13: It would be better to start this with null for better type safety
```suggestion
  const timer = useRef<Timer | null>(null)
```

src/app/detail/ui/TaskEditor.tsx
```diff
@@ -67,6 +68,12 @@ export const TaskEditor = ({
     }
   }
 
+  const _taskUpdateDebounced = async (title: string, details: string) => {
+    updateTaskDetail(title, details)
+  }
```
rrojan, line 73: Shouldn't `updateTaskDetail` be awaited here?

rrojan (changes requested):

src/app/detail/ui/TaskEditor.tsx
arpandhakal: I dont think this is necessary. the function is passed from server component and is awaited there.

rrojan (approved):
LGTM 🏆

PR #171, opened 2024-07-19
OUT-573 | Search for client has inconsistencies

rrojan:
### Tasks

- [OUT-573 | Search for client has inconsistencies](https://linear.app/copilotplatforms/issue/OUT-573/search-for-client-has-inconsistencies)

### Changes

- [x] Reenable commented out loading spinner
- [x] Implement search for both given and family name at same time
- [x] Implement `string contains`  search

### DIFF
--- a/src/components/layouts/FilterBar.tsx
+++ b/src/components/layouts/FilterBar.tsx
@@ -146,25 +146,25 @@ export const FilterBar = ({
                     placeholder="Assignee"
                     value={assigneeValue}
                     selectorType={SelectorType.ASSIGNEE_SELECTOR}
-                    //****Disabling re-assignment completely for now***
-                    // extraOption={NoAssigneeExtraOptions}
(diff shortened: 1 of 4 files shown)


src/components/layouts/FilterBar.tsx
```diff
@@ -146,25 +146,25 @@ export const FilterBar = ({
...
-                    //             setAnchorEl(anchorEl ? null : e.currentTarget)
-                    //             handleFilterOptionsChange(FilterOptions.ASSIGNEE, 'No assignee')
-                    //           }}
-                    //         />
-                    //         {loading && <MiniLoader />}
-                    //       </>
-                    //     )
-                    //   )
-                    // }}
+                    extraOption={NoAssigneeExtraOptions}
```
makeitraina, line 149: tech debt: we should be sharing this code for assignee drop-down right? Lets try to refactor this when we add back re-assignment, wdyt?

PR #173, opened 2024-07-19
OUT-572 | Only show tasks and clients which IU has access to

rrojan:
### Tasks

- [OUT-572 | Only show clients where IU has access](https://linear.app/copilotplatforms/issue/OUT-572/only-show-clients-where-iu-has-access)

### Changes

- [x] Filter out clients and companies that are not in IU's company access list
- [x] Filter out tasks that belong to client and companies that are not in IU's company access list 

### Testing Criteria

- [x] [LOOM](https://www.loom.com/share/3174cd95e4b647fe96c9786eaf7f47d8?sid=aeec395e-6998-460a-adc4-2c8f5477d153)

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -83,7 +83,8 @@ export class TasksService extends BaseService {
     const currentInternalUser = await copilot.getInternalUser(this.user.internalUserId)
     if (!currentInternalUser.isClientAccessLimited) return tasks
 
-    const clients = await copilot.getClients()
+    const hasClientTasks = tasks.some((task) => task.assigneeType === AssigneeType.client)
+    const clients = hasClientTasks ? await copilot.getClients() : { data: [] }
 
     return tasks.filter((task) => {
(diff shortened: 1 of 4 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -67,12 +67,37 @@ export class TasksService extends BaseService {
...
+    if (!this.user.internalUserId) {
+      return tasks
+    }
+
+    // Now we have the challenge of figuring out if a task is assigned to a client / company that falls in IU's access list
+    const copilot = new CopilotAPI(this.user.token)
+    const currentInternalUser = await copilot.getInternalUser(this.user.internalUserId)
+    if (!currentInternalUser.isClientAccessLimited) return tasks
+
+    const clients = await copilot.getClients()
```
sazanrjb, line 86:
```suggestion
    const hasClientTasks = tasks.some(task => task.assigneeType === AssigneeType.client);
    const clients = hasClientTasks ? await copilot.getClients() : []
```

can we do this to save an API to Copilot?

sazanrjb (approved):

PR #174, opened 2024-07-19
OUT-571 | Assignee dropdown continuously resets

rrojan:
### Tasks

- [OUT-571 | Assignee dropdown continuously resets](https://linear.app/copilotplatforms/issue/OUT-571/assignee-dropdown-continuously-resets)

### Changes

- [x] Fix assignee dropdown continuously resetting issue

### Testing Criteria

- [x] [LOOM](https://www.loom.com/share/ee3714e454db4440a416de0729bb5dc0?sid=03d4f52a-3370-4c2f-a224-b66c5c183a44)

### DIFF
--- a/src/app/ui/NewTaskForm.tsx
+++ b/src/app/ui/NewTaskForm.tsx
@@ -75,7 +75,7 @@ export const NewTaskForm = ({
   const assigneeValue = _assigneeValue as IAssigneeCombined //typecasting
   const templateValue = _templateValue as ITemplate //typecasting
   // use temp state pattern so that we don't fall into an infinite loop of assigneeValue set -> trigger -> set
-  const [tempAssignee, _] = useState<IAssigneeCombined | null>(assigneeValue)
+  const [tempAssignee, setTempAssignee] = useState<IAssigneeCombined | null>(assigneeValue)
 
   const router = useRouter()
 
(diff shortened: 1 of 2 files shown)


src/app/ui/NewTaskForm.tsx
```diff
@@ -75,6 +74,8 @@ export const NewTaskForm = ({
   const statusValue = _statusValue as WorkflowStateResponse //typecasting
   const assigneeValue = _assigneeValue as IAssigneeCombined //typecasting
   const templateValue = _templateValue as ITemplate //typecasting
+  // use temp state pattern so that we don't fall into an infinite loop of assigneeValue set -> trigger -> set
+  const [tempAssignee, setTempAssignee] = useState<IAssigneeCombined | null>(assigneeValue)
```
sazanrjb, line 78: When is this variable set?

rrojan: This state is initially set based on assigneeValue. assigneeValue is managed in our Redux state.

assigneeValue is set in one of these ways
- Value of Filter by assignee
- When user clicks "+" icon in workflow state
- When user selects an assignee from this selector

rrojan: In this case the infinite rerender was caused due to this flow
1. assigneeValue is set by Filter by assignee
2. Dropdown truncates assigneeValue
3. assigneeValue is inconsistent between the redux state and selector
4. We hit a network call and update assigneeValue
5. assigneeValue is truncated and we get inconsistent data. Back to step 2.

rrojan: The UI code is very tightly coupled. Issues like this will keep coming up until we fix the root cause.

sazanrjb (approved):

PR #180, opened 2024-07-22
OUT-576 | Selecting assignee sometimes doesn't save and open tasks gives error

rrojan:
### Tasks

- [OUT-576 | Selecting assignee sometimes doesn't save and open tasks gives error](https://linear.app/copilotplatforms/issue/OUT-576/selecting-assignee-sometimes-doesnt-save-and-open-tasks-gives-error)

### Changes

- [x] Make users route support results limit for grouped users
- [x] Implement time based ISR so we fetch details for required users without hampering performance on each page load 

### Testing Criteria

- [x] [LOOM](https://www.loom.com/share/3a5eafea0e3d43e4a2b6a100f93293b5?sid=59c7ab66-d135-424c-b114-af4b52c11c10)


src/app/client/page.tsx
```diff
@@ -30,8 +31,8 @@ async function getAllTasks(token: string): Promise<TaskResponse[]> {
 }
 
 async function getAssigneeList(token: string): Promise<IAssignee> {
-  const res = await fetch(`${apiUrl}/api/users/client?token=${token}`, {
-    next: { tags: ['getAssigneeList'] },
+  const res = await fetch(`${apiUrl}/api/users/client?token=${token}&limit=${MAX_FETCH_ASSIGNEE_COUNT}`, {
+    next: { tags: ['getAssigneeList'], revalidate: ASSIGNEE_REVALIDATION_INTERVAL },
```
aatbip, line 35: I think using revalidation on the individual fetch doesn't make any difference since we are using `export const fetchCache = 'force-no-store'` and it affects all the fetch requests in this server component so revalidation parameter for individual fetch request doesn't work.

aatbip (changes requested):
I found this issue while testing the PR. If I create a new user when the app is already loaded, then I can search the new user in the assignee selector but it still appears as `No assignee` and the same error is replicated (defined in this ticket) when I go to the details page. Please see the below loom video for a demonstrated issue - 
(LOOM)[https://www.loom.com/share/792537fe15bb4b538f4d87fca05d8ab1?sid=182a9bae-3a17-49e3-bfa7-ef0f47e847ff].

aatbip (changes requested):
Just noticed one more issue here. When I create a new client from the dashboard, and then create a new task, I don't see the assignee name instead I see "No assignee". Also, when I click on the task, I can now go to the details page without error, but when I click back, the task disappears from the task board. Please see the loom below for the demonstration - 

[LOOM](https://www.loom.com/share/ec78d28bf97543ba80dc26af474f0952?sid=dd9bdcf5-405f-4f3d-a2ea-4ad2a824a1a6)

src/app/actions.ts
```diff
@@ -11,6 +11,7 @@ export const handleCreate = async (token: string, payload: CreateTaskRequest) =>
       method: 'POST',
       body: JSON.stringify(payload),
     })
+    revalidateTag('getAssigneeList')
```
aatbip, line 14: We decided to remove revalidation to prevent the flickering issue and mainly real time udpates PR would not require to revalidate as revalidation philosophy has adapted to work with the supabase revalidate which resulted in high efficiency.

I think this way to re-fetch assignee each time after task is created is also increasing the overload of the app. My suggestion is, why fetch the list of assignee again if it has already been fetched. We can just mutate the existing `assignee` array with the newly selected assignee. 

The above practice is going to solve a lot of problem such as "No assignee" issue when new task is created with newly created client, and other flickering issue. Please let me know your thoughts regarding this. 

@rrojan

aatbip (changes requested):

aatbip (changes requested):
Good work. A minor issue still remains though. Please check the loom below. 

[loom](https://www.loom.com/share/4a7c4ed022d54d348b437d836b00111e?sid=729e6822-9e22-497d-b321-542eb02d6d90)

aatbip (approved):
lgtm

PR #188, opened 2024-07-24
OUT-633 Make assignee a required field while creating task

aatbip:
### Changes

- [x] Checks if assignee is null or undefined before task creation. If it is undefined/null then task creation doesn't proceed.


rrojan (approved):
LGTM :trophy:

PR #189, opened 2024-07-25
OUT-618 Rich text editor formatting incorrect on task details

aatbip:
)### Changes

- [x] This PR fixes the formatter styling using Tapwrite version bump [see here](https://github.com/aatbip/tapwrite)

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "react-dnd-touch-backend": "^16.0.1",
     "react-dom": "^18",
     "react-redux": "^9.1.0",
-    "tapwrite": "^1.0.52",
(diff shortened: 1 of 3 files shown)


rrojan (approved):
LGTM :trophy:

makeitraina:
@aatbip Can you add more information in the description. I would love to understand why we needed to remove this css, and why bumping tapwrite version was needed for this fix.

I saw the commit made in tapwrite, it also had not explanation there, all I see is that we added !important to many styles. Im guessing some css was being overwritten in this repo, but why was that happening. If we removed the css here, why did we have to add !important in tap-write?

PR #190, opened 2024-07-25
Hotfix: Grant Supabase privileges in a way it only runs on Prod and preview

rrojan:
### Description
- Currently, `supabase-privilige/migration.sql` fails locally because we use dockerized Postgres, which doesn't have the access roles in use in this migration file.
- Using it as a migration is also unintuitive - it would make much more sense to have it be a part of our CI / build step instead.

### Changes:
- Move migration to a separate lib smd
- Add `&& yarn db:grant-supabase-privileges` to our Vercel build command
image


aatbip (approved):
lgtm

sazanrjb (commented):
Supabase does not reset permissions unless there is a problem with our migrations.

Have you verified if we have any problems in migration? Not sure why we have to grant the permissions in each deploy.

aatbip:
> Supabase does not reset permissions unless there is a problem with our migrations.
> 
> Have you verified if we have any problems in migration? Not sure why we have to grant the permissions in each deploy.

This doc shares more about the missing grants - https://supabase.com/partners/integrations/prisma#troubleshooting

PR #192, opened 2024-07-26
OUT-638 Filter drop-down doesn't reset to default when filtered with a specific assignee and switched to another tab

arpandhakal:
### Tasks

- [Filter drop-down doesn't reset to default when filtered with a specific assignee and switched to another tab](https://linear.app/copilotplatforms/issue/OUT-638/filter-drop-down-doesnt-reset-to-default-when-filtered-with-a-specific)

### Changes

- [x] enforced change filteredAssignee if buttons are only changed
- [x] changed handleInputChange prop in selector from autcomplete's onInputChange to autocomplete's textfield's onChange. 

### Testing Criteria 

- [LOOM](https://www.loom.com/share/e73b54231dda4219a938c7e182063fa8)

### Notes

- solved the issue described in linear ticket, still found some debouncing issue here, making a PR to test it in preview.


rrojan (approved):
LGTM :trophy:

PR #198, opened 2024-07-29
OUT-658 Clients should have the ability to change the status from the details page for m1

arpandhakal:
### Tasks

- [Clients should have the ability to change the status from the details page for m1](https://linear.app/copilotplatforms/issue/OUT-658/clients-should-have-the-ability-to-change-the-status-from-the-details)

### Changes

- [x] changed the completeTask api to clientUpdateTask to fit the needs of mark done as well as change workflowstate from details page

### Testing Criteria

- [LOOM](https://www.loom.com/share/9f8ebecccc0e42689cf1a9e7f853f511)


src/app/api/tasks/tasks.controller.ts
```diff
@@ -50,10 +50,10 @@ export const deleteTask = async (req: NextRequest, { params: { id } }: IdParams)
   return new NextResponse(null, { status: httpStatus.NO_CONTENT })
 }
 
-export const completeTask = async (req: NextRequest, { params: { id } }: IdParams) => {
+export const clientUpdateTask = async (req: NextRequest, { params: { id } }: IdParams) => {
   const user = await authenticate(req)
-
+  const workflowStateId = req.nextUrl.searchParams.get('workflowStateId') ?? undefined
```
rrojan, line 55: I think we could use the default null return from `#nextUrl.searchParams.get`, and accomodate `clientUpdateTask` for this.

src/app/detail/[task_id]/[user_type]/actions.ts
```diff
@@ -45,6 +45,13 @@ export const updateWorkflowStateIdOfTask = async (token: string, taskId: string,
   }
 }
 
+export const clientUpdateTask = async (token: string, taskId: string, targetWorkflowStateId: string) => {
+  await fetch(`${apiUrl}/api/tasks/${taskId}/complete?token=${token}&workflowStateId=${targetWorkflowStateId}`, {
```
Comment (rrojan) [inline src/app/detail/[task_id]/[user_type]/actions.ts:49]: This works, but intuitively it wouldn't make sense for an endpoint named `/complete` to allow a task to be changed to other types of workflowStates as well. E.g. I could hit `/complete` to change my task from a state of complete to unstarted too.

Let me know your thoughts on this.

src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -267,10 +268,17 @@ export default async function TaskDetailPage({
...
-              }}
+              updateWorkflowState={
+                params.user_type === UserType.CLIENT_USER
+                  ? async (workflowState) => {
+                      'use server'
+                      await clientUpdateTask(token, task_id, workflowState.id)
+                    }
+                  : async (workflowState) => {
+                      'use server'
+                      await updateWorkflowStateIdOfTask(token, task_id, workflowState?.id)
```
Comment (rrojan) [inline src/app/detail/[task_id]/[user_type]/page.tsx:279]:
```suggestion
                async (workflowState) => {
                      'use server'
                      params.user_type === UserType.CLIENT_USER
                  ? await clientUpdateTask(token, task_id, workflowState.id)
                     : await updateWorkflowStateIdOfTask(token, task_id, workflowState?.id)
                }
```

src/app/detail/ui/Sidebar.tsx
```diff
@@ -117,7 +119,7 @@ export const Sidebar = ({
               updateStatusValue(value)
               updateWorkflowState(value)
             }}
-            disabled={disabled}
+            disabled={workflowDisabled}
```
rrojan, line 122: Is there a reason why we are completely removing `disabled` and replacing it with `workflowDisabled` here?

rrojan (changes requested):

src/app/detail/[task_id]/[user_type]/actions.ts
Comment (arpandhakal) [reply src/app/detail/[task_id]/[user_type]/actions.ts:49]: Im thinking about changing the name of the route from 'complete' to 'client'. does that sound good? Also, at first I thought about having separate methods/route for complete and updatetasks, but since both of these function are so similar and they are only used by clients. I decided to merge them. What do you think?

src/app/detail/ui/Sidebar.tsx
arpandhakal: disabled prop is being passed from page.tsx as true for clients, since this ticket requires clients to be able to edit workflowstates, i introduced workflowDisabled prop (default false) so that it could be changed according to need in the future.

rrojan (dismissed):
LGTM :trophy:

rrojan (approved):
LGTM :trophy:

PR #200, opened 2024-07-29
OUT 661 Default order of tasks should be most recent at the top

aatbip:
### Changes

- [x] Adds sort by descending order for tasks


rrojan (approved):
LGTM :trophy:

PR #205, opened 2024-07-30
OUT-663 We can not delete search text from search box. We should have cross icon once we type search text in Search text field

arpandhakal:
### Tasks

- [We can not delete search text from search box. We should have cross icon once we type search text in Search text field](https://linear.app/copilotplatforms/issue/OUT-663/we-can-not-delete-search-text-from-search-box-we-should-have-cross)

### Changes

- [x] added a cross icon and implemented the cross icon's functionality to close the search bar and clear contents.
- [x] faced an issue of rate limit and mitigated it with introducing debouncing for the search bar with 300ms delay.

### Testing Criteria

- [LOOM](https://www.loom.com/share/e3eac43742f64baeaa3f85f84bd30000)


rrojan (approved):
LGTM :trophy:

PR #208, opened 2024-07-31
OUT-674 Improve click target for task item

arpandhakal:
### Tasks

- [Improve click target for task item](https://linear.app/copilotplatforms/issue/OUT-674/improve-click-target-for-task-item)

### Changes

- [x] made taskcontainer the child of custom link from vice versa, the cause for this problem was that custom link was a child of task container and due to the padding applied in task container, the scope of custom link componet was reduced.

### Testing Criteria

- [LOOM](https://www.loom.com/share/10816b44ee3d430fbf93201b885d86bc)

PR #213, opened 2024-08-01
OUT-675 Extra scrollbars on Tasks

aatbip:
### Changes

- [x] Fixes issues with scrollbar created due to changes made in #208


arpandhakal (changes requested):
While clicking the task card, the custom link covers more width than the task card. For example, if i click like 8,10px right from the task card, which is an empty space, it still opens the details page of that corresponding task card.

arpandhakal (approved):
LGTM!

PR #215, opened 2024-08-02
OUT-654 Improving task details opening experience

aatbip:
### Changes

- [x] Adds a `_StateOptimizer_` hoc which contains a memoized looping script to find the `currentTask`, `currentAssignee`, and `currentWorkflowState` that are essential for the details page

**Note:** 
Previously, we were performing this calculation (to find `currenTask`, `currentAssignee`, and `currentWorkflowState` which is a long scripting operation) on both `TaskEditor` and `Sidebar` which was creating a long scripting time. This doesn't ensure a drastic performance improvement, however a overload has been decreased because now we are performing the calculation in a memoized fashion inside `_StateOptimizer_`. Another place where such long scripting operation is performed is in the `TaskCard` and `FilterBar`, which is not possible to take away. 

**Possible improvement -**
It would be nice if we can send the `assigneeDetails` from the API response along with the array of `tasks`. This will make sure all the required data are available during the initial fetch request for the tasks.

### DIFF
--- a/src/app/api/activity-logs/[id]/route.ts
+++ b/src/app/api/activity-logs/[id]/route.ts
@@ -5,7 +5,7 @@ import { IdParams } from '@api/core/types/api'
 import { unstable_noStore as noStore } from 'next/cache'
 
 export const GET = async (req: NextRequest, { params: { id } }: IdParams) => {
-  noStore()
+  // noStore()
   const user = await authenticate(req)
 
   const activityLogger = new ActivityLogService(user)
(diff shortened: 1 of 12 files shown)


makeitraina (changes requested):
> Previously, we were performing this calculation (to find currenTask, currentAssignee, and currentWorkflowState which is a long scripting operation) on both TaskEditor and Sidebar which was creating a long scripting time

Are the calculations being run again and again? What is that is causing those operations to take a long scripting time?

Im requesting changes here because this feels like an over-engineered solution right now to a problem that is not well understood. See my question above. Until we have some more conviction that these scripts are actually taking a long time, I really don't think we should merge this.

PR #223, opened 2024-08-07
re/OUT 654 Improving task details opening experience

aatbip:
### Changes

- [x] Adds a loader on route transition event delays 

**Problem**
The opening experience was laggy because of the router transition delay time. When we click on any route, it often takes time to complete the transition. This delay caused the laggy effect. 
Next js 14 doesn't provide a way to listen to the router events such as `routeChangeStart`, which made it difficult to detect the start of a route transition to show a transition loader. See [this](https://github.com/vercel/next.js/discussions/42016) issue thread for details on workarounds being discussed. In [this](https://demo-nextjs-progressbar.vercel.app/) demo, we can see the progressive loader on route transition being implemented. This was necessary because route transition often cause delay in Next js. 

**Solution**
As a solution to this, this PR adds a loader on the route transition delays as an overlay on the top of the layout. This way, we will have two loaders i.e. 

- that tracks the fetch requests handled natively by the Next js loading boundry, and 
- that tracks the transition delay handled using custom component and supporting [library](https://github.com/samuhmatos/nextjs-progressloader/blob/main/docs/ProgressLoader.md)

This implementation makes sure that user experience is improved because user will now know that the click event is being processing and page transition is taking place.


src/hoc/CustomLink.tsx
```diff
@@ -42,13 +43,16 @@ export const CustomLink = ({
 
   const { pathname, token } = getUrl()
 
+  const [shouldPrefetch, setShouldPrefetch] = useState(false)
+
   const handleMouseEnter = useCallback(() => {
-    router.prefetch(`${pathname}?token=${token}`)
+    // router.prefetch(`${pathname}?token=${token}`)
+    setShouldPrefetch(true)
```
makeitraina, line 50: the dependency of of useCallback should be pathname and token right?

makeitraina (commented):
is this always going to show the progress load? sometimes the click is fast so it feels unnecessary to show this always right?

src/hoc/CustomLink.tsx
aatbip: The useCallback dependency array should include all variables that the callback depends on. In this case, the getUrl function depends on `href` and not directly on `pathname` and `token`. The `pathname` and `token` are derived from `href` within the getUrl function. Therefore, it is appropriate to use `href` in the dependency array. Including `pathname` and `token` would not be necessary since they are not direct dependencies but rather derived values.

rrojan (changes requested):
Didn't we also agree to only fetch changes for a single task using Supabase real time filters, instead of listening to changes to all tasks for details page?

aatbip:
After putting all the fetch requests in the layout to reduce multiple data fetching, we are getting a hydration error in the '/client' route. I have attempted to debug and track this issue, was able to understand a few things but couldn't resolve completely. To reproduce the issue, please go to the '/client' route in local environment as hydration error is hidden in the production. Will continue with further debugging. 

Screenshot from 2024-08-09 11-53-08

aatbip:
Continuing fresh in #237

PR #224, opened 2024-08-08
Hotfix: Increase client / company limit that is fed into filter

rrojan:
### Tasks

- N/A

### Changes

- [x] Increase the `limit` param for CopilotAPI `ListArgs` so that we filter total set of client / company records.


arpandhakal (approved):
LGTM

PR #225, opened 2024-08-08
OUT-692, OUT-693, OUT-694 Reduce spacing between assignees in drop-down on mobile experience

arpandhakal:
### Tasks

- [Reduce spacing between assignees in drop-down on mobile experience](https://linear.app/copilotplatforms/issue/OUT-692/reduce-spacing-between-assignees-in-drop-down-on-mobile-experience)
- [First assignee is always shown as selected in Assignee drop-down](https://linear.app/copilotplatforms/issue/OUT-693/first-assignee-is-always-shown-as-selected-in-assignee-drop-down)
- [Search text does not reset to default state when switching between tabs](https://linear.app/copilotplatforms/issue/OUT-694/search-text-does-not-reset-to-default-state-when-switching-between)

### Changes

- [x] overwritten mui's css to apply own custom css and removed autohighlight prop from autocomplete
- [x] added global mui theme for autocompletes throughtout the application to align with figma design
- [x] added clear value function in onBlur prop of the selector's autocomplete's textfield component  

### Testing Criteria

- [LOOM](https://www.loom.com/share/02a6db3045684b8d84d12715b16e7905)

### DIFF
--- a/src/theme/theme.ts
+++ b/src/theme/theme.ts
@@ -187,8 +187,8 @@ export const theme = createTheme({
     MuiAutocomplete: {
       styleOverrides: {
         option: {
-          height: '32px !important',
-          minHeight: '32px !important', //minHeight required here because in selector component, props className MuiAutocomplete-options overrides by applying minWidth : 48px in smaller screens
+          height: '32px',
+          minHeight: '32px', //minHeight required here because in selector component, props className MuiAutocomplete-options overrides by applying minWidth : 48px in smaller screens
         },
       },
     },
--- a/src/theme/theme.ts
+++ b/src/theme/theme.ts
@@ -187,8 +187,8 @@ export const theme = createTheme({
     MuiAutocomplete: {
       styleOverrides: {
         option: {
-          height: '32px',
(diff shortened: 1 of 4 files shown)


rrojan (changes requested):
Some minor improvement, which aren't necessary directly related to PR tasks:

 
image
In figma, the gap between assignee avatar and "Filter by" is 8px not 4.

image

In figma, only the square "Close" div is highlighted when hovering on close icon, but in our implementation, the whole selector is highlighted, which could lead to a confusing user experience

src/theme/theme.ts
```diff
@@ -183,4 +183,14 @@ export const theme = createTheme({
...
+    MuiAutocomplete: {
+      styleOverrides: {
+        option: {
+          height: '32px !important',
+          minHeight: '32px !important', //minHeight required here because in selector component, props className MuiAutocomplete-options overrides by applying minWidth : 48px in smaller screens
+        },
+      },
+    },
+  },
 })
```
aatbip, line 196: I think it's not a good idea to override the autocomplete style from the theme. We should be able to do this using css selectors.

aatbip (changes requested):

aatbip (approved):
lgtm

rrojan (approved):
Looks good, LGTM :trophy:

PR #227, opened 2024-08-08
OUT-683 Assignee shows with ellipse on client end on mobile

arpandhakal:
### Tasks

- [Assignee shows with ellipse on client end on mobile](https://linear.app/copilotplatforms/issue/OUT-683/assignee-shows-with-ellipse-on-client-end-on-mobile)

### Changes

- [x] applied custom breakpoints to utilize space in client mobile view.

### Testing Criteria

- [LOOM](https://www.loom.com/share/9b19c7e5e4a34ede81ca0a0c6bc06a0e)

### DIFF
--- a/src/components/cards/ClientTaskCard.tsx
+++ b/src/components/cards/ClientTaskCard.tsx
@@ -14,6 +14,7 @@ import { DueDateLayout } from '@/components/layouts/DueDateLayout'
 import { UrlObject } from 'url'
 import { CustomLink } from '@/hoc/CustomLink'
 import { getAssigneeName } from '@/utils/assignee'
+import { GetMaxAssigneeNameWidth } from '@/utils/getMaxAssigneeNameWidth'
 
 export const ClientTaskCard = ({
   task,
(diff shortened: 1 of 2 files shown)


src/components/cards/ClientTaskCard.tsx
```diff
@@ -124,12 +128,20 @@ export const ClientTaskCard = ({
...
+                    '@media (max-width: 400px)': {
+                      maxWidth: task.dueDate ? '80px' : '150px',
+                    },
+                    '@media (min-width: 400px) and (max-width: 500px)': {
+                      maxWidth: task.dueDate ? '120px' : '220px',
+                    },
+                    '@media (min-width: 500px) and (max-width: 600px)': {
+                      maxWidth: task.dueDate ? '200px' : '300px',
+                    },
                   }}
```
aatbip, line 141: The current implementation of multiple breakpoints for maxWidth in the Typography component seems to be overly complex and potentially difficult to maintain. To reduce the complexity, consider simplifying the breakpoints or using a utility function that calculates maxWidth based on the screen size and task.dueDate condition. This approach could make the code more readable and maintainable while still achieving the desired responsiveness.

I also think we don't need to specify both max-width and min-width here. We can just give max-width and rest will be taken care. Some of the breakpoints seems not be used.

aatbip (changes requested):

aatbip (approved):
work of art!

PR #235, opened 2024-08-12
OUT-722 Special character '&' is shown as '&' on task list page on client end

arpandhakal:
### Tasks

- [Special character '&' is shown as '&' on task list page on client end](https://linear.app/copilotplatforms/issue/OUT-722/special-character-and-is-shown-as-andamp-on-task-list-page-on-client)

### Changes

- [x] added a decoding function to decode encoded special characters in extractHtml function. 

### Testing Criteria

image


rrojan (approved):
LGTM :trophy: 

Using textarea for this is a reliable and solid approach, however using the underlying DOMParser API would've been a solid option too. [Ref](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString)

PR #237, opened 2024-08-12
OUT 654 Improving task details opening experience

aatbip:
### Changes

- [x] Adds a loader on route transition event delays 
- [x] Optimized fetch requests by shifting all the fetches into the layout. Now fetch is happening only during the first load in the layout page
- [x] Adds middleware to get the token into headers
- [x] Improves performance 

### Testing Criteria
- [x] [LOOM](https://www.loom.com/share/436c8aeac99149d09175ef2d9107aaea?sid=e9ea20d0-6d59-4c6d-882d-7f6213b38fe8) 
- [x] Try navigating to the task details page or try using filters and navigate to the task details page, you will see the top loader on top of the page and you will realize the improved performance. 

**Problem**
The opening experience was laggy because of the router transition delay time. When we click on any route, it often takes time to complete the transition. This delay caused the laggy effect. 
Next js 14 doesn't provide a way to listen to the router events such as `routeChangeStart`, which made it difficult to detect the start of a route transition to show a transition loader. See [this](https://github.com/vercel/next.js/discussions/42016) issue thread for details on workarounds being discussed. In [this](https://demo-nextjs-progressbar.vercel.app/) demo, we can see the progressive loader on route transition being implemented. This was necessary because route transition often cause delay in Next js. 

**Solution**
As a solution to this, this PR adds a loader on the route transition delays as an overlay on the top of the layout. This way, we will have two loaders i.e. 

- that tracks the fetch requests handled natively by the Next js loading boundry, and 
- that tracks the transition delay handled using custom component and supporting [library](https://github.com/samuhmatos/nextjs-progressloader/blob/main/docs/ProgressLoader.md)

This implementation makes sure that user experience is improved because user will now know that the click event is being processing and page transition is taking place.


makeitraina:
@aatbip where is the loom video showing the testing for this?

src/hoc/ForceDisableServerRender.tsx
```diff
@@ -0,0 +1,17 @@
...
+  //this state tracks if the component is in client side or not during hydration
+  const [isCLient, setIsClient] = useState(false)
+
+  useEffect(() => {
+    setIsClient(true)
+  }, [])
+
+  if (!isCLient) return null
+
+  return children
```
rrojan, line 16: I'm sure you're doing this for a valid reason.. but what is the reason behind checking `isClient` for a client component? Aren't client components only supposed to render on the client side after hydration? So by that logic what does `if (!isClient) return null` do?

src/middleware.ts
```diff
@@ -0,0 +1,10 @@
+import { NextResponse, NextRequest } from 'next/server'
+
+export function middleware(request: NextRequest) {
+  const token = request.nextUrl.searchParams.get('token')
+  const requestHeaders = new Headers(request.headers)
+  if (token) {
+    requestHeaders.set('clientToken', token)
```
rrojan, line 7: Naming this `clientToken`, presumably meaning something like "token that http client passes" might conflict with our understanding of user scopes (internalUser / client / company).

src/utils/sortTask.ts
```diff
@@ -1,7 +1,7 @@
 import { TaskResponse } from '@/types/dto/tasks.dto'
 
 export const sortTaskByDescendingOrder = (tasks: TaskResponse[]) => {
-  return tasks.sort((a, b) => {
+  return (tasks ?? []).sort((a, b) => {
```
rrojan, line 4: If `tasks` is typed as `TaskResponse[]`, why would we have to check for `tasks ?? []`? If this is meant for error handling, we should be better off throwing / catching errors here instead of rendering empty tasks imo.

package.json
```diff
@@ -48,6 +48,8 @@
     "http-status": "^1.7.4",
     "immutability-helper": "^3.1.1",
     "next": "^14.1.4",
+    "nextjs-progressloader": "^1.2.0",
+    "nprogress": "^0.2.0",
```
rrojan, line 52: Using `NProgress` as a direct dependency here seems unnecessary since it's only a peer dependency of `nextjs-progressloader`

rrojan (changes requested):

src/hoc/ForceDisableServerRender.tsx
aatbip: In NextJS, any component that is placed in the server component (such as page.tsx) are server rendered even though the component has `use client` directive. The `use client` directive simply means the component is a client component and thus we are allowed to use client side states, hooks, and effects in such a components. 

In the above case, the `ForceDisableServerRender` is used in the server component (page.tsx). Initially, NextJS renders the page.tsx server component in the server and thus this component also runs and renders in the server itself. However, initially in the server, since this is a client component, the `isClient` state would be undefined and this component would return `null` in the server rendering nothing at all. This approach is used to get rid of hydration errors. If we don't return `null` in the server then the server would still return `children` and pre-render the `children`, and then when the pre-rendered html tries to hydrate with the client side html there will be a mismatch scenario which would create the hydration error. Thus, this component would force stop pre-rendering in the server side to avoid hydration error.

src/utils/sortTask.ts
aatbip: fixed in the commits below

package.json
aatbip: removes in the commit below

src/hoc/ForceDisableServerRender.tsx
priosshrsth: Yeah client components are also rendered on the server, while the server components are rendered and run exclusively on the server.

rrojan: Since we are returning `null` instead of our component until hydration, wouldn't that degrade performance instead?

aatbip: There is no issue with performance because 
1. Even though we don't return `null`, the server would return the `children` and the hydration would still take place and would look like  ***server returned `children` X client side `children`***
2. If we return `null` in the server then the server would return `null` to the client and during hydration it would look like ***server returned `null` X client side `children`***

where X means "hydrates with". 

Client side `children` is returned in the client side after states and effects are run while server returned `children` is returned on the server side without running states and effects.  

cc @priosshrsth

makeitraina (commented):
One thing we should confirm is if there is actually any change in api calls made when you click on the details page or clients page.

To see this we can add logs to the functions that fetch data and then see if when you click into a details page (or hover for prefetch) do those functions get called again?

It looks like the main here is the progress-loader, but the real fix for this will be better suspense boundaries and improving the loading experience so that fetching data does not block the navigation.

src/app/client/page.tsx
```diff
@@ -42,38 +17,22 @@ async function getAssigneeList(token: string): Promise<IAssignee> {
...
-      <ClientSideStateUpdate
-        workflowStates={workflowStates}
-        tasks={tasks}
-        token={token}
-        assignee={assignee}
-        tokenPayload={tokenPayload}
-      >
-        <RealTime>
+      <ClientSideStateUpdate assignee={assignee}>
+        <ForceDisableServerRender>
```
makeitraina, line 28: this isn;t a great solution, it removes the benefit of server rendering and the nice experience we get from loading from the server

src/utils/sortTask.ts
```diff
@@ -1,6 +1,9 @@
 import { TaskResponse } from '@/types/dto/tasks.dto'
 
 export const sortTaskByDescendingOrder = (tasks: TaskResponse[]) => {
+  if (!tasks) {
+    throw new Error('Something went wrong!')
```
makeitraina, line 5: why throw an error? just return empty array right?

src/hoc/ForceDisableServerRender.tsx
priosshrsth: This is not an ideal solution. But I don't think it should have noticable performance impact. For the record @aatbip I think we can ignore the hyrdration error if we want. That won't show up in production. 

But this seems ok as well, because we are always showing empty message anyway, and the data are not empty in reality. So shouldn't have impact on seo as well. I assume the UI has some sort of loader though. Otherwise, like @rrojan mentioned, user would be flashed with blank UI which switcher to content all of a sudden.

aatbip:
Closing because continued in #246

PR #239, opened 2024-08-13
OUT-715 Unnecessary scrollbar is shown on task details page when task title is long

arpandhakal:
### Tasks

- [Unnecessary scrollbar is shown on task details page when task title is long](https://linear.app/copilotplatforms/issue/OUT-715/unnecessary-scrollbar-is-shown-on-task-details-page-when-task-title-is)

### Changes

- [x] min width of tapwrite (task description) in task details page is kept according to screen width sizes.

### Testing Criteria

- [LOOM](https://www.loom.com/share/e3a6c47851274eed932e3467d3f869d6?focus_title=1&muted=1&from_recorder=1)


src/app/globals.css
```diff
@@ -5,5 +5,17 @@
...
+@media (min-width: 768px) {
+  .tapwrite-details-page {
+    min-height: 55vh;
+  }
+}
+
+@media (min-width: 1200px) {
+  .tapwrite-details-page {
+    min-height: 65vh;
+  }
```
aatbip, line 20: Still seeing unnecessary scrollbar in some screen sizes. In the video below it shown for ~1330px where scrollbar is appearing. 
This approach is creating another problems. For example, I won't be able to click on the tapwrite to place the cursor from bottom of the page wince min-height is only 45px. Also, in the future milestone, we will be adding activity list which should appear after the tapwrite so adding various min-height would create problems there as well. 

Issue 1: Unnecessary scrollbar still exists on various screen sizes
[Screencast from 13-08-24 02:57:46 PM +0545.webm](https://github.com/user-attachments/assets/1229c189-1cb5-4a42-a056-d7ca92061c04)

Issue 2: I can't click on the tapwrite to place my cursor if height of tapwrite in less
[Screencast from 13-08-24 03:02:55 PM +0545.webm](https://github.com/user-attachments/assets/1ddbf010-b36d-4ca2-bbb7-74d828def948)

I recommend to think about various conditions when we are solving an issue rather than directly focusing on solving the issue.

aatbip (changes requested):

PR #241, opened 2024-08-13
OUT-732 Task Title cannot be removed all at once while editing in task description page.

arpandhakal:
### Tasks

- [Task Title cannot be removed all at once while editing in task description page.](https://linear.app/copilotplatforms/issue/OUT-732/task-title-cannot-be-removed-all-at-once-while-editing-in-task)

### Changes

- [x] added a validation in server component to not hit updateDetail api for title update
- [x] created useDebounce hook with cancel feature 
- [x] debounced the task title with error message and last title revert mechanism and cancel it if title is changed within 2000ms   

### Testing Criteria

- [LOOM](https://www.loom.com/share/de2405e7480f4fbba0ef742bec784516)


aatbip (changes requested):
I think the approach that we are building here can be drastically improved to provide at most user experience as the current implementation is producing some unintended behaviour. Copilot team has enforced that we can take hints from the Linear to build up on our implementation. In Linear, if task title is empty then auto update doesn't happen at all. So you see there are two flows in which the title updates can happen. 

1. Auto updates - This is the usual flow in the case `title` is not empty. 
2. Manual updates - This is our flow of interest where we will opt out of the auto updates and proceed with manual updates, if and only if `title` is empty. 

For example, if you check Linear, when `title` is empty, it doesn't auto save at all. It will throw and error if the `textfield` is blurred out, then it will re-establish the `title` with the old `title` prior to the error. 

In our case, we are auto updating everytime and reverting back to the same old `title`. In that way the user experience becomes very unusual with the below questions - 
1. Can't a user erase the `title` completely and sit back to think of a new title? 
2. Can't a user erase the first letter of the title? 

The video below demonstrate the UX - 
[Screencast from 14-08-24 12:54:57 PM +0545.webm](https://github.com/user-attachments/assets/e3f50a07-baaf-4072-94a9-771fe8a69b48)

I think the better approach would be to - 
1. Opt out of auto updates when title is empty. Don't re-render the `title` back if user wants to erase it. 
2. Show error and revert back the `title` if user navigates to different page or blurs the `textfield`

arpandhakal:
@aatbip thank you so much for this thorough review, I thought about this issue. Pushed some new changes to make the flow look like linear. Regarding the error message, Linear uses a chip or a snackbar. I think we are waiting for the design on building a chip right now. So I have used the helperText from mui just like we did on Task creation.

rrojan (approved):
While thorough testing, i could see the task title flicker into the previous data for a split second if i was only using single character task titles and edit -> blurring them very quickly. This doesn't seem very likely to occur in real use cases.

LGTM :trophy:

aatbip (approved):
lgtm

PR #242, opened 2024-08-14
OUT-734 Hide sensitive error messages from production

arpandhakal:
### Tasks

- [Hide sensitive error messages from production](https://linear.app/copilotplatforms/issue/OUT-734/hide-sensitive-error-messages-from-production)

### Changes

- [x] added a condition to hide error message other than tokenError message in the error suspense component

### Testing Criteria

- [LOOM](https://www.loom.com/share/c3e0e3f23e1f43c0a5cdc3a9a93078b1)

### Notes

- In the loom video, I first created a fetch failed error by inserting random string in VERCEL_ENV variable, and for the second time I changed the VERCEL_ENV variable to 'production' which still causes fetch failed error locally, but the error is masked here.

### DIFF
--- a/src/app/error.tsx
+++ b/src/app/error.tsx
@@ -1,14 +1,17 @@
 'use client'
 
+import { PrimaryBtn } from '@/components/buttons/PrimaryBtn'
 import { isProd } from '@/config'
+import { Box, Stack } from '@mui/material'
+import { useRouter } from 'next/navigation'
 import React from 'react'
 
-const ClientErrorBoundary = ({ error }: { error: Error & { digest: string } }) => {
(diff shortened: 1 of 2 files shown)


aatbip (changes requested):
I think it is a good idea to make a way for users to reset from the error states. Can we add a `reset` prop on this component and a button to initiate the `reset` that is available in the error components?

rrojan:
> I think it is a good idea to make a way for users to reset from the error states. Can we add a `reset` prop on this component and a button to initiate the `reset` that is available in the error components?

Yeah, a "Try again" type of button would make sense.

rrojan (approved):
LGTM :trophy:

aatbip (approved):
lgtm

PR #245, opened 2024-08-15
OUT-677 | In-product notification & Task do not disappear from client end when company is deleted/unassigned

rrojan:
### Tasks

- [OUT-677 | In-product notification & Task do not disappear from client end when company is deleted/unassigned](https://linear.app/copilotplatforms/issue/OUT-677/in-product-notification-and-task-do-not-disappear-from-client-end-when)

### Changes

- [x] Add core logic in Notification Service to bulk read notifications for a client / company
- [x] Add core logic in Tasks Service to bulk delete all tasks for a particular client / IU / company 
- [x] Implement bulk delete and mass notification read in webhook event for `entity.deleted`
- [x] Remove transacation from tasks.service because it was throwing a cryptic error occasionally 

### Testing Criteria

- [x] [LOOM](https://www.loom.com/share/88d0a2f1fdfd44dc9df420623550c2af?sid=85122500-c63c-40c7-a979-6d21d0819700)

### DIFF
--- a/src/app/api/notification/notification.service.ts
+++ b/src/app/api/notification/notification.service.ts
@@ -7,6 +7,7 @@ import { CompanyResponse, CopilotUser, MeResponse, NotificationCreatedResponse }
 import { getEmailDetails, getInProductNotificationDetails } from '@api/notification/notification.helpers'
 import APIError from '@api/core/exceptions/api'
 import httpStatus from 'http-status'
+import Bottleneck from 'bottleneck'
 
 export class NotificationService extends BaseService {
   async create(action: NotificationTaskActions, task: Task, disable: { email: boolean } = { email: false }) {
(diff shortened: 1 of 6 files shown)


src/app/api/notification/notification.service.ts
```diff
@@ -139,6 +140,42 @@ export class NotificationService extends BaseService {
...
+    if (userType === AssigneeType.company) {
+      // In this case, we need to find out all the clients inside this deleted company, then pass that array to markAllAsReadForClients
+      const clients = await copilot.getCompanyClients(userId)
+      return await this.markAllAsReadForClients(clients.map((client) => client.id))
+    }
+
+    throw new APIError(httpStatus.INTERNAL_SERVER_ERROR, 'Failed to parse assignee')
+  }
+
+  async markAllAsReadForClients(clientIds: string[]) {
```
sazanrjb, line 161: Since we have used this in a single method above, this should better be defined as `private`

Also, copilot instance can be passed as a param?

src/app/api/tasks/tasks.service.ts
```diff
@@ -275,12 +275,10 @@ export class TasksService extends BaseService {
     }
     const labels = tasks.map((task) => task.label)
 
-    await this.db.$transaction([
-      this.db.task.deleteMany({
-        where: { assigneeId, assigneeType, workspaceId: this.user.workspaceId },
-      }),
-      this.db.label.deleteMany({ where: { label: { in: labels } } }),
-    ])
+    await this.db.task.deleteMany({
```
sazanrjb, line 278: any reason for removing transaction?

rrojan: It was throwing an obscure error (only at times)... I couldn't find a solution for it so after multiple tries, the transaction has been removed.

Removing it altogether isn't the best way to approach this, I will look into this in the future.

src/app/api/notification/notification.service.ts
rrojan: I actually removed the `private` identifier on this one since this method is very handy for a number of use cases - I want it to be accessible as a public method.

rrojan: In that case, passing `copilot` as param would be counterintuitive as well :bow:

sazanrjb (approved):

PR #250, opened 2024-08-15
OUT-642 | Prevent invalid / missing token to propagate over app flow

rrojan:
### Tasks

- [Error: Unable to authorize Copilot SDK.](https://linear.app/copilotplatforms/issue/OUT-642/error-unable-to-authorize-copilot-sdk)

### Changes

- [x] Token wasn't being parsed in Client list page, so empty token was being propagated into lots of functions, triggering a whole chain of errors.

image


arpandhakal (approved):
Is this error only coming from client view? LGTM

PR #255, opened 2024-08-18
OUT-678 Unnecessary scrollbar is shown on task details page when opening task from notifications.

arpandhakal:
### Tasks

- [Unnecessary scrollbar is shown on task details page when opening task from notifications.](https://linear.app/copilotplatforms/issue/OUT-678/unnecessary-scrollbar-is-shown-on-task-details-page-when-opening-task)

### Changes

- [x] applied a component mounted check for customScrollbar component so that default styles would not override if the component is not mounted

### Testing Criteria

- [LOOM](https://www.loom.com/share/b47d33478ed0485eb109f22e3ead8798)

### DIFF
--- a/src/hoc/CustomScrollbar.tsx
+++ b/src/hoc/CustomScrollbar.tsx
@@ -1,26 +1,36 @@
 'use client'
 
-import { CSSProperties, ReactNode } from 'react'
+import { CSSProperties, ReactNode, useEffect, useState } from 'react'
 import Scrollbars from 'react-custom-scrollbars'
 
 export const CustomScrollbar = ({ children, style }: { children: ReactNode; style: CSSProperties }) => {
+  const [mounted, setMounted] = useState(false)
+
+  useEffect(() => {
+    setMounted(true)
+  }, [])
+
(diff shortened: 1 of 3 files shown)


rrojan (commented):
LGTM
Hiding this component until mounted does seem to be the solid approach here :rocket:

rrojan (approved):

saugatmaharjan (approved):

src/hoc/CustomScrollbar.tsx
```diff
@@ -1,23 +1,33 @@
...
+            {...props}
+            className="thumb"
+            style={{
+              borderRadius: '6px',
+              backgroundColor: 'rgba(0, 0, 0, 0.5)',
+              ...style,
+            }}
+          />
+        )
+      }}
```
aatbip, line 30: We can use the `CustomScrollbar` here by making the `children` prop as optional. So that, 
`if (!children) return <Scrollbars ... />`

aatbip (changes requested):

PR #260, opened 2024-08-20
OUT-718 Some clients are not shown in Assignee drop-down

aatbip:
### Changes

- [x] Changes limit from 2000 to 10000 in `getAssigneeList` at `setDebouncedFilteredAssignee`

### Testing Criteria
- [x] [LOOM](https://www.loom.com/share/d1cd236126d2435e9b9af12b3d97b66f?sid=2e961417-36d8-49ab-8135-36fe8561c03a)

### DIFF
--- a/src/utils/users.ts
+++ b/src/utils/users.ts
@@ -56,7 +56,7 @@ export const setDebouncedFilteredAssignees = (
   }
   const newTimeoutId = setTimeout(async () => {
     setLoading(true)
-    const newAssignees = await getAssigneeList(z.string().parse(token), newInputValue, 2000, '0') // 2000 is hardcoded for now
+    const newAssignees = await getAssigneeList(z.string().parse(token), newInputValue, 5000, '0') // 2000 is hardcoded for now
     setAssigneeState(addTypeToAssignee(newAssignees))
     setLoading(false)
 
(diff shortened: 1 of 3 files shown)


src/utils/users.ts
```diff
@@ -56,7 +56,7 @@ export const setDebouncedFilteredAssignees = (
   }
   const newTimeoutId = setTimeout(async () => {
     setLoading(true)
-    const newAssignees = await getAssigneeList(z.string().parse(token), newInputValue, 2000, '0') // 2000 is hardcoded for now
+    const newAssignees = await getAssigneeList(z.string().parse(token), newInputValue, 10000, '0') // 2000 is hardcoded for now
```
sazanrjb, line 59: will it hit any perfomance issue?

the comment still says 2000

aatbip: Having 10000 items in the payload will definitely increase the response time but since this is suppose to filter by keyword we won't have lot of payload so performance should be fine.

Removed comment in the commit below.

sazanrjb: Approved but please test this is load testing branch

sazanrjb (approved):
Lets test in load testing branch

PR #261, opened 2024-08-20
OUT-750 "Filter by assignee" drop-down doesn't work as expected

arpandhakal:
### Tasks

- ["Filter by assignee" drop-down doesn't work as expected](https://linear.app/copilotplatforms/issue/OUT-750/filter-by-assignee-drop-down-doesnt-work-as-expected)

### Changes

- [x] added a userType for search functionality hits the copilot getUsers method with a filter of userType : client_companies, ius according to the selected tab

### Testing Criteria

- [LOOM](https://www.loom.com/share/7da487a2448c4252b736f6bb617ee0f4)


src/app/api/users/users.controller.ts
```diff
@@ -8,14 +8,13 @@ export const getUsers = async (req: NextRequest) => {
   const user = await authenticate(req)
   const usersService = new UsersService(user)
   const keyword = req.nextUrl.searchParams.get('search')
-
   const rawLimit = req.nextUrl.searchParams.get('limit')
+  const userType = req.nextUrl.searchParams.get('userType') || undefined
```
sazanrjb, line 12: isnt it undefined if not found, already?

src/app/api/users/users.service.ts
```diff
@@ -65,10 +66,25 @@ class UsersService extends BaseService {
    * @param keyword
    * @returns
    */
-  async getFilteredUsersStartingWith(keyword: string, limit?: number, nextToken?: string) {
+  async getFilteredUsersStartingWith(keyword: string, limit?: number, nextToken?: string, userType?: string) {
     const filterByKeyword = (users: FilterableUser[]) => filterUsersByKeyword(users, keyword)
-
     const { internalUsers, clients, companies } = await this.getGroupedUsers(limit || 500, nextToken)
+
```
rrojan, line 72: We are still hitting all 3 endpoints for IU, client and company, even if we only want IU for `FilterOptionsKeywords.TEAM`, and only client and company for `FilterOptionsKeywords.CLIENTS`

rrojan (changes requested):

src/app/api/users/users.controller.ts
arpandhakal: If not sent, the `searchparams.get('userType')` will be null and not undefined. Since this is an optional parameter, i think we need to change it to undefined it we receive null. The `keyword`  is not changed into undefined because it was made clear to call the function only if keyword is not null : 
```
  const users = await (keyword
    ? usersService.getFilteredUsersStartingWith(keyword, userType, limit, nextToken)
    : usersService.getGroupedUsers(limit))
```

sazanrjb (approved):

aatbip (dismissed):
lgtm

rrojan (approved):
LGTM :trophy:

PR #264, opened 2024-08-21
OUT-760, OUT-767, OUT-761, OUT-758 Task description is out of sync on details page vs. list view

aatbip:
### Changes

- [x] Changes real time update approach
- [x] Updates `tasks` array on each real time updates
- [x] Refactored lazy task fetchers
- [x] Removes `TopLoader` for client page completely because the `Mark done` button was working unpredictably in the client page where it sometimes shows top loader and sometimes doesn't show. Should work further to introduce a different way for clients.

### Testing criteria
- OUT-758 When starting with empty state, IU & CU can’t see any new tasks assigned to them in real time - [LOOM](https://www.loom.com/share/23f02b47cb0f4683bbfafc36ab644e3d?sid=fee9b4a2-e266-4c5c-80f6-190d08a1ea64)
- OUT-761 Unnecessary progress bar after clicking “Mark done” button - [LOOM](https://www.loom.com/share/c964643e13154e7292fd8a17a27794b5?sid=01a5a825-80e4-4228-b6f4-a5434e843ff6)
- OUT-760 Task description is out of sync on details page vs. list view - [LOOM](https://www.loom.com/share/c90056f91c424d7db5582f7d7a4e2d97?sid=ea4fbfd3-e3fc-4edf-b529-7ff2b5ab1b6c)


src/components/TopLoader.tsx
```diff
@@ -1,8 +1,11 @@
 'use client'
+import { usePathname } from 'next/navigation'
 import { ProgressLoader } from 'nextjs-progressloader'
 import { Suspense } from 'react'
 
 export const ProgressLoad = () => {
+  const pathname = usePathname()
+  if (pathname.includes('client')) return null
```
rrojan, line 8: Do we not need progress loader in client list page at all?

Also as detail route for client is `/detail/:id/cu`, i think we will get progress loader going from detail -> list, but not list -> detail

src/redux/features/taskBoardSlice.tsx
```diff
@@ -57,10 +55,7 @@ const taskBoardSlice = createSlice({
       let taskToUpdate = state.tasks.find((task) => task.id === action.payload.taskId)
       if (taskToUpdate) {
         taskToUpdate.workflowStateId = action.payload.targetWorkflowStateId
-        const updatedTasks = sortTaskByDescendingOrder([
-          ...state.tasks.filter((task) => task.id !== action.payload.taskId),
-          taskToUpdate,
-        ])
+        const updatedTasks = [...state.tasks.filter((task) => task.id !== action.payload.taskId), taskToUpdate]
```
rrojan, line 58: I think this makes it so that when a task is updated it gets pushed to the last element of tasks array. Where do we do sorting after this?

src/hoc/RealTime.tsx
```diff
@@ -34,58 +32,37 @@ export const RealTime = ({ children }: { children: ReactNode }) => {
...
-        supabase.removeChannel(channel)
-      }
-    } else {
-      const channel = supabase
-        .channel('realtime tasks')
-        .on('postgres_changes', { event: '*', schema: 'public', table: 'Tasks' }, handleTaskRealTimeUpdatesInRoot)
-        .subscribe()
+    const channel = supabase
+      .channel('realtime tasks')
+      .on('postgres_changes', { event: '*', schema: 'public', table: 'Tasks' }, handleTaskRealTimeUpdates)
```
rrojan, line 59: This is a major security concern because we are getting CRUD events for all tasks, not just tasks that IU has access to. 

E.g. if I had two companies Apple and Samsung, Samsung will be subscribed to changes for tasks in Apple as well.

We should at least add filters for workspaceId.

```js
.on('postgres_changes', { event: '*', schema: 'public', table: 'Tasks', filter: `workspaceId=eq.${tokenPayload.workspaceId}` }, handleTaskRealTimeUpdates)
```

Other required filters should be:
- Adding a filter for current user's companyAccessList if user is IU (using `in.` filter)
- Only subscribing tasks where `assigneeId=eq.${tokenPayload.clientId}` for client users.

These later ones can be added and dealt with in later tickets since it requires some additional logic for isClientAccessRestricted, etc, and we are in a time crunch right now.

rrojan (changes requested):

src/app/_fetchers/AssigneeFetcher.tsx
```diff
@@ -37,5 +37,9 @@ const fetchAssignee = async (
 
 export const AssigneeFetcher = async ({ token, userType, viewSettings }: Props) => {
   const assignableUsersWithType = addTypeToAssignee(await fetchAssignee(token, userType))
-  return <ClientSideStateUpdate assignee={assignableUsersWithType} viewSettings={viewSettings} />
+  return (
+    <ClientSideStateUpdate assignee={assignableUsersWithType} viewSettings={viewSettings}>
+      {null}
+    </ClientSideStateUpdate>
```
rrojan, line 43: This doesn't hamper anything but it still looks weird :sweat_smile:

src/components/TopLoader.tsx
aatbip: This is an immediate solution as we are approaching release. The `nextjs-progressloader` component has a weird issue where it is sometimes showing loader while clicking the mark done button that's why this is only included in the list page for clients.

src/redux/features/taskBoardSlice.tsx
aatbip: We are doing sorting in the TaskBoard component for each workflowState columns.

src/hoc/RealTime.tsx
aatbip: I think this thing can be approached separately after the release because there are a lot of open holes for optimizations which may block the release. May be creating a separate ticket would be better.

src/app/_fetchers/AssigneeFetcher.tsx
aatbip: Would you like me to revert this to what it was before? This was just a personal preference.

src/components/TopLoader.tsx
rrojan: Got it... since next 13.1 or so we can't listen to route events so fixing it some other way should be tricky

src/hoc/RealTime.tsx
rrojan: Could you still please add the workspaceId filter

rrojan: Other filters can wait but this is a major one, especially since the code is open sourced

rrojan (approved):
Awesome work. LGTM :trophy:

PR #265, opened 2024-08-21
OUT-755, OUT-763 On task creation modal, Assignee drop-down shows a "Required" error even after assignee is selected

arpandhakal:
### Tasks

- [On task creation modal, Assignee drop-down shows a "Required" error even after assignee is selected](https://linear.app/copilotplatforms/issue/OUT-755/on-task-creation-modal-assignee-drop-down-shows-a-required-error-even)
- [Simultaneous task creation in very less interval of time causes an error.](https://linear.app/copilotplatforms/issue/OUT-763/simultaneous-task-creation-in-very-less-interval-of-time-causes-an)

### Changes

- [x] filter by assignee options not loading at first try when the page loads
- [x]  checked if there is assignee filter options while clearing the create task states.
- [x]  viewmode settings not being updated after adding suspense for assignee fetching.
- [x] also applied changes of OUT-740 which was lost due to merging

### Testing Criteria

- [LOOM](https://www.loom.com/share/245463dfa3ab463c928f2d121d1f19a7)

### DIFF
--- a/src/app/_fetchers/AssigneeFetcher.tsx
+++ b/src/app/_fetchers/AssigneeFetcher.tsx
@@ -5,13 +5,19 @@ import { apiUrl } from '@/config'
 import { MAX_FETCH_ASSIGNEE_COUNT } from '@/constants/users'
 import { addTypeToAssignee } from '@/utils/addTypeToAssignee'
 import { ClientSideStateUpdate } from '@/hoc/ClientSideStateUpdate'
+import { CreateViewSettingsDTO } from '@/types/dto/viewSettings.dto'
 
 interface Props {
   token: string
+  viewSettings?: CreateViewSettingsDTO
   userType?: UserType
 }
 
-const fetchAssignee = async (token: string, userType?: UserType): Promise<IAssignee> => {
+const fetchAssignee = async (
+  token: string,
+
+  userType?: UserType,
+): Promise<IAssignee> => {
   if (userType === UserType.CLIENT_USER) {
     const res = await fetch(`${apiUrl}/api/users/client?token=${token}&limit=${MAX_FETCH_ASSIGNEE_COUNT}`, {
       next: { tags: ['getAssigneeList'] },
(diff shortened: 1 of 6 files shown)


saugatmaharjan (approved):

src/app/_fetchers/AssigneeFetcher.tsx
```diff
@@ -29,8 +35,7 @@ const fetchAssignee = async (token: string, userType?: UserType): Promise<IAssig
   return (await res.json()).users as IAssignee
 }
 
-export const AssigneeFetcher = async ({ token, userType }: Props) => {
+export const AssigneeFetcher = async ({ token, userType, viewSettings }: Props) => {
   const assignableUsersWithType = addTypeToAssignee(await fetchAssignee(token, userType))
-
-  return <ClientSideStateUpdate assignee={assignableUsersWithType} />
+  return <ClientSideStateUpdate assignee={assignableUsersWithType} viewSettings={viewSettings} />
```
priosshrsth, line 40: I don't understand this. viewSettings is being fetched in page itself, isn't it? And so it should have been already set there.

src/app/ui/Modal_NewTaskForm.tsx
```diff
@@ -39,26 +39,25 @@ export const ModalNewTaskForm = ({
...
+              title,
+              body: description,
+              workflowStateId,
+              assigneeType,
+              assigneeId,
+              dueDate: formattedDueDate,
+            }
+
+            store.dispatch(clearCreateTaskFields({ isFilterOn: !!filterOptions[FilterOptions.ASSIGNEE] }))
+            const createdTask = await handleCreate(token as string, CreateTaskRequestSchema.parse(payload))
```
priosshrsth, line 54: Can we extract this in its own function and pass it as props?

priosshrsth (changes requested):

src/app/_fetchers/AssigneeFetcher.tsx
arpandhakal: The problem here is, viewsettings is updated in page itself which has dependencies in the filter by assignee button list too. For example, if client tasks tab is saved in viewsetting for a user, the filter by assignee button should show only client and company as options. Viewsettings being updated way before assignee list gets updated causes all the assigne to show in the filter by assignee options regardless the tab saved in viewsettings. For this purpose, viewSettings needs to be updated once agains when assignee list gets updated.

priosshrsth (approved):

PR #268, opened 2024-08-21
OUT-748 | Tasks & notifications don't appear & disappear in real-time on company assignment/unassignment + Fix Notification flow for task update

rrojan:
### Tasks

- [OUT-748 | Tasks & notifications don't appear & disappear in real-time on company assignment/unassignment](https://linear.app/copilotplatforms/issue/OUT-748/tasks-and-notifications-dont-appear-and-disappear-in-real-time-on)
- Hotfix: edge case while editing task where notification was triggered

### Changes

- [x] Fixed the entire company flow to react based on `client.updated` instead of `company.deleted` since that caused a lot of side effects
- [x] Massive refactor of webhooks logic 

### Testing Criteria

- [x] Tested locally changing db to prod and with ngrok webhook

### DIFF
--- a/src/app/api/webhook/webhook.service.ts
+++ b/src/app/api/webhook/webhook.service.ts
@@ -141,14 +141,23 @@ class WebhookService extends BaseService {
     if (prevCompanyId === newCompanyId || !prevCompanyId) return
 
     // Only delete prev notifications if client had a valid company before
-    const prevCompany = await this.copilot.getCompany(prevCompanyId)
-    const newCompany = await this.copilot.getCompany(newCompanyId)
+    let prevCompany, newCompany
+    try {
+      prevCompany = await this.copilot.getCompany(prevCompanyId)
+    } catch (e: unknown) {
+      prevCompany = null
+    }
+    try {
+      newCompany = await this.copilot.getCompany(newCompanyId)
+    } catch (e: unknown) {
+      newCompany = null
+    }
 
-    if (prevCompany.name === '' && newCompany.name !== '') {
+    if (!prevCompany?.name && newCompany?.name) {
       await this.handleCompanyAssignment(clientId)
-    } else if (newCompany.name === '' && prevCompany.name !== '') {
+    } else if (!newCompany?.name && prevCompany?.name) {
       await this.handleCompanyUnassignment(clientId, prevCompanyId)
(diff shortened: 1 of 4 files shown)


src/app/api/webhook/webhook.service.ts
```diff
@@ -120,53 +140,82 @@ class WebhookService extends BaseService {
...
+      newCompany = await this.copilot.getCompany(newCompanyId)
+    } catch (e: unknown) {
+      newCompany = null
+    }
+
+    if (!prevCompany?.name && newCompany?.name) {
+      await this.handleCompanyAssignment(clientId)
+    } else if (!newCompany?.name && prevCompany?.name) {
+      await this.handleCompanyUnassignment(clientId, prevCompanyId)
+    } else if (prevCompany?.name && newCompany?.name) {
```
sazanrjb, line 162: what if the prev company and new company are same?

src/app/api/webhook/webhook.service.ts
```diff
@@ -120,53 +140,82 @@ class WebhookService extends BaseService {
...
+    } catch (e: unknown) {
+      prevCompany = null
+    }
+    try {
+      newCompany = await this.copilot.getCompany(newCompanyId)
+    } catch (e: unknown) {
+      newCompany = null
+    }
+
+    if (!prevCompany?.name && newCompany?.name) {
```
sazanrjb, line 158: Can a company name be empty? 
Just asking because if prev company if is falsy, we are returning above.

src/app/api/webhook/webhook.service.ts
sazanrjb, line 162: what if prev company name and new company name are same?

src/app/api/webhook/webhook.service.ts
rrojan: Yes, in case of individual clients a unique company is generated for them.
This company returns a result in their retrieve company endpoint, and the name is an empty string.

The above snippet checks for both falsy prevCompany as well as empty string in prevCompany.name.

src/app/api/webhook/webhook.service.ts
rrojan: It doesn't matter because this doesn't check equality, only that both company names aren't falsey.

At the top of this method I check equality using company id instead.

PR #270, opened 2024-08-21
OUT-754 When selected assignee in Filter by drop-down is deleted, Board/List view shows blank until I click the tab

arpandhakal:
### Tasks

- [When selected assignee in Filter by drop-down is deleted, Board/List view shows blank until I click the tab](https://linear.app/copilotplatforms/issue/OUT-754/when-selected-assignee-in-filter-by-drop-down-is-deleted-boardlist)

### Changes

- [x] added a check if client was deleted while updating view settings

### Testing Criteria

- [LOOM](https://www.loom.com/share/f4a7650571f14ea7beb254d4b506f739)

### DIFF
--- a/src/redux/features/taskBoardSlice.tsx
+++ b/src/redux/features/taskBoardSlice.tsx
@@ -4,7 +4,7 @@ import { WorkflowStateResponse } from '@/types/dto/workflowStates.dto'
 import { TaskResponse } from '@/types/dto/tasks.dto'
 import { AssigneeType, FilterByOptions, FilterOptions, IAssigneeCombined, IFilterOptions, View } from '@/types/interfaces'
 import { ViewMode } from '@prisma/client'
-import { CreateViewSettingsDTO } from '@/types/dto/viewSettings.dto'
+import { CreateViewSettingsDTO, FilterOptionsType } from '@/types/dto/viewSettings.dto'
 import { sortTaskByDescendingOrder } from '@/utils/sortTask'
 
 interface IInitialState {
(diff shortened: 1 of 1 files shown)


src/redux/features/taskBoardSlice.tsx
```diff
@@ -71,6 +71,12 @@ const taskBoardSlice = createSlice({
     },
     setViewSettings: (state, action: { payload: CreateViewSettingsDTO }) => {
       state.view = action.payload.viewMode
+      if (action.payload.filterOptions.assignee) {
+        const assigneeCheck = state.assignee.find((assignee) => assignee.id == action.payload.filterOptions.assignee)
+        if (!assigneeCheck) {
+          action.payload.filterOptions.assignee = ''
+        }
+      }
```
saugatmaharjan, line 79: I am unaware of the context here but looking at the code, feels like this should be handled in separate action, if the logic is within the context LGTM.

arpandhakal: I think if we handled this outside the redux slice, there will be extra checks and fetches. So I think it is better to handle this in the taskBoardSlice itself because other states are easily accessible since filterOptions state depends on assigneeList. 

cc : @aatbip @rrojan

rrojan: @arpandhakal This seems like a case we could frequently come across in the future. It makes sense to refactor this to a separate action imo too.

saugatmaharjan: @arpandhakal i make sense to keep in the slice itself, I'm suggesting only a different action within the slice

arpandhakal: sure. thanks. I misunderstood 😅

rrojan (approved):
LGTM :trophy:

PR #274, opened 2024-08-22
OUT-766 | Opening task from notification center shows blank task details

rrojan:
### Tasks

- [OUT-766 | Opening task from notification center shows blank task details](https://linear.app/copilotplatforms/issue/OUT-766/opening-task-from-notification-center-shows-blank-task-details)

### Changes

- [x] Add `DetailStateUpdate` component to fetch context when redirected from notifications CTA button
- [x] Implement `DetailStateUpdate` instead of normal `ClientStateUpdate` for details page. 

### Testing Criteria

- [x] [LOOM](https://www.loom.com/share/d21b7e6d9e7d4d74b9976e232440aac5?sid=f74fab6d-8cd3-41d1-936a-7c8d4a7217b9)

### DIFF
--- a/src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx
+++ b/src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx
@@ -9,7 +9,7 @@ interface DetailStateUpdateProps {
   children: React.ReactNode
 }
 
-export const DetailStateUpdate = async ({ isRedirect, token, tokenPayload, children }: any) => {
+export const DetailStateUpdate = async ({ isRedirect, token, tokenPayload, children }: DetailStateUpdateProps) => {
   if (!isRedirect) {
     return (
       <ClientSideStateUpdate token={token} tokenPayload={tokenPayload}>
--- a/src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx
+++ b/src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx
@@ -3,7 +3,7 @@ import { getAllTasks, getAllWorkflowStates, getViewSettings } from '@/app/page'
 import { Token } from '@/types/common'
 
 interface DetailStateUpdateProps {
-  isRedirect?: 'true'
+  isRedirect?: boolean
   token: string
   tokenPayload: Token | null
   children: React.ReactNode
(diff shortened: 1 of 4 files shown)


src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx
```diff
@@ -0,0 +1,39 @@
+import { ClientSideStateUpdate } from '@/hoc/ClientSideStateUpdate'
+import { getAllTasks, getAllWorkflowStates, getViewSettings } from '@/app/page'
+import { Token } from '@/types/common'
+
+interface DetailStateUpdateProps {
+  isRedirect?: 'true'
```
Comment (saugatmaharjan) [inline src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx:6]: any specific reason to use string `true` @rrojan

Comment (rrojan) [reply src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx:6]: Switched to use `&isRedirect` instead of `&isRedirect=true`

src/utils/redirect.ts
```diff
@@ -6,7 +6,7 @@ import { UserType } from '@/types/interfaces'
 export const redirectIfTaskCta = (searchParams: Record<string, string>) => {
   const taskId = z.string().safeParse(searchParams.taskId)
   if (taskId.data) {
-    redirect(`${apiUrl}/detail/${taskId.data}/iu?token=${z.string().parse(searchParams.token)}`)
+    redirect(`${apiUrl}/detail/${taskId.data}/iu?token=${z.string().parse(searchParams.token)}&isRedirect=true`)
```
saugatmaharjan: what happens if parse fails here @rrojan ?

rrojan: We `console.error` a ZodError in logs and user gets a properly buillt error response, which is handled by `withErrorHandler` HOF in `src/app/core/utils/withErrorHandler.ts`

rrojan: But this case should almost never happen, since this function is triggered by an iframe path by Copilot platform and the platform always sends a token with every request

src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx
Comment (saugatmaharjan) [reply src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx]: I used to use `query-string` from npm to parse the search param but looks like that won't work here as well, `query-string` accepts raw `searchParam`.

However, my actual concerned is  component `DetailStateUpdate` defining prop as `isRedirect?: 'true'`. this couples the component to parent component, rather parent component could add the logic to send `boolean`, search param issue should not be issue of `DetailStateUpdate` component IMO

src/utils/redirect.ts
saugatmaharjan: so you mean if parse fails zod throws an exception. thanks for explanation.

saugatmaharjan (approved):

PR #275, opened 2024-08-22
OUT-767 When IU opens Tasks app and changes the tab or view(board/list), Tasks app refreshes

arpandhakal:
### Tasks

- [When IU opens Tasks app and changes the tab or view(board/list), Tasks app refreshes](https://linear.app/copilotplatforms/issue/OUT-767/when-iu-opens-tasks-app-and-changes-the-tab-or-viewboardlist-tasks-app)

### Changes

- [x] removed revalidateTag function to call getViewSettings api after updating view settings. The revalidateTag was unnecessary because view settings was stored on redux store on view settings update. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/55ab67a65ba94536b87ca3e457b6c088)

### Notes 

update(Aug 28) : The method we implemented here is completely different. Instead of dealing with cache, we decided to make a temp redux state for viewSettings and only update it while the user updates the components in the filterBar, apply a check for temp view settings for values in components of filterBar. This way while the app reloads for the first time, we use original viewSettings state from redux on our values, and if user updates something, then we use tempViewSettings values on filter bar component. This way, our non-outdated value persist throughout the lifecycle of the application flow. Instead of using hacky ways to disable caching which potentially had effects on optimization we decided to use this.


makeitraina (approved):
in pr description can you explain why the revalidateTag is not needed anymore.

aatbip (changes requested):
I see revalidateTag being removed but getViewSettings called again is not visible in the file changes. Are you sure it is called? 

Also if possible can we try the approach of optimistic updates over here? Since we are already launched I think it is worth trying a better approach for such small task so that it may give us a more solid idea for API updates and revalidation moving forward. May be we can also.employ react query or ssr, make their setup and then update it optimistically while returning back if error are detected rather than waiting for API call to complete in order to revalidate, sounds promising to me.


This can pave a solid pathway. 🔥

saugatmaharjan (dismissed):

aatbip:
There are some inconsistencies following the current implementation. Please kindly check it. Below loom video demonstrates the issues - 
https://www.loom.com/share/387fbd68fa564845bf2919cd15f0e4a8?sid=e7fcc492-fcca-4213-a631-09cb430b99dc

makeitraina:
> There are some inconsistencies following the current implementation. Please kindly check it. Below loom video demonstrates the issues - https://www.loom.com/share/387fbd68fa564845bf2919cd15f0e4a8?sid=e7fcc492-fcca-4213-a631-09cb430b99dc

thanks for testing this properly. This definitely shows that we need to fix this before merging.

src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -84,6 +84,7 @@ export default async function TaskDetailPage({
                   <Stack direction="row" alignItems="center" columnGap={3}>
                     <CustomLink
                       href={{ pathname: params.user_type === UserType.INTERNAL_USER ? `/` : `/client`, query: { token } }}
+                      revalidate={true}
                     >
```
Comment (aatbip) [inline src/app/detail/[task_id]/[user_type]/page.tsx:88]: This is still not solving the underlying issue of caching behaviour during browser navigation. I think browser navigation part can't be ignored because users are actually navigate back the iframe through the main app. 

I think this change is not required because #280 already removes the caching by upgrading the version.

aatbip (changes requested):

src/hoc/ClientSideStateUpdate.tsx
```diff
@@ -66,11 +66,12 @@ export const ClientSideStateUpdate = ({
...
+      const view = viewSettingsTemp ? viewSettingsTemp.filterOptions : viewSettings.filterOptions
+      view?.type == FilterOptionsKeywords.CLIENTS
         ? store.dispatch(setFilteredAssgineeList({ filteredType: FilterByOptions.CLIENT }))
-        : viewSettings.filterOptions?.type == FilterOptionsKeywords.TEAM
+        : view?.type == FilterOptionsKeywords.TEAM
           ? store.dispatch(setFilteredAssgineeList({ filteredType: FilterByOptions.IUS }))
-          : viewSettings.filterOptions?.type == ''
+          : view.type == ''
             ? store.dispatch(setFilteredAssgineeList({ filteredType: FilterByOptions.NOFILTER }))
             : store.dispatch(setFilteredAssgineeList({ filteredType: FilterByOptions.NOFILTER }))
```
rrojan, line 76: Why is this tertiary operator going down 4 whole levels??

You can fetch the dispatch function using an object lookup for view.type

rrojan, line 76:
```
? store.dispatch(setFilteredAssgineeList({ filteredType: FilterByOptions.NOFILTER }))
            : store.dispatch(setFilteredAssgineeList({ filteredType: FilterByOptions.NOFILTER }))
```

Aren't the expressions on the first and second part of the ternary operator doing the exact same thing? Why do we even need a check for `view.type == ''` in the first place then?

src/hoc/ClientSideStateUpdate.tsx
```diff
@@ -66,11 +66,12 @@ export const ClientSideStateUpdate = ({
...
       store.dispatch(setViewSettings(viewSettings))
-      viewSettings.filterOptions?.type == FilterOptionsKeywords.CLIENTS
+      const view = viewSettingsTemp ? viewSettingsTemp.filterOptions : viewSettings.filterOptions
+      view?.type == FilterOptionsKeywords.CLIENTS
         ? store.dispatch(setFilteredAssgineeList({ filteredType: FilterByOptions.CLIENT }))
-        : viewSettings.filterOptions?.type == FilterOptionsKeywords.TEAM
+        : view?.type == FilterOptionsKeywords.TEAM
           ? store.dispatch(setFilteredAssgineeList({ filteredType: FilterByOptions.IUS }))
-          : viewSettings.filterOptions?.type == ''
+          : view.type == ''
```
rrojan, line 74: `view` can be falsey on 70 and 72 but not on line 74(?)

Looking at your implementation it is highly unlikely `view` will ever be falsey at all considering `initialState`.

src/components/layouts/FilterBar.tsx
```diff
@@ -64,24 +75,24 @@ export const FilterBar = ({
...
-    filterOptions.type == FilterOptionsKeywords.CLIENTS
+    viewModeFilterOptions.type == FilterOptionsKeywords.CLIENTS
       ? 2
-      : filterOptions.type == FilterOptionsKeywords.TEAM
+      : viewModeFilterOptions.type == FilterOptionsKeywords.TEAM
         ? 1
-        : filterOptions.type == ''
+        : viewModeFilterOptions.type == ''
           ? 3
           : 0
```
rrojan, line 84: Use object lookup instead.

rrojan (changes requested):

src/types/objectMaps.ts
```diff
@@ -0,0 +1,20 @@
+import { FilterByOptions, FilterOptionsKeywords, IAssigneeCombined } from '@/types/interfaces'
+
+export const filterTypeToButtonIndexMap: Record<string, number> = {
+  [FilterOptionsKeywords.CLIENTS]: 2,
+  [FilterOptionsKeywords.TEAM]: 1,
+  '': 3,
```
aatbip, line 6: This can be interpreted as `default` just like below? It is creating a bit of confusion because same entity is defined as `default` and `null`.

aatbip (changes requested):

aatbip (approved):

rrojan (approved):
LGTM

PR #277, opened 2024-08-23
OUT-770 | OUT-762 | OUT-739 | OUT-738 | Error: Unable to authorize Copilot SDK.

rrojan:
### Tasks

- [OUT-739 | Error: Unable to authorize Copilot SDK.](https://linear.app/copilotplatforms/issue/OUT-739/)
- [OUT-762 | Error: Unable to authorize Copilot SDK.](https://linear.app/copilotplatforms/issue/OUT-762/)
- [OUT-770 | Error: Unable to authorize Copilot SDK.](https://linear.app/copilotplatforms/issue/OUT-770/)

### Changes

- [x] Add `SilentError` component to suppress error `throw` when token is empty
- [x] Make error  digest optional

### Testing Criteria

- [x] Checked logs but no errors were triggered


src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -67,7 +68,7 @@ export default async function TaskDetailPage({
   const [task, tokenPayload] = await Promise.all([getOneTask(token, task_id), copilotClient.getTokenPayload()])
   // Basic validation
   if (!tokenPayload) {
-    throw new Error('Token cannot be found')
+    return <SilentError message="Please provide a Valid Token" />
```
Comment (aatbip) [inline src/app/detail/[task_id]/[user_type]/page.tsx:71]: Please help me understand what is the purpose of building `SilentError`? Why can't the general `ClientErrorBoundary` would just work fine?

aatbip (changes requested):

src/app/detail/[task_id]/[user_type]/page.tsx
Comment (rrojan) [reply src/app/detail/[task_id]/[user_type]/page.tsx]: @aatbip for reusability. This same logic is used in all 3 of our pages. I don't want to have to pass `new Error("Please provide a valid token" and reset on all 3 pages.

Comment (rrojan) [reply src/app/detail/[task_id]/[user_type]/page.tsx]: Is the question why I made a separate component for this, or why i'm not throwing an error to render ClientErrorBoundary?

Comment (aatbip) [reply src/app/detail/[task_id]/[user_type]/page.tsx]: First comment answered my question. Thank you for the clarification.

aatbip (approved):
lgtm

makeitraina:
Im jsut curious, when is token empty? I thought it should never be empty right?

makeitraina:
> > Im jsut curious, when is token empty? I thought it should never be empty right?
> 
> @makeitraina After looking into this recurring issue I found that in requests where token was empty, most of them had user-agents set to http client libraries (mostly go-http-client).
> 
> This led me to believe the requests were either being sent by Vercel itself for site previews / healthchecks, or by other bots / crawlers. Those requests will not have a token provided in the URL.
> 
> Screenshot from 2024-08-23 17-12-57
> 
> Http clients implemented using [net/http](https://pkg.go.dev/net/http) have `go-http-client` by default unless they're modified so they are most likely bots/crawlers, vercel's or others.
> 
> image
> 
> Have you experienced something like this in other Copilot apps, Neil?

Ahh interesting, this makes sense. Copilot team also has a health check that we do for apps but we are not using a token there. So we should definitely pass in a token for it!

When you do health check with uptime robot, are you using token there?

PR #281, opened 2024-08-28
OUT-790 | OUT-774 | Improve logging for all parts of Tasks App | Unable to authorize Copilot SDK

rrojan:
### Tasks

- [OUT-790 | Improve logging for all parts of Tasks App](https://linear.app/copilotplatforms/issue/OUT-790/improve-logging-for-all-parts-of-tasks-app)
- [OUT-774 | Unable to authorize Copilot SDK](https://linear.app/copilotplatforms/issue/OUT-774)

### Changes

- [x] Add verbose logging for all pages, APIs and CopilotAPI for better debugging. Currently we are unable to debug so many tasks because we aren't aware of the user that came across it.

### DIFF
--- a/src/app/detail/[task_id]/[user_type]/page.tsx
+++ b/src/app/detail/[task_id]/[user_type]/page.tsx
@@ -63,7 +63,7 @@ export default async function TaskDetailPage({
   searchParams: { token: string; isRedirect?: string }
 }) {
   const { token } = searchParams
-  const { task_id } = params
+  const { task_id, user_type } = params
 
   if (z.string().safeParse(token).error) {
     return <SilentError message="Please provide a Valid Token" />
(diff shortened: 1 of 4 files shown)


src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -76,6 +76,8 @@ export default async function TaskDetailPage({
     throw new Error('Please provide a Valid Token')
   }
 
+  console.info(`app/detail/${task_id}/${user_type}/page.tsx | Serving user ${token} with payload`, tokenPayload)
+
```
Comment (aatbip) [inline src/app/detail/[task_id]/[user_type]/page.tsx:80]: More than doing this in multiple places in the server components (page.tsx), why don't we create another server component that handles logging for all the server components and use that in the layout.tsx? 

For the CopilotAPI, it looks good.

aatbip (changes requested):

src/app/detail/[task_id]/[user_type]/page.tsx
Comment (rrojan) [reply src/app/detail/[task_id]/[user_type]/page.tsx:80]: Great point... but as you can see we're logging the token payload as well so its easier to debug auth issues later on. And token payload is currently being fetched in the page level only. That's why I'm replicating those `console.info`s on all pages.

Same for not using `middleware.ts`

aatbip (approved):
Awesome work of engineering. 👍

PR #283, opened 2024-08-29
OUT-537 | Implement 'New' action for each status

rrojan:
### Tasks

- [OUT-537 | Implement 'New' action for each status](https://linear.app/copilotplatforms/issue/OUT-537/implement-new-action-for-each-status)

### Changes

- [x] Add Plus icon on top of each task column
- [x] Implement new task modal popup with workflow state pre-selected when clicking on this icon
- [x] Refactor some existing code 

### Testing Criteria

- [x] [LOOM](https://www.loom.com/share/78cddf4d4d7c4cf88c393ac72c77d762?sid=644e7f8f-6b87-44d6-87ce-e4e0d4ba68c7)


aatbip (changes requested):
1. Right padding is not proper for the plus icon. 
    
Screenshot from 2024-09-03 09-34-48

2. The plus icon is not aligned properly with the filter bar layout in the medium and small screen sizes. 

Screenshot from 2024-09-03 09-42-58
Screenshot from 2024-09-03 09-42-53

src/app/ui/Modal_NewTaskForm.tsx
```diff
@@ -24,17 +30,16 @@ export const ModalNewTaskForm = ({
   const { title, description, workflowStateId, assigneeId, assigneeType, attachments, dueDate, showModal } =
     useSelector(selectCreateTask)
 
+  const handleModalClose = useCallback(async () => {
+    store.dispatch(setShowModal())
```
aatbip, line 34: `useCallback` is not particularly necessary here because - 

No Significant Re-Renders: The handleModalClose function is not being passed as a prop to deeply nested components or as a dependency to other hooks that would benefit from memoization to avoid unnecessary re-renders.

Minimal Performance Impact: The function is only used in the Modal and NewTaskForm components. Recreating this function on re-renders when filterOptions changes is unlikely to have a significant performance impact.

src/app/ui/TaskBoard.helpers.tsx
```diff
@@ -0,0 +1,7 @@
+import { setActiveWorkflowStateId, setShowModal } from '@/redux/features/createTaskSlice'
+import store from '@/redux/store'
+
+export const handleAddBtnClicked = (workflowStateId: string) => {
+  store.dispatch(setActiveWorkflowStateId(workflowStateId))
+  store.dispatch(setShowModal())
+}
```
aatbip, line 7: I think it is not a good idea to place the helpers for UI state inside the `ui` directory. This might also be useful for the `New Task` button on the top layout. 
We are also not using this directly from the `TaskBoard` component but from the `TaskColumn` and `TaskRow`. 
It is a good idea to abstract away such repetitive UI flows but I think we should make a clear structural decision on where to place such things as there are many more we can abstract away.

aatbip (changes requested):

src/app/ui/TaskBoard.helpers.tsx
rrojan: I've already put this in a separate file in OUT-794 which was created from this branch. Is this better?
image

aatbip: I mean the same thing where the file reside. Can we revamp?

rrojan: As in.. in the task row and column component itself?
Since this functionality was shared I wanted to not duplicate it across files @aatbip

aatbip: I appreciate the effort to abstract the UI flow, but I think we should reconsider where we're placing this code. The `ui` directory should ideally focus on UI components and their styling. Since the `handleAddBtnClicked` function is more about managing state and could be reused across different components (like the New Task button in the top layout), it would make more sense to place it in a more centralized location, perhaps under a redux or a different directory.

By keeping these helpers outside of the ui directory, we not only maintain a clear separation of concerns but also make the code more reusable and organized. This way, if we need similar functionality elsewhere, we're not tied to a specific UI component structure.
@rrojan

aatbip (approved):
The folder structure might have to be considered - [here](https://github.com/copilot-platforms/tasks-app/pull/283#discussion_r1741946932). LGTM other than. Great job with this one!

PR #284, opened 2024-08-30
OUT-791 Task creation doesn't work as expected when IU is on the List view

arpandhakal:
### Tasks

- [Task creation doesn't work as expected when IU is on the List view](T[ask Link](https://linear.app/copilotplatforms/issue/OUT-791/task-creation-doesnt-work-as-expected-when-iu-is-on-the-list-view))

### Changes

- [x] the useHandler component was responsible for overriding the createTask states in createTask slice from list view assignee selector, added a prop to disable update from list view assignee selector.
- [x]  the above functionality basically fixed 3 scenarios mentioned in the issue.

### Testing Criteria

- [LOOM](https://www.loom.com/share/341bfefb419d42a5bbbe06ecadff0857)

### DIFF
--- a/src/app/ui/NewTaskForm.tsx
+++ b/src/app/ui/NewTaskForm.tsx
@@ -63,12 +63,14 @@ export const NewTaskForm = ({
     item: workflowStates[0],
     type: SelectorType.STATUS_SELECTOR,
   })
+
   const { renderingItem: _assigneeValue, updateRenderingItem: updateAssigneeValue } = useHandleSelectorComponent({
     item:
       filteredAssignees.find(
         (item) => item.id == filterOptions[FilterOptions.ASSIGNEE] || item.id == filterOptions[FilterOptions.TYPE],
       ) ?? null,
     type: SelectorType.ASSIGNEE_SELECTOR,
+    createTaskfieldUpdate: true,
   })
   const { renderingItem: _templateValue, updateRenderingItem: updateTemplateValue } = useHandleSelectorComponent({
     item: undefined, //initially we don't want any value to be selected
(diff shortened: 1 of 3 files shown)


src/components/layouts/FilterBar.tsx
```diff
@@ -76,17 +76,19 @@ export const FilterBar = ({
...
     type: SelectorType.ASSIGNEE_SELECTOR,
+    createTaskfieldUpdate: true,
   })
   useFilter(viewSettingsTemp ? viewSettingsTemp.filterOptions : filterOptions)
   const filterButtons = [
     {
       name: 'My tasks',
       onClick: () => {
-        handleFilterOptionsChange(FilterOptions.TYPE, IUTokenSchema.parse(tokenPayload)?.internalUserId)
+        const selfAssigneeId = IUTokenSchema.parse(tokenPayload)?.internalUserId
```
rrojan, line 90: Optional chaining (?.) shouldn't be necessary after a strict parse

src/app/ui/NewTaskForm.tsx
```diff
@@ -63,12 +63,14 @@ export const NewTaskForm = ({
...
     type: SelectorType.STATUS_SELECTOR,
   })
+
   const { renderingItem: _assigneeValue, updateRenderingItem: updateAssigneeValue } = useHandleSelectorComponent({
     item:
       filteredAssignees.find(
         (item) => item.id == filterOptions[FilterOptions.ASSIGNEE] || item.id == filterOptions[FilterOptions.TYPE],
       ) ?? null,
     type: SelectorType.ASSIGNEE_SELECTOR,
+    createTaskfieldUpdate: true,
```
rrojan, line 73: This is a quick and easy fix but perhaps we could make an `opts` param or something similar to pass `mode: 'createTaskFieldUpdate'`.

The reason for this is we are going to have to deal with both multi and single assignee selector in the near future, and this `opts` param could be used to pass additional params to `userHandleSelectorComponent` while still keeping the code clean.

rrojan (changes requested):

rrojan (approved):
LGTM :trophy:

PR #287, opened 2024-09-03
Improvements: 001 | Split prisma schema files + Optimize relation load strategy with db-level JOIN

rrojan:
### Tasks

- N/A

### Changes

- [x] Enable "prismaSchemaFolder" to split `schema.prisma` into dedicated model files for better organization and usability
- [x] Enable "relationJoins" to use DB-level joins instead of application level merges


sazanrjb (approved):
This is a great upgrade! 🔥

PR #288, opened 2024-09-04
OUT-800 "+ New Task" button shouldn't appear for clients if task board is empty

aatbip:
### Changes

- [x] Adds strict check for "New Task" button

### Testing Criteria

- [LOOM](https://www.loom.com/share/d34059d35b8d4222b06e713c1bf813a7?sid=4e180309-3fa7-497b-bb12-fc0a2c44261f)

### DIFF
--- a/src/components/layouts/EmptyState/DashboardEmptyState.tsx
+++ b/src/components/layouts/EmptyState/DashboardEmptyState.tsx
@@ -8,8 +8,11 @@ import store from '@/redux/store'
 import { UserType } from '@/types/interfaces'
 import { SxCenter } from '@/utils/mui'
 import { Box, Stack, Typography } from '@mui/material'
+import { usePathname } from 'next/navigation'
 
 const DashboardEmptyState = ({ userType }: { userType: UserType }) => {
+  const pathname = usePathname()
+  console.log(pathname.includes('client'))
   return (
     <>
       <AppMargin size={SizeofAppMargin.LARGE} py="20px">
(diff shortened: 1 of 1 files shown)


src/components/layouts/EmptyState/DashboardEmptyState.tsx
```diff
@@ -44,7 +46,7 @@ const DashboardEmptyState = ({ userType }: { userType: UserType }) => {
                   : 'Tasks will show here once they’ve been assigned to you. '}
               </Typography>
             </Stack>
-            {userType == UserType.INTERNAL_USER && (
+            {userType === UserType.INTERNAL_USER && !!!pathname.includes('client') && (
```
rrojan, line 49: Shouldn't a `!pathname.includes('client')` be enough here?

rrojan (changes requested):

src/components/layouts/EmptyState/DashboardEmptyState.tsx
aatbip: It was just a personal preference though single `!` operator is also enough :D

rrojan: I think the main question here is why `userType === UserType.INTERNAL_USER` isn't working @aatbip 

This fix just seems like patchwork over another deeper problem

rrojan (changes requested):
I think the main question here is why `userType === UserType.INTERNAL_USER` isn't working @aatbip . There might be other consequences to this if so.

This fix just seems like patchwork over another deeper problem

aatbip:
> I think the main question here is why `userType === UserType.INTERNAL_USER` isn't working @aatbip . There might be other consequences to this if so.
> 
> This fix just seems like patchwork over another deeper problem

This issue was not reproducible anywhere when I tried yesterday. The check `userType == UserType.INTERNAL_USER` should have always worked.
The context of this issue is not properly mentioned in the ticket such as ways to reproduce, how many times it occurred, or a loom video demonstrating the occurrence, and seems like it was happening while running locally according to the SS attached. 

I think we should merge this PR to main because it still improves the check and maintains stricter checking. However, further digging might not be necessary as we are not able to reproduce this issue and more context is not available atm. 

cc: @rrojan @sazanrjb

PR #296, opened 2024-09-12
OUT-819 Support for deleting images from bucket

arpandhakal:
### Tasks

- [Support for deleting images from bucket](https://linear.app/copilotplatforms/issue/OUT-819/support-for-deleting-images-from-bucket)

### Changes

- [x] removed unused image by applying cron job. 

### Notes

- some checks are commented out to test this in preview


src/app/api/workers/scrap-images/scrap-images.controller.ts
```diff
@@ -0,0 +1,15 @@
+import { NextRequest, NextResponse } from 'next/server'
+import { ScrapImageService } from '@/app/api/workers/scrap-images/scrap-images.service'
+import { cronSecret } from '@/config'
+
+export const RemoveScrapImages = async (request: NextRequest) => {
```
rrojan, line 5: Functions should be camelCased. Also please wrap this with withErrorHandler for proper error logging

src/app/api/workers/scrap-images/scrap-images.controller.ts
```diff
@@ -0,0 +1,15 @@
...
+  const authHeader = request.headers.get('authorization')
+  if (authHeader !== `Bearer ${cronSecret}`) {
+    return new Response('Unauthorized', {
+      status: 401,
+    })
```
rrojan, line 10: Where did this Response come from? This is not imported

src/app/api/workers/scrap-images/scrap-images.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    for (const image of scrapImages) {
+      const task = await db.task.findUnique({
+        where: {
+          id: image.taskId,
+        },
+      })
+      await db.scrapImages.delete({
+        where: { id: image.id },
+      })
+      if (task && (task.body as string).includes(image.filePath)) {
```
rrojan, line 34: In our schema task.body can be undefined as well. Using `includes` on undefined will crash the endpoint.

image

rrojan (changes requested):

prisma/schema.prisma
```diff
@@ -212,3 +213,17 @@ model Label {
 
   @@map("Labels")
 }
+
+model ScrapImages {
+  id          String       @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  filePath    String       @unique
```
rrojan, line 219: We are frequently querying ScrapImages table using `where filePath = ?`, and filePath is unique. I suggest we add an index to this column for faster queries

prisma/schema.prisma
```diff
@@ -212,3 +213,17 @@ model Label {
 
   @@map("Labels")
 }
+
+model ScrapImages {
+  id          String       @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
```
rrojan, line 218: If you pull the latest feature/M2 I believe the schema files should be split into separate model files

rrojan (changes requested):

rrojan (changes requested):

src/app/api/workers/scrap-images/scrap-images.service.ts
```diff
@@ -0,0 +1,41 @@
...
+        continue
+      }
+
+      await supabase.supabase.storage.from(supabaseBucket).remove([image.filePath])
```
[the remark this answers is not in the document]
rrojan: Fixed

src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -152,6 +154,11 @@ export default async function TaskDetailPage({
                       const data = await getSignedUrlUpload(token, fileName)
                       return data
                     }}
+                    postScrapImage={async (payload: ScrapImageRequest) => {
+                      'use server'
+                      const data = await postScrapImage(token, payload)
+                      return data
+                    }}
```
Comment (priosshrsth) [inline src/app/detail/[task_id]/[user_type]/page.tsx:161]: Let's extract this in a variable. And also 159, 160 could be simplified in a single return statement.

src/app/detail/ui/TaskEditor.tsx
```diff
@@ -171,6 +175,16 @@ export const TaskEditor = ({
...
+          deleteEditorAttachments={async (url: string) => {
+            const filePath = await getFilePathFromUrl(url)
+            if (filePath) {
+              const payload: ScrapImageRequest = {
+                filePath: filePath,
+                taskId: task_id,
+              }
+              postScrapImage(payload)
+            }
+          }}
```
priosshrsth, line 187: Again, it's better to extract these are into it's own variable and pass that variable as a function rather than declaring them inline.

priosshrsth (changes requested):

priosshrsth:
@arpandhakal I am not sure if TapWrite is memoized but, I think its an expensive component to re render because of all the thingg it do (I'm guessing is it the editor, right?). What I'm seeing is we are passing inline function as a props. This might be very bad approach if my uderstanding is correct because we'll be creating new anonymous function everytime the parent is re-rendered. And essentially our child component will be re rendered every time the parent is re rendered. So I think we should start extracting those inline function in it's own variable. At least the complex ones. And maybe we should even start using **useCallBack**.

rrojan (approved):
LGTM :trophy: 
As we're leaving vercel.json as-is, please be sure to remove cron-job.org's cron task after this is merged to prod

PR #301, opened 2024-09-19
OUT-817 Fix DnD - browser globe icon follows cursor when moving a task to a different status

aatbip:
### Changes

- [x] Removes the globe icon appearing in Chrome/Arc browsers on MacOS during the drag and drop operations

### DIFF
--- a/src/hoc/DragDropHandler.tsx
+++ b/src/hoc/DragDropHandler.tsx
@@ -67,7 +67,9 @@ export const DragDropHandler = ({
     if (window) {
       //we only need custom drag preview in the list view
       if (draggable) {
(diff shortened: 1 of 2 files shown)


src/hoc/DragDropHandler.tsx
```diff
@@ -53,9 +53,7 @@ export const DragDropHandler = ({
 
   const [{ isDragging }, drag, preview] = useDrag({
     type: accept,
-    item: () => {
-      return { task: task }
-    },
+    item: { task: task },
```
priosshrsth, line 56: This can be simplified. But I think we should provide dependency array here here with task as input.

priosshrsth (approved):

PR #307, opened 2024-09-23
OUT 834 Tabs and the filter assignee drop-down are shown for Client in mobile view

aatbip:
### Changes

- [x] Hides adv filters in the client side mobile view


arpandhakal (commented):
The tasks tells us not to show tabs and the filter assignee dropdown in mobile view. But it seems like we removed it in desktop view too.

rrojan (approved):
LGTM :trophy:

arpandhakal (approved):

PR #310, opened 2024-09-24
OUT-840 add back top padding on task details

arpandhakal:
### Tasks

- [add back top padding on task details](https://linear.app/copilotplatforms/issue/OUT-840/add-back-top-padding-on-task-details)

### Changes

- [x] applied vertical and horizontal padding and adjusted max width to be 672px with padding 36px on each sides which makes the max width to be 600px

### Testing Criteria

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "react-dnd-touch-backend": "^16.0.1",
     "react-dom": "^18",
     "react-redux": "^9.1.0",
-    "tapwrite": "^1.1.9",
+    "tapwrite": "^1.1.10",
     "zod": "^3.23.8"
   },
   "devDependencies": {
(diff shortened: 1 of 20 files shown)


aatbip (changes requested):
Left and right padding doesn't match - 

You can see below that left padding ~36px and right is ~26px which is a huge difference. 

Screenshot from 2024-09-24 13-17-07

Screenshot from 2024-09-24 13-20-18




Whereas in Figma it matches with a proportionate fraction - 

Screenshot from 2024-09-24 13-20-26

aatbip (changes requested):
The task title and description section doesn't align proportionately to the top layout. Is it supposed to be like that? 
Also, the image has a slightly off left margin. Please consider resolving. 

Screenshot from 2024-09-25 11-02-49

PR #312, opened 2024-09-25
OUT-837 | OUT-836 | Add retry & delay mechanism to CopilotAPI

rrojan:
### Tasks

- [OUT-836 | TypeError: fetch failed](https://linear.app/copilotplatforms/issue/OUT-836/typeerror-fetch-failed)
- [OUT-837 | TypeError: fetch failed](https://linear.app/copilotplatforms/issue/OUT-837)
- And other tasks in the past regarding rate-limits / 500 response from CopilotAPI

### Changes

- [x] Implement `withRetry` HOF utils for flexible promise retries & delay
- [x] Implement `withRetry` on each CopilotAPI method
- [x] Clean up withErrorHandler to better detect and respond to Copilot API errors 

### Testing Criteria

- [x] General testing of Tasks App where data is fetched from / added to Copilot API
- [x] Stress testing retry mechanism: [LOOM](https://www.loom.com/share/f27f600e17d64e9683d1fa4f082f2916?sid=a3e46150-1a8b-400b-83fb-0f13db577a5e)

### DIFF
--- a/src/app/api/core/utils/withErrorHandler.ts
+++ b/src/app/api/core/utils/withErrorHandler.ts
@@ -1,18 +1,9 @@
 import { NextRequest, NextResponse } from 'next/server'
 import { ZodError, ZodFormattedError, ZodIssue } from 'zod'
-import { CopilotApiError } from '@/types/CopilotApiError'
+import { CopilotApiError, MessagableError, StatusableError } from '@/types/CopilotApiError'
 import APIError from '@api/core/exceptions/api'
 import { PrismaClientKnownRequestError } from '@prisma/client/runtime/library'
 import httpStatus from 'http-status'
-
-interface StatusableError {
-  status: number
-}
-interface MessageableError {
-  body?: {
-    message: string
-  }
-}
 
 type RequestHandler = (req: NextRequest, params: any) => Promise<NextResponse>
 
@@ -49,7 +40,7 @@ export const withErrorHandler = (handler: RequestHandler): RequestHandler => {
 
       // Default staus and message for JSON error response
       let status: number = (error as StatusableError).status || httpStatus.BAD_REQUEST
-      let message: string | ZodFormattedError<string> = (error as MessageableError).body?.message || 'Something went wrong'
+      let message: string | ZodFormattedError<string> = (error as MessagableError).body?.message || 'Something went wrong'
       let errors: unknown[] | undefined = undefined
 
       // Build a proper response based on the type of Error encountered
--- a/src/app/api/core/utils/withErrorHandler.ts
+++ b/src/app/api/core/utils/withErrorHandler.ts
@@ -1,9 +1,9 @@
-import { NextRequest, NextResponse } from 'next/server'
-import { ZodError, ZodFormattedError, ZodIssue } from 'zod'
 import { CopilotApiError, MessagableError, StatusableError } from '@/types/CopilotApiError'
 import APIError from '@api/core/exceptions/api'
 import { PrismaClientKnownRequestError } from '@prisma/client/runtime/library'
 import httpStatus from 'http-status'
+import { NextRequest, NextResponse } from 'next/server'
+import { ZodError, ZodFormattedError } from 'zod'
 
 type RequestHandler = (req: NextRequest, params: any) => Promise<NextResponse>
 
--- a/src/app/api/core/utils/withErrorHandler.ts
+++ b/src/app/api/core/utils/withErrorHandler.ts
@@ -5,6 +5,15 @@ import APIError from '@api/core/exceptions/api'
 import { PrismaClientKnownRequestError } from '@prisma/client/runtime/library'
 import httpStatus from 'http-status'
 
+interface StatusableError {
+  status: number
+}
+interface MessageableError {
+  body?: {
+    message: string
+  }
+}
+
 type RequestHandler = (req: NextRequest, params: any) => Promise<NextResponse>
 
 /**
@@ -39,8 +48,8 @@ export const withErrorHandler = (handler: RequestHandler): RequestHandler => {
       console.error(formattedError)
 
       // Default staus and message for JSON error response
-      let status: number = httpStatus.BAD_REQUEST
-      let message: string | ZodFormattedError<string> = 'Something went wrong'
+      let status: number = (error as StatusableError).status || httpStatus.BAD_REQUEST
+      let message: string | ZodFormattedError<string> = (error as MessageableError).body?.message || 'Something went wrong'
       let errors: unknown[] | undefined = undefined
 
       // Build a proper response based on the type of Error encountered
--- a/src/app/api/core/utils/withRetry.ts
+++ b/src/app/api/core/utils/withRetry.ts
@@ -1,7 +1,7 @@
 import pRetry, { FailedAttemptError } from 'p-retry'
 
-export const withRetry = <T>(fn: (...args: any[]) => Promise<T>, args: any[]): Promise<T> => {
-  return pRetry(() => fn(...args), {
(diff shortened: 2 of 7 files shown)


src/app/api/core/utils/withRetry.ts
```diff
@@ -0,0 +1,49 @@
+import { StatusableError } from '@/types/CopilotApiError'
+import pRetry, { FailedAttemptError } from 'p-retry'
+import * as Sentry from '@sentry/nextjs'
+
+export const withRetry = async <T>(fn: (...args: any[]) => Promise<T>, args: any[]): Promise<T> => {
+  let isEventProcessorRegisterd = false
```
aatbip, line 6: I think that setting `isEventProcessorRegisterd` to true after adding the event processor ensures that we don’t re-register the event processor. However, there's a potential edge case where, if the scope is not reused or gets cleared between retries, the processor may not be registered again.

src/app/api/core/utils/withRetry.ts
```diff
@@ -0,0 +1,49 @@
...
+        return await fn(...args)
+      } catch (error) {
+        // Hopefully now sentry doesn't report retry errors as well. We have enough triage issues as it is
+        Sentry.withScope((scope) => {
+          if (isEventProcessorRegisterd) return
+
+          isEventProcessorRegisterd = true
+          scope.addEventProcessor((event) => {
+            if (event.level === 'error' && event.message && event.message.includes('An error occurred during retry')) {
+              return null // Discard the event as it occured during retry
```
aatbip, line 20: Here we are relying on `event.message` including "An error occurred during retry" string, which can be fragile if error messages change in the future. Can we consider adding a more explicit marker to the error or checking for more reliable properties, like a specific error code or type, to prevent accidentally filtering out unrelated errors?

src/app/api/core/utils/withRetry.ts
```diff
@@ -0,0 +1,49 @@
...
+      factor: 2, // Exponential factor for timeout delay. Tweak this if issues still persist
+      onFailedAttempt: (error: FailedAttemptError) => {
+        console.warn(
+          `CopilotAPI#withRetry | Attempt ${error.attemptNumber} failed. There are ${error.retriesLeft} retries left. Error:`,
+          error,
+        )
+      },
+      shouldRetry: (error: any) => {
+        // Typecasting because Copilot doesn't export an error class
+        const err = error as StatusableError
```
aatbip, line 43: We are casting the error to `StatusableError` to check the status. This makes sense given the context, but doesn't ensure that the cast is safe and handled, especially in production where an unexpected error type could occur.

src/app/api/core/utils/withRetry.ts
```diff
@@ -0,0 +1,49 @@
+import { StatusableError } from '@/types/CopilotApiError'
+import pRetry, { FailedAttemptError } from 'p-retry'
+import * as Sentry from '@sentry/nextjs'
+
+export const withRetry = async <T>(fn: (...args: any[]) => Promise<T>, args: any[]): Promise<T> => {
```
aatbip, line 5: The `fn` parameter can take any number of arguments due to the `(...args: any[])` definition. This is fine but I think it does reduce type safety.
If possible, let's define the function signature more explicitly based on your use case for better type safety.

src/app/api/core/utils/withRetry.ts
```diff
@@ -0,0 +1,49 @@
...
+  return await pRetry(
+    async () => {
+      try {
+        return await fn(...args)
+      } catch (error) {
+        // Hopefully now sentry doesn't report retry errors as well. We have enough triage issues as it is
+        Sentry.withScope((scope) => {
+          if (isEventProcessorRegisterd) return
+
+          isEventProcessorRegisterd = true
```
aatbip, line 17: Is this a typo? `isEventProcessorRegisterd` should be `isEventProcessorRegistered`.

src/app/api/core/utils/withRetry.ts
```diff
@@ -0,0 +1,49 @@
...
+        // Rethrow the error so pRetry can rety
+        throw error
+      }
+    },
+
+    {
+      retries: 3,
+      minTimeout: 500,
+      maxTimeout: 500,
+      factor: 2, // Exponential factor for timeout delay. Tweak this if issues still persist
```
aatbip, line 34: The backoff logic with `factor: 2` is a good approach for retrying. Could you please help me understand the reason behind setting `maxTimeout` and `minTimeout` to 500?

Could setting both `minTimeout` and `maxTimeout` to the same value negate the exponential increase? I think we should consider allowing a larger `maxTimeout` (e.g., 2000) to provide actual exponential backoff benefits. What do you think?

src/utils/CopilotAPI.ts
```diff
@@ -159,13 +165,54 @@ export class CopilotAPI {
...
     await this.copilot.deleteNotification({ id })
   }
+
+  // Methods wrapped with retry
+  getTokenPayload = async (...args: unknown[]) => await withRetry(this._getTokenPayload.bind(this), args)
+
+  me = async (...args: unknown[]) => await withRetry(this._me.bind(this), args)
+
+  getWorkspace = async (...args: unknown[]) => await withRetry(this._getWorkspace.bind(this), args)
+
```
aatbip, line 184: Instead of writing `withRetry` for each public method, consider using a dynamic approach or decorator to reduce code repetition.

```javascript
wrapWithRetry(method: Function) {
  return async (...args: unknown[]) => await withRetry(method.bind(this), args)
}
constructor(...) {
  this.getTokenPayload = this.wrapWithRetry(this._getTokenPayload)
}
```

src/app/api/core/utils/withErrorHandler.ts
```diff
@@ -39,8 +39,8 @@ export const withErrorHandler = (handler: RequestHandler): RequestHandler => {
       console.error(formattedError)
 
       // Default staus and message for JSON error response
-      let status: number = httpStatus.BAD_REQUEST
-      let message: string | ZodFormattedError<string> = 'Something went wrong'
+      let status: number = (error as StatusableError).status || httpStatus.BAD_REQUEST
+      let message: string | ZodFormattedError<string> = (error as MessagableError).body?.message || 'Something went wrong'
```
aatbip, line 43: The code uses type assertions like `error as StatusableError` and `error as MessagableError` repeatedly, which is risky if error doesn't match the expected structure. Instead of type casting, using type guards (custom functions to check if error is of a particular type) would be safer.

src/app/api/core/utils/withRetry.ts
rrojan: This scope should only exist after withRetry is triggered and the child exception is caught. So it not being registered again will not be an issue.

src/app/api/core/utils/withRetry.ts
rrojan: Using a generic makes it unusable for our needs. Do you have any suggestion here?

src/app/api/core/utils/withErrorHandler.ts
rrojan: Type guards aren't necessary here since we have a `||` clause following closesly by. If error doesn't satisfy `StatusableError` or `MessagableError`, the default value will simply substitute this.

src/app/api/core/utils/withRetry.ts
rrojan: This shouldn't be an issue because only CopilotApiError will be caught here. I'll rework the typecasting to make this more verbose

rrojan: We are returning error.status here. If  error does not satisfy StatusableError, `error.status === 429` will simply return a falsey value. Type assumption is completely safe in use cases like this.

aatbip (approved):

PR #314, opened 2024-09-27
OUT-851 Image is broken when IU changes the task status from properties sidebar

arpandhakal:
### Tasks

- [Image is broken when IU changes the task status from properties sidebar](https://linear.app/copilotplatforms/issue/OUT-851/image-is-broken-when-iu-changes-the-task-status-from-properties)

### Changes

- [x] updated tasks on getOneTask with signedUrl which revalidates in 60 minutes. This causes the realtime updates to set fresh data with fresh signed url on the store.

### Testing Criteria

- [LOOM](https://www.loom.com/share/ea6b78278936421eb47aa1fd9731dc92)


aatbip (changes requested):
Please remove `console.log`s

PR #319, opened 2024-09-30
OUT-857 Improve image resizing responsiveness

arpandhakal:
### Tasks

- [Improve image resizing responsiveness](https://linear.app/copilotplatforms/issue/OUT-857/improve-image-resizing-responsiveness)

### Changes

- [x] fixed newly uploaded image with height auto width 100% resizing issue.
- [x] made image's aspect ratio with respect to container to fully attain the width of the container. 
- [x] changing screen size from dev tools causing issues. some fixes are still remaning here. 
- [x] responsiveness on switching from larger screen to smaller, the vice versa is remaining.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "react-dnd-touch-backend": "^16.0.1",
     "react-dom": "^18",
     "react-redux": "^9.1.0",
-    "tapwrite": "^1.1.21",
+    "tapwrite": "^1.1.23",
     "zod": "^3.23.8"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -7261,10 +7261,10 @@ tapable@^2.2.0:
   resolved "https://registry.yarnpkg.com/tapable/-/tapable-2.2.1.tgz#1967a73ef4060a82f12ab96af86d52fdb76eeca0"
   integrity sha512-GNzQvQTOIP6RyTfE2Qxb8ZVlNmw0n88vp1szwWRimP02mnTsx3Wtn5qRdqY9w2XduFNUgvOwhNnQsjwCp+kqaQ==
 
-tapwrite@^1.1.21:
-  version "1.1.21"
-  resolved "https://registry.yarnpkg.com/tapwrite/-/tapwrite-1.1.21.tgz#0577a1e778c94896961db0e37627e496946c2e2b"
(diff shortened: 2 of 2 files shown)


src/app/detail/ui/Sidebar.tsx
```diff
@@ -79,7 +81,15 @@ export const Sidebar = ({
...
 
-  const matches = useMediaQuery('(max-width:600px)')
+  const windowWidth = useWindowWidth()
+  const isMobile = windowWidth < 600 && windowWidth !== 0
+
+  useEffect(() => {
+    if (isMobile) {
+      store.dispatch(setShowSidebar(false))
+    }
+  }, [isMobile])
```
priosshrsth, line 91: We seem to have sm breakpoint for this. I think using that would give you much better result.

src/app/detail/ui/Sidebar.tsx
```diff
@@ -88,7 +98,7 @@ export const Sidebar = ({
         borderLeft: (theme) => `1px solid ${theme.color.borders.border2}`,
         height: '100vh',
         display: showSidebar ? 'block' : 'none',
-        width: matches && showSidebar ? '100vw' : '25vw',
+        width: isMobile && showSidebar ? '100vw' : '25vw',
```
priosshrsth, line 101:
```suggestion
        width: showSidebar ? '100vw' : '25vw',
        md: {
         width '25vw'
        }
```

Something like this should be possible and would give you much better result as you would not always have window.width defined.

priosshrsth (changes requested):

src/hooks/useWindowWidth.ts
```diff
@@ -0,0 +1,18 @@
...
+  function handleResize() {
+    setWindowWidth(window.outerWidth)
+  }
+  React.useEffect(() => {
+    window.addEventListener('resize', handleResize)
+
+    handleResize()
+
+    return () => window.removeEventListener('resize', handleResize)
+  }, [])
```
priosshrsth, line 15: While this might be necessay, I suggest you to crete a useLessThanBreakPoint hook.

that utilizes theme.breakpoints.down(brekpoint) to get isLessThanSm. This would prevent too many re renders.

priosshrsth (changes requested):

PR #324, opened 2024-10-02
OUT-866 Image loading forever when trying to upload

arpandhakal:
### Tasks

- [Image loading forever when trying to upload](https://linear.app/copilotplatforms/issue/OUT-866/image-loading-forever-when-trying-to-upload)

### Changes

- [x] added error handling and error logs while image fails to upload on supabase

### Notes

If supabase fails to upload the given image due to internet connectivity or something else, then the image won't get stuck in loading state forever, added a rejection promise if error is returned from supabase. In the future, we need to apply chips for overall application's error management to show user that certain tasks have failed.

### DIFF
--- a/src/app/ui/NewTaskForm.tsx
+++ b/src/app/ui/NewTaskForm.tsx
@@ -363,7 +363,7 @@ const NewTaskFooter = ({ handleCreate, handleClose, getSignedUrlUpload }: NewTas
     if (files && files.length > 0) {
       const file = files[0]
       const signedUrl: ISignedUrlUpload = await getSignedUrlUpload(generateRandomString(file.name))
-      const filePayload = await supabaseActions.uploadAttachment(file, signedUrl, '')
+      const { filePayload, error } = await supabaseActions.uploadAttachment(file, signedUrl, '')
       if (filePayload) {
         store.dispatch(setCreateTaskFields({ targetField: 'attachments', value: [...attachments, filePayload] }))
       }
--- a/src/app/ui/NewTaskForm.tsx
+++ b/src/app/ui/NewTaskForm.tsx
@@ -364,6 +364,9 @@ const NewTaskFooter = ({ handleCreate, handleClose, getSignedUrlUpload }: NewTas
       const file = files[0]
       const signedUrl: ISignedUrlUpload = await getSignedUrlUpload(generateRandomString(file.name))
       const { filePayload, error } = await supabaseActions.uploadAttachment(file, signedUrl, '')
+      if (error) {
+        console.error('Failed to upload image') //handle error through chip here in the future
+      }
       if (filePayload) {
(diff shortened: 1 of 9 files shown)


src/utils/SupabaseActions.ts
```diff
@@ -29,7 +29,7 @@ export class SupabaseActions extends SupabaseService {
       .from(supabaseBucket)
       .uploadToSignedUrl(signedUrl.path, signedUrl.token, file)
     if (error) {
-      throw new APIError(httpStatus.BAD_REQUEST, error.message)
+      throw new APIError(httpStatus.BAD_REQUEST, 'could not upload file to supabase', error.message)
```
rrojan, line 32: SupabaseActions is called from a client components so using APIError doesn't work here. Instead you should manually `console.error` the error without throwing an exception - since throwing an error will crash the app completely.

src/utils/inlineImage.ts
```diff
@@ -11,12 +11,14 @@ import { getSignedUrlUpload, getSignedUrlFile } from '@/app/actions'
...
   const supabaseActions = new SupabaseActions()
-  const fileName = generateRandomString(file.name)
 
+  const fileName = generateRandomString(file.name)
   const signedUrl: ISignedUrlUpload = await getSignedUrlUpload(token, fileName)
-  const filePayload = await supabaseActions.uploadAttachment(file, signedUrl, task_id)
+  const { filePayload, error } = await supabaseActions.uploadAttachment(file, signedUrl, task_id)
   const url = await getSignedUrlFile(token ?? '', filePayload?.filePath ?? '')
-
+  if (error) {
```
rrojan, line 19: if token or filePayload.filePath is empty we don't want to get a signed url file from it in the first place. So handling it separately before this step would be the best to prevent unintended side effects

rrojan (changes requested):

src/app/ui/NewTaskForm.tsx
```diff
@@ -363,7 +363,7 @@ const NewTaskFooter = ({ handleCreate, handleClose, getSignedUrlUpload }: NewTas
     if (files && files.length > 0) {
       const file = files[0]
       const signedUrl: ISignedUrlUpload = await getSignedUrlUpload(generateRandomString(file.name))
-      const filePayload = await supabaseActions.uploadAttachment(file, signedUrl, '')
+      const { filePayload, error } = await supabaseActions.uploadAttachment(file, signedUrl, '')
```
priosshrsth, line 366: error has been added but, I see no usage has been added to the file.

priosshrsth (changes requested):

makeitraina:
I would love to understand what the problem here was. Could you leave a comment here or update PR description explaining

aatbip (approved):
lgtm

PR #328, opened 2024-10-03
OUT-833 'Required' error is not shown for the 'Assignee' drop-down and able to create task

arpandhakal:
### Tasks

- ['Required' error is not shown for the 'Assignee' drop-down and able to create task](https://linear.app/copilotplatforms/issue/OUT-833/required-error-is-not-shown-for-the-assignee-drop-down-and-able-to)
- [MaxWidth is not right on images on smaller screen when screen is changed from devtools. Image max width greater in task creation.](https://linear.app/copilotplatforms/issue/OUT-861/maxwidth-is-not-right-on-images-on-smaller-screen-when-screen-is)

### Changes

- [x] cherry picked updates which were already solved in OUT-791

### DIFF
--- a/src/hooks/useHandleSelectorComponent.tsx
+++ b/src/hooks/useHandleSelectorComponent.tsx
@@ -3,7 +3,7 @@ import { setCreateTaskFields } from '@/redux/features/createTaskSlice'
 import { setCreateTemplateFields } from '@/redux/features/templateSlice'
 import store from '@/redux/store'
 import { WorkflowStateResponse } from '@/types/dto/workflowStates.dto'
-import { IAssigneeCombined, handleSelectorComponentModes } from '@/types/interfaces'
+import { IAssigneeCombined, HandleSelectorComponentModes } from '@/types/interfaces'
 import { getAssigneeTypeCorrected } from '@/utils/getAssigneeTypeCorrected'
 import { useEffect, useState } from 'react'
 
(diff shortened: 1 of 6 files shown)


src/hooks/useHandleSelectorComponent.tsx
```diff
@@ -21,7 +29,7 @@ export const useHandleSelectorComponent = ({ item, type }: { item: unknown; type
       store.dispatch(setCreateTemplateFields({ targetField: 'workflowStateId', value: (item as WorkflowStateResponse)?.id }))
     }
 
-    if (type === SelectorType.ASSIGNEE_SELECTOR && item) {
+    if ((mode === handleSelectorComponentModes.CreateTaskFieldUpdate, type === SelectorType.ASSIGNEE_SELECTOR) && item) {
```
rrojan, line 32: `mode === handleSelectorComponentModes.CreateTaskFieldUpdate,` 

What's up with the comma here? I'm not sure if this is valid JS...

src/types/interfaces.ts
```diff
@@ -58,6 +58,11 @@ export enum FilterOptionsKeywords {
   TEAM = 'ius',
 }
 
+export enum handleSelectorComponentModes {
+  CreateTaskFieldUpdate,
```
rrojan, line 62: In typescript enums are generally written in PascalCase since enums are not a core-JS contruct and are more similar to types

rrojan, line 62: We shouldn't be adding an enum type to an interfaces file imo, since we are grouping them based on their behavior

rrojan (changes requested):

rrojan (approved):
LGTM :trophy:

PR #329, opened 2024-10-04
OUT-821 | Opening "Filter by assignee" drop-down for second time shows incorrect assignees

rrojan:
### Tasks

- [OUT-821 | Opening "Filter by assignee" drop-down for second time shows incorrect assignees](https://linear.app/copilotplatforms/issue/OUT-821/opening-filter-by-assignee-drop-down-for-second-time-shows-incorrect)

### Changes

- [x] Fix issue where changing filter option for second time showed incorrect assignee state array 
- [x]  Remove unused imports & Organize imports via TS lsp
- [x] Fixed another pending issue where pending fetch updates our assignee list, even after the selector is clicked away from / unfocused 

### Testing Criteria

- [x] Video:
[Screencast from 2024-09-24 13-43-24.webm](https://github.com/user-attachments/assets/1f913fc1-6d17-4662-a6f4-6a366ea4ab2b)

### DIFF
--- a/src/components/layouts/FilterBar.tsx
+++ b/src/components/layouts/FilterBar.tsx
@@ -17,7 +17,7 @@ import { FilterOptions, FilterOptionsKeywords, IAssigneeCombined, IFilterOptions
 import { filterOptionsToAssigneeMap, filterTypeToButtonIndexMap } from '@/types/objectMaps'
 import { checkAssignee } from '@/utils/assignee'
 import { NoAssigneeExtraOptions } from '@/utils/noAssignee'
-import { setDebouncedFilteredAssignees } from '@/utils/users'
+import { getDebouncedFilteredAssignees, setDebouncedFilteredAssignees } from '@/utils/users'
 import { Box, IconButton, Stack } from '@mui/material'
 import { useEffect, useState } from 'react'
 import { useSelector } from 'react-redux'
@@ -44,6 +44,10 @@ export const FilterBar = ({ mode, updateViewModeSetting }: FilterBarProps) => {
   useEffect(() => {
     initialAssignees.length === 0 && setInitialAssignees(filteredAssignee)
   }, [initialAssignees.length, filteredAssignee])
+
+  useEffect(() => {
+    console.log('aaa load', loading)
+  }, [loading])
 
   const handleFilterOptionsChange = async (optionType: FilterOptions, newValue: string | null) => {
     store.dispatch(setFilterOptions({ optionType, newValue }))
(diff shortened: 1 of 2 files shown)


src/components/layouts/FilterBar.tsx
```diff
@@ -1,60 +1,86 @@
...
+      setLoading(false)
+      return
+    }
+
+    // If the selector is clicked away from, set loading state to false right away
+    setLoading(false)
+    // Reset selector to have initial assignee list. DO NOT use filteredAssignee as it can cause inconsistencies after filteroption is changed
+    setFilteredAssignee(initialAssignees)
+  }
+
```
aatbip, line 76: It seems that the `Selector` component for filtering by assignee has accumulated a significant amount of state management and logic. Given its current complexity, it would be beneficial to refactor the entire Selector component, along with its associated logic, into a separate module. Making this change now would be ideal since the implementation is still fresh in our minds, and delaying it could complicate future refactoring efforts as the context might be lost over time. What are your thoughts on proceeding with this refactor?

aatbip (changes requested):
I think there is an issue with the search functionality. I don't get the result for the search query as expected and it happens often. Please see the below video - 

[Screencast from 07-10-24 08:14:35 AM +0545.webm](https://github.com/user-attachments/assets/c1b9cf9d-c75f-4977-a952-a1518c2406f4)

src/components/layouts/FilterBar.tsx
rrojan: There seems to be a considerable amount of tech debt on the frontend side but refactoring the selector component right now might not be the best approach. We have a ton of priorities for M3, but more importantly we have no idea / confirmation on the new multiple select component. Refactoring could be best done after we have confirmation on that @aatbip

aatbip: I am sorry for the misunderstanding. I didn't mean to write that we refactor the `Selector` component but I meant that we move the filter by assignee `Selector` component implementation from this file to a separate module so that there will be separation of concern. 
Regarding refactoring the `Selector` component, I agree that we should have a better confirmation on it. Thanks for considering. @rrojan

rrojan: @aatbip Ah I see. Solid point!

Still I'm hesitant to refactor this as of right now though... refactoring doesn't make much sense unless we change a lot of code around, and we don't want this to introduce new bugs right now because of it.

aatbip: I completely agree with your point of view. I was suggesting and looking forward to know your thoughts. Thank you for considering. @rrojan

PR #333, opened 2024-10-07
OUT-861 MaxWidth is not right on images on smaller screen when screen is changed from devtools. Image max width greater in task creation.

arpandhakal:
### Tasks

- [MaxWidth is not right on images on smaller screen when screen is changed from devtools. Image max width greater in task creation.](https://linear.app/copilotplatforms/issue/OUT-861/maxwidth-is-not-right-on-images-on-smaller-screen-when-screen-is)

### Changes

- [x] recalculated maxWidth of images on resize observer too, so that if we insert bigger images then the max width would be calculated without considering the width of the scrollbar that came in task creation modal.
- [COMMIT](https://github.com/aatbip/tapwrite/pull/12/commits/ef819e1214b12e2415a3650ad93e4718074cf117)

### DIFF
--- a/src/hoc/ModifiedBackend.ts
+++ b/src/hoc/ModifiedBackend.ts
@@ -3,7 +3,7 @@ import { HTML5Backend } from 'react-dnd-html5-backend'
 import { TouchBackend } from 'react-dnd-touch-backend'
 
 const shouldIgnoreTarget = (target: any) => {
-  return target.closest('.ProseMirror') !== null
+  return target.closest('.tiptap') !== null
 }
 
 const createModifiedBackend = (Backend: BackendFactory, manager?: any, context?: any) => {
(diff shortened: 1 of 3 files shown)


src/hoc/ModifiedBackend.ts
```diff
@@ -3,7 +3,7 @@ import { HTML5Backend } from 'react-dnd-html5-backend'
 import { TouchBackend } from 'react-dnd-touch-backend'
 
 const shouldIgnoreTarget = (target: any) => {
-  return target.closest('.ProseMirror') !== null
+  return target.closest('.tiptap') !== null
```
aatbip, line 6: Any particular reason for changing `ProseMirror` class to `tiptap`?

aatbip (approved):
lgtm

src/hoc/ModifiedBackend.ts
arpandhakal: they both lie in the same document. So, I think tiptap is more recognizable, so.

PR #336, opened 2024-10-09
OUT-887 Issues with inline images on real time

arpandhakal:
### Tasks

- [Issues with inline images on real time](https://linear.app/copilotplatforms/issue/OUT-887/issues-with-inline-images-on-real-time)

### Changes

- [x] fixed : Image loads again if multiple users open the task details for the same task
- [x] fixed the problem of placeholder being bigger than the image itself.


src/app/detail/ui/TaskEditor.tsx
```diff
@@ -163,7 +165,12 @@ export const TaskEditor = ({
           readonly={userType === UserType.CLIENT_USER}
           editorClass="tapwrite-details-page"
           placeholder="Add description..."
-          uploadFn={(file) => uploadImageHandler(file, token ?? '', task_id)}
+          uploadFn={async (file) => {
+            setIsUserTyping(true)
+            const url = await uploadImageHandler(file, token ?? '', task_id)
+            setIsUserTyping(false)
```
rrojan, line 171: What if the user types a sentence while upload is in progress?

Looking at the previous implementation of `handleDetailChange` I noticed we could have edge cases here so I tested it out... image upload disappears if I type something while waiting for the image to upload.

[Screencast from 2024-10-09 18-19-20.webm](https://github.com/user-attachments/assets/7cd9e4bd-4e81-4a59-a1c3-ed0fb93d2bf7)

rrojan (changes requested):

src/app/detail/ui/TaskEditor.tsx
rrojan: As i mentioned in the slack message, there are other consequences to this as well. Another one I can think of is uploading two or more images, where settings isUserTyping to false from one callback will affect the rest.

src/hoc/RealTime.tsx
```diff
@@ -48,6 +51,15 @@ export const RealTime = ({ children, task }: { children: ReactNode; task?: TaskR
           }
           //if the task is updated
         } else {
+          if (oldTask && oldTask.body && updatedTask.body) {
+            const oldImgSrcs = extractImgSrcs(oldTask.body)
+            const newImgSrcs = extractImgSrcs(updatedTask.body)
+
+            if (oldImgSrcs.length > 0 && newImgSrcs.length > 0) {
+              updatedTask.body = replaceImgSrcs(updatedTask.body, newImgSrcs, oldImgSrcs)
```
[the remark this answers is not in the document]
arpandhakal: Need to extract new image Srcs and replace it with old ones, because since we are creating a new url of images on each task details navigation, a second user navigating the task details will generate a new src and replace it in the database which causes the previous user to load the src again(because its new)

Added a comment in the code block. thanks.

PR #338, opened 2024-10-16
re-OUT-833 'Required' error is not shown for the 'Assignee' drop-down and able to create task

arpandhakal:
### Tasks

- ['Required' error is not shown for the 'Assignee' drop-down and able to create task](https://linear.app/copilotplatforms/issue/OUT-833/required-error-is-not-shown-for-the-assignee-drop-down-and-able-to)

### Changes

- [x] missed code block causing issues in this merged PR. Removed a condition which was blocking the assigneeId in store to reset back to null. Ref : comment on the task.

### Testing Criteria

- [LOOM](https://www.loom.com/share/6516af554d294832ae10bba616d19a98)


aatbip (approved):
lgtm

PR #339, opened 2024-10-16
OUT-893,OUT-862 Activity log (activity + comment) Frontend Implementation

aatbip:
### Changes

- [x] Brings back activity log and comments
- [x] Adds polling support for activity logs realtime updates

### Testing Criteria
- [x] DEMO
[Screencast from 16-10-24 04:42:15 PM +0545.webm](https://github.com/user-attachments/assets/4464a1df-86c2-41f7-a0c0-4ec7fdb9512d)


src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -40,8 +39,8 @@ import { CustomLink } from '@/hoc/CustomLink'
 import { DetailStateUpdate } from '@/app/detail/[task_id]/[user_type]/DetailStateUpdate'
 import { SilentError } from '@/components/templates/SilentError'
 import { z } from 'zod'
-import { ScrapImageRequest } from '@/types/common'
 import { signedUrlTtl } from '@/types/constants'
+import { ActivityWrapper } from '../../ui/ActivityWrapper'
```
Comment (arpandhakal) [inline src/app/detail/[task_id]/[user_type]/page.tsx:43]: please use alias imports here

src/app/detail/ui/ActivityWrapper.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { postComment, deleteComment } from '../[task_id]/[user_type]/actions'
+import { CreateComment } from '@/types/dto/comment.dto'
+import { fetcher } from '@/utils/fetcher'
+import useSWR from 'swr'
+
+export const ActivityWrapper = ({ token, task_id }: { token: string; task_id: string }) => {
+  const { data: activities, isLoading } = useSWR(`/api/tasks/${task_id}/activity-logs/?token=${token}`, fetcher, {
+    refreshInterval: 1000,
+  })
+  console.log('activities', activities)
```
arpandhakal, line 18: remove logs

arpandhakal (changes requested):

src/app/detail/ui/ActivityWrapper.tsx
```diff
@@ -0,0 +1,70 @@
+'use client'
+
+import { LogResponse } from '@/app/api/activity-logs/schemas/LogResponseSchema'
+import { Box, Skeleton, Stack, Typography } from '@mui/material'
+import { Comments } from './Comments'
+import { ActivityLog } from './ActivityLog'
```
rrojan, line 6: Prefer aliased imports, also consider proper import arrangement using `vtsls` or `tsserver`

src/app/detail/ui/ActivityWrapper.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { CommentInput } from '@/components/inputs/CommentInput'
+import { ActivityType } from '@prisma/client'
+import { postComment, deleteComment } from '../[task_id]/[user_type]/actions'
+import { CreateComment } from '@/types/dto/comment.dto'
+import { fetcher } from '@/utils/fetcher'
+import useSWR from 'swr'
+
+export const ActivityWrapper = ({ token, task_id }: { token: string; task_id: string }) => {
+  const { data: activities, isLoading } = useSWR(`/api/tasks/${task_id}/activity-logs/?token=${token}`, fetcher, {
+    refreshInterval: 1000,
```
rrojan, line 16: 1 seconds is too low for a polling implementation. There will be multiple concurrent requests on top of each other if the requests are not returned within 1 second.
Something like 10-20 seconds is more viable

src/hoc/ClientSideStateUpdate.tsx
```diff
@@ -87,6 +90,10 @@ export const ClientSideStateUpdate = ({
       const updatedTasks = tasksInStore.map((t) => (t.id === task.id ? task : t))
       store.dispatch(setTasks(updatedTasks))
     } //for updating a task in store with respect to task response from db in task details page
+
+    if (activities) {
+      store.dispatch(setActivities(activities))
+    }
```
rrojan, line 96: Why are we storing activities in a redux global state if this is only needed in a particular section?

src/utils/fetcher.ts
```diff
@@ -0,0 +1,5 @@
+export const fetcher = (url: string | null) => {
+  if (url) {
+    return fetch(url).then((res) => res.json())
+  }
+}
```
rrojan, line 5: Is there a reason why the argument `url` is nullable? We should not be handling this in a reusable util.

src/utils/fetcher.ts
```diff
@@ -0,0 +1,5 @@
+export const fetcher = (url: string | null) => {
+  if (url) {
+    return fetch(url).then((res) => res.json())
```
rrojan, line 3: Implement proper `catch`, even if it is just console.error'ing the response.

In the future this will help us in implementing error messages.

rrojan (changes requested):

src/hoc/ClientSideStateUpdate.tsx
```diff
@@ -87,6 +90,10 @@ export const ClientSideStateUpdate = ({
       const updatedTasks = tasksInStore.map((t) => (t.id === task.id ? task : t))
       store.dispatch(setTasks(updatedTasks))
     } //for updating a task in store with respect to task response from db in task details page
+
+    if (activities) {
+      store.dispatch(setActivities(activities))
```
rrojan, line 95: I don't see activities being passed as a prop to `ClientSideStateUpdate` anywhere(?)

src/app/detail/ui/ActivityWrapper.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { ActivityLog } from './ActivityLog'
+import { CommentInput } from '@/components/inputs/CommentInput'
+import { ActivityType } from '@prisma/client'
+import { postComment, deleteComment } from '../[task_id]/[user_type]/actions'
+import { CreateComment } from '@/types/dto/comment.dto'
+import { fetcher } from '@/utils/fetcher'
+import useSWR from 'swr'
+
+export const ActivityWrapper = ({ token, task_id }: { token: string; task_id: string }) => {
+  const { data: activities, isLoading } = useSWR(`/api/tasks/${task_id}/activity-logs/?token=${token}`, fetcher, {
```
rrojan, line 15: As discussed in our standup about leaving tasks app open all day and it hitting our server at regular intervals throughout that time, we should disable SWR fetches when the tab is not in focus.

An easy way to do this is to set `refreshWhenHidden` to `true`.

rrojan (changes requested):

src/utils/fetcher.ts
aatbip: this is the type expected by `fetcher` function at `useSWR` hook

src/hoc/ClientSideStateUpdate.tsx
aatbip: it is not needed anymore. will remove.

rrojan:
@aatbip DB issue has been fixed :fire:

src/app/detail/ui/ActivityWrapper.tsx
aatbip: The purpose to set 1 second polling interval time is because it is strictly mentioned in the ticket that it should be nearly realtime "Remember that this has to be near real-time", and while using polling, this is the best way to achieving near real time by increasing the amount of polling.

rrojan (approved):
LGTM :trophy:

PR #341, opened 2024-10-17
OUT-821 | Opening "Filter by assignee" drop-down for second time shows incorrect assignees

rrojan:
### Tasks

- [OUT-821 | Opening "Filter by assignee" drop-down for econd time shows incorrect assignees](https://linear.app/copilotplatforms/issue/OUT-821/opening-filter-by-assignee-drop-down-for-second-time-shows-incorrect)

### Changes

- [x] Fix issue where changing filter option for second time showed incorrect assignee state array 
- [x] Fixed another pending issue where pending fetch updates our assignee list, even after the selector is clicked away from / unfocused
- [x]  Remove few unused imports and minor refactors

### Testing Criteria

- [x] Video:
[Screencast from 2024-09-24 13-43-24.webm](https://github.com/user-attachments/assets/1f913fc1-6d17-4662-a6f4-6a366ea4ab2b)

### DIFF
--- a/src/components/inputs/Selector.tsx
+++ b/src/components/inputs/Selector.tsx
@@ -1,4 +1,4 @@
-import { Avatar, Box, Button, Popper, Stack, Typography } from '@mui/material'
+import { Box, Button, Popper, Stack, Typography } from '@mui/material'
 import { StyledAutocomplete } from '@/components/inputs/Autocomplete'
 import { statusIcons } from '@/utils/iconMatcher'
 import { useFocusableInput } from '@/hooks/useFocusableInput'
@@ -7,7 +7,6 @@ import { StyledTextField } from './TextField'
 import { WorkflowStateResponse } from '@/types/dto/workflowStates.dto'
 import { IAssigneeCombined, IExtraOption, ITemplate, UserTypesName } from '@/types/interfaces'
 import { TruncateMaxNumber } from '@/types/constants'
-
 import { truncateText } from '@/utils/truncateText'
 import { CopilotAvatar } from '@/components/atoms/CopilotAvatar'
 import { getAssigneeName } from '@/utils/assignee'
@@ -28,6 +27,8 @@ interface Prop {
   selectorType: SelectorType
   options: unknown[]
   buttonContent: ReactNode
+  inputStatusValue: string
+  setInputStatusValue: React.Dispatch<React.SetStateAction<string>>
   disabled?: boolean
   placeholder?: string
   extraOption?: IExtraOption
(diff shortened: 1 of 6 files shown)


src/components/layouts/FilterBar.tsx
```diff
@@ -47,16 +44,41 @@ export const FilterBar = ({ mode, updateViewModeSetting }: FilterBarProps) => {
...
+    if (filteredAssignee.length && initialAssignees.length && !inputStatusValue) {
+      setFilteredAssignee(initialAssignees)
+    }
+  }, [filteredAssignee, initialAssignees, inputStatusValue])
+
+  useEffect(() => {
+    if (!inputStatusValue) {
+      loading && setLoading(false)
+    }
+  }, [inputStatusValue])
```
aatbip, line 71: I think it is not a good idea to bring forward a lot of effects with similar dependency. Can we try to merge the logic into a single useEffect? I think the below suggestion would work - 

```javascript
useEffect(() => {
  // Base these initial values off of first fetch of filteredAssignee
  if (initialAssignees.length === 0) {
    setInitialAssignees(filteredAssignee);
  }

  // When focus is taken away from selector, reset filteredAssignee if needed
  if (filteredAssignee.length && initialAssignees.length && !inputStatusValue) {
    setFilteredAssignee(initialAssignees);
  }

  // Ensure loading state is reset if inputStatusValue is falsy
  if (!inputStatusValue && loading) {
    setLoading(false);
  }
}, [initialAssignees, filteredAssignee, inputStatusValue, loading]);

```

src/components/inputs/Selector.tsx
```diff
@@ -82,8 +85,6 @@ export default function Selector({
   const open = Boolean(anchorEl)
   const id = open ? 'autocomplete-popper' : ''
 
-  const [inputStatusValue, setInputStatusValue] = useState('')
```
aatbip, line 85: Is there any particular reason for us to remove this from the component? I think it is better if we maintain this state internally in the component scope.

aatbip (changes requested):

src/components/inputs/Selector.tsx
rrojan: The `FilterBar.tsx` component requires us to track the value of selector text value to see where it resets back to an empty string. This was extracted out because hooks like `useState` cannot be called called conditionally so we can't just have that for the FilterBar only.

I tried to use a ref before this to extract the text value but that was lagging a render behind as well.

aatbip (approved):
lgtm

PR #343, opened 2024-10-17
OUT-890 | IU/Client can see the tasks that are assigned to another IU/client in real-time

rrojan:
### Tasks

- [OUT-890 | IU/Client can see the tasks that are assigned to another IU/client in real-time](https://linear.app/copilotplatforms/issue/OUT-890/iuclient-can-see-the-tasks-that-are-assigned-to-another-iuclient-in)

### Changes

- [x] Filter out tasks that a client does not have access to in real-time
- [x] Change single real-time filter for clients to filter by assigneeId instead of workspaceId because:

1. Client users aren't streamed tasks that belong to the workspace but are not assigned to them, so it is more secure.
2. There are a LOT less change events to listen and react to.

### Testing Criteria

- [x] Screencast:
[Screencast from 2024-10-17 17-50-04.webm](https://github.com/user-attachments/assets/c9843358-a74f-4c9d-bd42-1cda6344c2d6)

### DIFF
--- a/src/hoc/RealTime.tsx
+++ b/src/hoc/RealTime.tsx
@@ -90,7 +90,17 @@ export const RealTime = ({ children, task }: { children: ReactNode; task?: TaskR
         'postgres_changes',
         // Because of the way supabase realtime is architected for postgres_changes, it can only apply one filter at a time.
         // Ref: https://github.com/supabase/realtime-js/issues/97
-        { event: '*', schema: 'public', table: 'Tasks', filter: `assigneeId=eq.${userId}` },
+        {
+          event: '*',
+          schema: 'public',
+          table: 'Tasks',
+          filter:
+            userRole === AssigneeType.internalUser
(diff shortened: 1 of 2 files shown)


src/hoc/RealTime.tsx
```diff
@@ -69,11 +84,28 @@ export const RealTime = ({ children, task }: { children: ReactNode; task?: TaskR
...
+        {
+          event: '*',
+          schema: 'public',
+          table: 'Tasks',
+          filter:
+            userRole === AssigneeType.internalUser
+              ? `workspaceId=eq.${tokenPayload?.workspaceId}`
+              : // The reason we are explicitly using an assigneeId filter for clients is so they are not streamed
+                // tasks they don't have access to in the first place.
+                `assigneeId=eq.${userId}`,
```
makeitraina, line 107: one thing to consider here is tasks assigned to copmanies

aatbip (approved):
lgtm

PR #346, opened 2024-10-18
OUT-898 | Task is not being set in Redux on redirects

rrojan:
### Tasks

- [OUT-898 | Task is not being set in Redux on redirects](https://linear.app/copilotplatforms/issue/OUT-898/task-is-not-being-set-in-redux-on-redirects)

### Changes

- [x] Fix `task` state overwriting and removing entire `tasks` list in redux slice

### Testing Criteria

- [x] `isRedirect` is working fine now. SS:
image
- [x] Checked for regressions in normal task load - everything is working fine

### DIFF
--- a/src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx
+++ b/src/app/detail/[task_id]/[user_type]/DetailStateUpdate.tsx
@@ -34,7 +34,6 @@ export const DetailStateUpdate = async ({ isRedirect, token, tokenPayload, task,
       token={token}
       viewSettings={viewSettings}
       tokenPayload={tokenPayload}
-      task={task}
     >
       {children}
     </ClientSideStateUpdate>


sazanrjb (approved):

makeitraina:
Nice, thanks for doing so quickly. Im just curious, do we know how this regressed? Part of M1 as well or because of some other change in M2 branch, if so, do you know which one that was?

rrojan:
@makeitraina Yes, it seems as a part of M2 ([OUT-776](https://linear.app/copilotplatforms/issue/OUT-776/add-back-support-for-inline-images)) @arpandhakal added a new redux state called `task` which would overwrite our existing `tasks` array state to make it easier for him to track the task passed to Tiptap.

However this caused a regression since passing it would overwrite our initially fetched tasks

PR #348, opened 2024-10-21
OUT 905, OUT-906, OUT-918 Comment takes too long to appear in UI once created

aatbip:
### Changes

- [x] Adds optimistic updates to create and delete comments

### Testing Criteria
- [LOOM](https://www.loom.com/share/75db1a3af40b46a19eebc5cc6c6d04a9)


rrojan (commented):
The comments are flickering constantly after posting a comment. Can you please look into this

src/app/detail/ui/ActivityWrapper.tsx
```diff
@@ -10,29 +10,108 @@ import { postComment, deleteComment } from '../[task_id]/[user_type]/actions'
...
+    }
+
+    // Optimistically update the SWR cache
+    const cacheKey = `/api/tasks/${task_id}/activity-logs/?token=${token}`
 
-  const [isFirstPageLoad, setIsFirstPageLoad] = useState(true)
+    // Optimistic UI update
+    mutate(
+      cacheKey,
+      (currentData: { data: (LogResponse | TempCommentType)[] } | undefined) => {
```
makeitraina, line 66: hmmm... tbh I dnt understand why we need `TempCommentType` and cant use `LogResponse` directly. But Im probably just missing some context. This comment is not blocking review/merge, just adding my 2 cents

makeitraina:
I think we are doing the optimistic update incorrectly. The swr docs show that you only need to call `mutate` once and `optimisticData` is a property in the options passed to mutate. Please correct me if Im wrong about this though. Again adding my 2 cents here and not blocking review.

src/app/detail/ui/ActivityWrapper.tsx
aatbip: We can make use of `LogResponse` but need to handle a lot of fields and also need to probably change the current type of `LogResponse`.  I chose to create a new `tempComment` assignment with `TempCommentType` structure to minimize the unnecessary need to handle multiple fields that are not wanted for a short period of optimistic update until a new mutation takes place. @makeitraina

aatbip:
> I think we are doing the optimistic update incorrectly. The swr docs show that you only need to call `mutate` once and `optimisticData` is a property in the options passed to mutate. Please correct me if Im wrong about this though. Again adding my 2 cents here and not blocking review.

SWR allows both the use of the `mutate` function with a callback (like in the current implementation) and using the `optimisticData` option. The method that is used provides flexibility, allowing us to mutate the cache immediately before the actual request completes. This callback-based method can be useful when we want finer control over how the data is updated, such as combining or filtering complex data structures like arrays.

While using `optimisticData` is a shorthand and simpler in some cases, the callback pattern offers more customization and is equally correct according to the SWR docs.

rrojan:
Please make sure to account for all edge cases here. I can post empty comments.
[Screencast from 2024-10-22 20-00-30.webm](https://github.com/user-attachments/assets/9418a24e-9344-4a85-a8ca-3da780b05d6f)

makeitraina:
> > I think we are doing the optimistic update incorrectly. The swr docs show that you only need to call `mutate` once and `optimisticData` is a property in the options passed to mutate. Please correct me if Im wrong about this though. Again adding my 2 cents here and not blocking review.
> 
> SWR allows both the use of the `mutate` function with a callback (like in the current implementation) and using the `optimisticData` option ([refer here](https://swr.vercel.app/docs/mutation#mutate)). The method that is used provides flexibility, allowing us to mutate the cache immediately before the actual request completes. This callback-based method can be useful when we want finer control over how the data is updated, such as combining or filtering complex data structures like arrays.
> 
> While using `optimisticData` is a shorthand and simpler in some cases, the callback pattern offers more customization and is equally correct according to the SWR docs.

Of course,  the mutate can be done with a callback, but the whole point of doing mutate with the api call and optimistic update in one function is so that you don't have to have all the extra boiler plate and try catch logic. This code would look a lot simpler and more robust if it was done directly with the optimistic update. There are even a few more options that are provided to make this work:
1. optimisticData can accept a function, so you can modify the existing array
2. setting `rollbackOnError: true` automatically handles the error and rolling back, it knows exactly what to rollback
3. [populateCache](https://swr.vercel.app/docs/mutation#update-cache-after-mutation) lets you update the cache from the response.

src/app/detail/ui/ActivityWrapper.tsx
makeitraina: is it that complicated? I think the defaults can be set pretty easily right?

makeitraina:
@aatbip @rrojan whats the update on this PR? i thought we discussed making sure we get this one completed in standup but its open. The majority of PRS have been merged against this branch so in our team testing session today we will be missing many fixes.

src/app/detail/ui/ActivityWrapper.tsx
aatbip: changed added in the latest commit @makeitraina

rrojan (approved):
Good job on the new updates. LGTM, but I think having @priosshrsth go through this once again would be great since I'm not too familiar with the nuances of `SWR`

src/app/detail/ui/ActivityWrapper.tsx
```diff
@@ -9,30 +9,102 @@ import { ActivityType } from '@prisma/client'
...
+        async () => {
+          // Post the actual comment to the server
+          await postComment(token, postCommentPayload)
+          // Return the actual updated data (this will trigger revalidation)
+          return await fetcher(cacheKey)
+        },
+        {
+          optimisticData: { data: optimisticData },
+          rollbackOnError: true,
+          revalidate: true, // Make sure to revalidate after mutation
```
makeitraina, line 77: this is fine for now, not blocking merging this PR, but I dont think you have to do revalidate. This is what [populateCache](https://swr.vercel.app/docs/mutation#update-cache-after-mutation) is for.

makeitraina (approved):
looking good, one testing and reviewed by your team lets merge it!!

PR #351, opened 2024-10-21
OUT 913 Fix pluralization issues with elapsed time in activity feed and comments

aatbip:
### Changes

- [x] Adds a strong pluraliation check on top of that provided by `dayjs`


rrojan (approved):
This approach works fine, but can you please explore if `dayjs` has some plugins for it? i remember working with something like this some time ago

PR #352, opened 2024-10-21
OUT 921 Add hover state to properties section in task details

aatbip:
### Changes

- [x] Adds hover on sidebar properties


rrojan:
I think we aren't implementing the hover state correctly here:
image
image

Why is the label also being highlighted here

rrojan:
Also it seems we don't need `cursor: pointer` here according to the cursor in the figma design
image

rrojan (approved):
LGTM :trophy:

PR #362, opened 2024-10-23
OUT 928 Change hover states based on IU/CU

aatbip:
### Changes

- [x] Fixes hover states for IU/CU in the sidebar properties 

### Testing Criteria

- CU should only see hover state on Status 
- IU should only see hover state on Status and Due Date


aatbip:
Continued in #364.

PR #365, opened 2024-10-23
OUT-950 | OUT-923 | OUT-912 | OUT-911 | Notification Fixes for Tasks main

rrojan:
### Changes

- [x] Fix issue where product notification does not decrement for CU when IU completes a task
- [x] Fix issue with short-circuited IU notifications logic, which was blocking in-product notification creation  

### Testing Criteria

- [x] Screencast for OUT-923 | 
[Screencast from 2024-10-23 17-37-26.webm](https://github.com/user-attachments/assets/1ae67d5b-c790-4380-a706-368fef24007d)

- [x] Screencast for OUT-912 | 
[Screencast from 2024-10-23 17-34-35.webm](https://github.com/user-attachments/assets/ee7ff441-bf0b-4247-8be2-245965a167e2)

- [x] Screencast for OUT-911 | 
[Screencast from 2024-10-23 17-30-58.webm](https://github.com/user-attachments/assets/7e0cfcc1-cd3f-49bc-adcf-e6e61798c780)

- [x] Also tested most of the surface area for other regressions here

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -379,9 +379,15 @@ export class TasksService extends BaseService {
     // --------------------------
     // --- Notifications Logic
     // --------------------------
+
+    // Cases:
+    // 1. Task has been moved back to a non-complete state from completed for client task
+    // 2. Task has been moved back to a non-complete state from completed for company task
+    // 3. Task has been moved to complete state for client task
+    // 4. Task has been moved to complete state for company task
     const notificationService = new NotificationService(this.user)
 
-    // If task has been moved back to another non-completed state from Completed
+    // Case 1 & 2 | If task has been moved back to another non-completed state from Completed
     if (
       updatedWorkflowState &&
       updatedWorkflowState.type !== StateType.completed &&
@@ -391,7 +397,7 @@ export class TasksService extends BaseService {
       await this.sendTaskCreateNotifications({ ...updatedTask, workflowState: updatedWorkflowState })
     }
 
-    // If task has been moved to completed from a non-complete state, remove all notification counts
+    // Case 3 & 4 | If task has been moved to completed from a non-complete state, remove all notification counts
     if (updatedWorkflowState?.type === StateType.completed && prevTask.workflowState.type !== StateType.completed) {
       if (updatedTask.assigneeType === AssigneeType.company) {
         const copilot = new CopilotAPI(this.user.token)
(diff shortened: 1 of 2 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -529,24 +541,29 @@ export class TasksService extends BaseService {
...
       if (updatedTask.assigneeType === AssigneeType.internalUser) {
-        await notificationService.create(NotificationTaskActions.CompletedByIU, updatedTask, { email: true })
+        shouldCreateNotification &&
+          (await notificationService.create(NotificationTaskActions.CompletedByIU, updatedTask, { email: true }))
         // TODO: Clean code and handle notification center notification deletions here instead
       } else if (updatedTask.assigneeType === AssigneeType.company) {
         // Don't do this in parallel since this can cause rate-limits, each of them has their own bottlenecks for avoiding ratelimits
-        await notificationService.create(NotificationTaskActions.CompletedForCompanyByIU, updatedTask, { email: true })
+        shouldCreateNotification &&
+          (await notificationService.create(NotificationTaskActions.CompletedForCompanyByIU, updatedTask, { email: true }))
```
sazanrjb, line 562: so just to understand with the context:
> Don't send task notifications if the IU created the task themselves

And if the IU is the assignee, and s/he completes the task, then do we need to send the notification to himself/herself? or is it skipped in the notificationService?

src/app/api/tasks/tasks.service.ts
```diff
@@ -529,24 +541,29 @@ export class TasksService extends BaseService {
...
         await notificationService.markAsReadForAllRecipients(updatedTask)
       } else if (updatedTask.assigneeType === AssigneeType.client) {
+        shouldCreateNotification &&
```
sazanrjb, line 565: so if the IU is the creator of the task,
and assignee is the client, then we dont send NotificationTaskActions.CompletedByIU?

src/app/api/tasks/tasks.service.ts
rrojan: When a notification is completed by IU, the notification are sent to task creator.

In the case that task creator is the IU themselves, this is skipped by the `shouldCreateNotification` initialization
image

Else the notification is sent, but the creator is confirmed not to be the IU themselves

src/app/api/tasks/tasks.service.ts
rrojan: There are two ways that can happen in this context.

1. IU assigns the task to client, but that IU / another IU completes the task. If the task is created by the IU themselves, it is caught by `shouldCreateNotification` flag check above. If the task is marked as done by some other IU, the notification says that this client task has been completed by IU, hence `NotificationTaskActions.CompletedByIU`

2. This client, or another company client completes this task. This is handled by another service to keep concerns separate and isn't affected here. I did verify this behavior again in this PR, and it is working as intended.

rrojan: We aren't sending new notifications to client when task is completed. Just removing it from his task notifications count.
See the usage of `markClientNotificationAsRead` here
image

sazanrjb (approved):
Approving this. 

A lot is going on in the method. May be we can simplify by breaking them down based on different conditions. No need to do this now.

PR #367, opened 2024-10-24
OUT 951 Images load infinitely when added to task description

aatbip:
### Changes

- [x] Adds a clone of `Tapwrite` with the name `TippyTappy` to solve the conflict of tiptap instances issue
- [x] Adds a `activeUploads` counter state that checks if there is an active upload in place. This prevents the component while being rerendered while upload/s are taking place


rrojan (commented):
@aatbip Please squash these "test" commits into one if you want the previous `fix` commits to remain. Or you can just squash this PR on merge.

rrojan (changes requested):
The upload functionality is working now, but I see that while two users both have tasks app open, the page for the user who receives a new realtime update suddenly scrolls to the top multiple times.

[Screencast from 2024-10-24 16-33-57.webm](https://github.com/user-attachments/assets/e3bf0c4e-289d-4a84-a157-0d15255ef7d8)


This seems to happen with text as well. While I was typing on one window the next was constantly scrolling to the top making the experience for the other user very bad.

This should be an easy fix right @aatbip ?

PR #371, opened 2024-10-24
OUT-908 | Delete task for CU and decrement notification badge when the assignee company is deleted

rrojan:
### Changes

- [x] Fix case where deleting company wouldn't decrement client notification

### Testing Criteria

- [x] Tested all webhook methods related to user deletion:

(refreshing to change notification chip count is not needed - it just seems to reflect a bit faster this way)

#### Company assigned
[Screencast from 2024-10-24 14-34-53.webm](https://github.com/user-attachments/assets/1d33d443-2a31-491a-a691-828d46ad1b37)

### Company unassigned
[Screencast from 2024-10-24 14-37-21.webm](https://github.com/user-attachments/assets/219d3d3d-c52d-48a0-8857-907b92964f71)

### Company reassigned
[Screencast from 2024-10-24 14-38-47.webm](https://github.com/user-attachments/assets/76d4dbb5-87d4-4e82-b54d-b65db7f01ff9)

### Company deleted
[Uploading Screencast from 2024-10-24 14-40-01.webm…]()

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -332,6 +332,7 @@ export class TasksService extends BaseService {
       where: { assigneeId, assigneeType, workspaceId: this.user.workspaceId },
     })
     await this.db.label.deleteMany({ where: { label: { in: labels } } })
+    return tasks
   }
 
   async clientUpdateTask(id: string, targetWorkflowStateId?: string | null) {
--- a/src/app/api/webhook/webhook.service.ts
+++ b/src/app/api/webhook/webhook.service.ts
@@ -140,7 +140,6 @@ class WebhookService extends BaseService {
         taskId: { in: taskIds },
       },
     })
-    console.log('c', clientTaskNotifications)
 
     const clientNotificationIds = clientTaskNotifications.map((notification) => notification.notificationId)
     if (clientNotificationIds.length) {
--- a/src/app/api/webhook/webhook.service.ts
+++ b/src/app/api/webhook/webhook.service.ts
@@ -121,16 +121,32 @@ class WebhookService extends BaseService {
   }
 
   async handleUserDeleted(assigneeId: string, assigneeType: AssigneeType) {
-    if (assigneeType === AssigneeType.company) return
-
     const tasksService = new TasksService(this.user)
     // Delete corresponding tasks
     console.info(`Deleting all tasks for ${assigneeType} ${assigneeId}`)
-    await tasksService.deleteAllAssigneeTasks(assigneeId, assigneeType)
+    const tasks = await tasksService.deleteAllAssigneeTasks(assigneeId, assigneeType)
 
-    // So how do we delete corresponding notifications, you ask? Good question.
-    // Copilot calls multiple `client.updated` when unassigning clients companies after `company.deleted` is called.
-    // So handling deleting company task notifications should be done in `client.updated` company unassignment logic instead
+    // Now delete any and all associated notifications triggered on behalf of company tasks for clients.
+    // i.e. decrement client task count in CU portal
+    if (assigneeType === AssigneeType.company) {
+      const deletedTaskIds = tasks.map((task) => task.id)
+      await this.handleDeletingTaskNotifications(deletedTaskIds)
+    }
+  }
+
+  private async handleDeletingTaskNotifications(taskIds: string[]) {
+    const clientTaskNotifications = await this.db.clientNotification.findMany({
+      where: {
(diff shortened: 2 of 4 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -332,6 +332,7 @@ export class TasksService extends BaseService {
       where: { assigneeId, assigneeType, workspaceId: this.user.workspaceId },
     })
     await this.db.label.deleteMany({ where: { label: { in: labels } } })
+    return tasks
```
makeitraina, line 335: just curious, what is `db.label`? why do we need to delete that?

src/app/api/webhook/webhook.service.ts
```diff
@@ -121,16 +121,31 @@ class WebhookService extends BaseService {
...
+    // Now delete any and all associated notifications triggered on behalf of company tasks for clients.
+    // i.e. decrement client task count in CU portal
+    if (assigneeType === AssigneeType.company) {
+      const deletedTaskIds = tasks.map((task) => task.id)
+      await this.handleDeletingTaskNotifications(deletedTaskIds)
+    }
+  }
+
+  private async handleDeletingTaskNotifications(taskIds: string[]) {
+    const clientTaskNotifications = await this.db.clientNotification.findMany({
```
makeitraina, line 138: is `db.clientNotification` created to seprately store client notifications and not IU notifications?

makeitraina (approved):
looks good, just some clarification questions

src/app/api/webhook/webhook.service.ts
```diff
@@ -121,16 +121,31 @@ class WebhookService extends BaseService {
   }
 
   async handleUserDeleted(assigneeId: string, assigneeType: AssigneeType) {
-    if (assigneeType === AssigneeType.company) return
-
     const tasksService = new TasksService(this.user)
     // Delete corresponding tasks
     console.info(`Deleting all tasks for ${assigneeType} ${assigneeId}`)
-    await tasksService.deleteAllAssigneeTasks(assigneeId, assigneeType)
+    const tasks = await tasksService.deleteAllAssigneeTasks(assigneeId, assigneeType)
```
sazanrjb, line 127: if `tasks` is empty, lets return here so that we don't do extra db call

src/app/api/webhook/webhook.service.ts
```diff
@@ -121,16 +121,31 @@ class WebhookService extends BaseService {
...
+  private async handleDeletingTaskNotifications(taskIds: string[]) {
+    const clientTaskNotifications = await this.db.clientNotification.findMany({
+      where: {
+        taskId: { in: taskIds },
+      },
+    })
+
+    const clientNotificationIds = clientTaskNotifications.map((notification) => notification.notificationId)
+    if (clientNotificationIds.length) {
+      await this.copilot.bulkMarkNotificationsAsRead(clientNotificationIds)
```
sazanrjb, line 146: can we run these operations in async?

```
Promise.all([
    this.copilot.bulkMarkNotificationsAsRead(clientNotificationIds),
    this.db.clientNotification.deleteMany({ where: { notificationId: { in: clientNotificationIds } } })
])
```

src/app/api/tasks/tasks.service.ts
sazanrjb, line 335: lets add return type `Promise<Task[]>`

src/app/api/webhook/webhook.service.ts
rrojan: Yes

src/app/api/webhook/webhook.service.ts
rrojan: > or do we need to delete only after marking them as read?

Yes, not doing this could cause inconsistencies, and I don't think doing them in parallel for each notification (markAsRead -> delete from db) would be very performant, although that's the best way to approach this.

Same for not doing those async operations in parallel. Doing markAsRead again to a read notification won't cause issues, but deleting them before they are marked as read in a different thread definitely will.

With our current promise-retry mechanism this should be very reliable.

sazanrjb (approved):

src/app/api/tasks/tasks.service.ts
rrojan: As for deleting it, I'm not very familiar with the implementation, and the implementation for `label` seems to be a bit awkward... I think it was Ananta or Arpan who implemented this the last time around - and I didn't want to remove this in case deleting the corresponding label was necessary.

PR #373, opened 2024-10-24
OUT-956 | Backfill activity log for tasks to have created event

rrojan:
### Changes

- [x] Add cmd script to backfill activity logs for previous tasks

### Testing Criteria

- [x] Screenshot
(fixed count starting from 0 after screenshot)
image

- [x] Validated counts by running a query for checking task count with activity log count


src/cmd/fill-activity-logs/index.ts
```diff
@@ -0,0 +1,44 @@
...
+      assigneeType: task.assigneeType,
+    }
+    data.push({
+      taskId: task.id,
+      workspaceId: task.workspaceId,
+      type: ActivityType.TASK_CREATED,
+      userId: task.createdById,
+      userRole: AssigneeType.internalUser,
+      details,
+      createdAt: task.createdAt,
```
sazanrjb, line 37: dont we have updatedAt field?

rrojan: This is not required or used in any way in TASK_CREATED activityLog. So just let prisma take over. It will also help us spot the manually added activitylogs later in case something goes wrong, since all these records will have the same timestamp.

sazanrjb (approved):

PR #375, opened 2024-10-24
OUT-961 Disable image functionality inside comment input box

aatbip:
### Changes

- [x] Disables DnD and CP for comment input


src/components/inputs/CommentInput.tsx
```diff
@@ -88,6 +88,7 @@ export const CommentInput = ({ createComment, task_id }: Prop) => {
           placeholder="Leave a comment..."
           suggestions={assigneeSuggestions}
           editorClass="tapwrite-comment-input"
+          disablePasteAndDnd
```
makeitraina, line 91: something for future, I would have probably named this prop disableImage. Under the hood the library knows that it needs to disable. In the future we might have another create method and in that case this would be less accurate. Also this prop is probably what should control the floating menu item being disabled.

makeitraina (approved):
Something to consider for future, approving this for now

makeitraina:
Do we need to have this commit that rojan just added: https://github.com/rrojan/tapwrite/commit/703f824c9f8abcb1ba54c750ea01136786280748? Its not in the version we used right now right?

aatbip:
> Do we need to have this commit that rojan just added: [rrojan/tapwrite@703f824](https://github.com/rrojan/tapwrite/commit/703f824c9f8abcb1ba54c750ea01136786280748)? Its not in the version we used right now right?

I think the changes are same so we are good here.

PR #376, opened 2024-10-25
OUT 966 Comments get cut off when sent via enter key

aatbip:
### Changes

- [x] Fixes issue with comment cut off

_Note: This doesn't bring back the issue reported as "Enter command creates a new line" in OUT-952._


makeitraina (approved):
just confirm that you've tested the changes.

Also, im just curious for my understanding: Why are we removing all this? In Rojan's initial fix for enter new-line I saw he kept this and thought it was intetional.

Otherwise this is good to go. Can merge and remember to promote to prod once confirmed.

makeitraina:
i think the empty check might be not working though? maybe we should check in a different way anyway

src/components/inputs/CommentInput.tsx
```diff
@@ -26,15 +26,9 @@ export const CommentInput = ({ createComment, task_id }: Prop) => {
...
-    const detailsText = detail.replaceAll('<p>', '').replaceAll('</p>', '').replaceAll(' ', '').replaceAll('<br>', '')
-    const isEmpty = detailsText === ''
-    if (isEmpty) return
-
-    // Tiptap / Hardbreak appends this at the end if you submit a comment using enter
-    const enterBreakEl = '<p></p>'
+  const handleSubmit = () => {
     const commentPayload: CreateComment = {
-      content: opts?.isEnter ? detail.slice(0, detail.length - enterBreakEl.length) : detail,
+      content: detail,
```
rrojan, line 31: @aatbip Tiptap sends an extra `<p></p>` at the end when submitting a comment using enter key. Where are we handling this?

rrojan: I don't see a version bump for `tippytappy` either... where is this being handled instead?

rrojan (commented):
I see that enter issue has been fixed from Tiptap side... I did a temp fix for this anyways, it was always meant to be fixed in the related Tiptap extension eventually @makeitraina .

rrojan (changes requested):
Now we have issue with optimistic update where empty comments can be sent, but they are removed after 2-3 seconds after response is received.

We still need to check the empty text condition in `handleSubmit` itself like I did @aatbip . Maybe parse the markup in a better way?

[Screencast from 2024-10-25 12-27-36.webm](https://github.com/user-attachments/assets/be18790f-5e40-43de-8411-258e47a01363)

aatbip:
> just confirm that you've tested the changes.
> 
> Also, im just curious for my understanding: Why are we removing all this? In Rojan's initial fix for enter new-line I saw he kept this and thought it was intetional.
> 
> Otherwise this is good to go. Can merge and remember to promote to prod once confirmed.

@makeitraina The enter command for line break has been disabled from the Tapwrite itself which seems to be working fine now.

src/components/inputs/CommentInput.tsx
aatbip: This is working because we made some changes in the `HardBreak` extension's configuration in `tippytappy` yesterday. Regarding extra `p` tags at the end, seems like this has also been resolved and not happening anymore. Since the `enter` command line break is already disabled, `p` tag won't come at all. Please see the video below - 
[Screencast from 25-10-24 01:53:45 PM +0545.webm](https://github.com/user-attachments/assets/40fd125f-698f-4b11-ba5d-1106ee994f1b)

src/components/inputs/CommentInput.tsx
```diff
@@ -27,13 +27,15 @@ export const CommentInput = ({ createComment, task_id }: Prop) => {
...
-      taskId: task_id,
-      mentions: getMentionsList(detail),
+    if (detail) {
+      const commentPayload: CreateComment = {
+        content: detail,
+        taskId: task_id,
+        mentions: getMentionsList(detail),
+      }
+      createComment(commentPayload)
+      setDetail('')
```
rrojan, line 37: Please check for all cases surrounding this feature as well.
- Can post comment with spaces only
- Can post comment with newline only
image

rrojan (changes requested):

src/components/inputs/CommentInput.tsx
aatbip: Fixed in the commit below @rrojan

rrojan (approved):
LGTM :trophy:

PR #382, opened 2024-10-28
OUT-949 | The initiator of the action shouldn't get a notification (if they are also the recipient)

rrojan:
### Changes

- [x] Don't send IU taskCreated notification if they are moving task assigned to themselves to an incomplete state

### Testing Criteria

- [x] Tested with preview notifications
Screencast:
[Screencast from 2024-10-28 20-28-17.webm](https://github.com/user-attachments/assets/aed036b9-04f9-4b1a-bc3a-62543989e58a)

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -578,12 +578,16 @@ export class TasksService extends BaseService {
 
     // Case 4
     // --- Handle task moved from completed to incomplete IU logic
+    const isAssignedIU =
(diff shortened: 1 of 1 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -578,12 +578,16 @@ export class TasksService extends BaseService {
 
     // Case 4
     // --- Handle task moved from completed to incomplete IU logic
+    const isAssignedIU =
```
sazanrjb, line 581: since we are checking if the assignee is currentUser, `isSelfAssignedIU` makes sense?

sazanrjb (approved):

PR #383, opened 2024-10-29
OUT 935 Vertically align Activity title with divider line and remove padding

aatbip:
### Changes

- [x] Removes left padding from the activity log divs

### Testing Criteria

- Please see the activity log section and notice the left padding of the header div container "Activity" text and the entire log aligns with the top border


rrojan (approved):
Looks good on all screen sizes.
LGTM :trophy:

PR #384, opened 2024-10-29
OUT 903 Safely handle when a user opens a notification that is linked to a deleted task

aatbip:
### Changes

- [x] Adds a redirection page for the deleted tasks


src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -83,7 +84,9 @@ export default async function TaskDetailPage({
 
   console.info(`app/detail/${task_id}/${user_type}/page.tsx | Serving user ${token} with payload`, tokenPayload)
 
-  redirectIfResourceNotFound(searchParams, task, !!tokenPayload.internalUserId)
+  if (!!!task) {
```
Comment (rrojan) [inline src/app/detail/[task_id]/[user_type]/page.tsx:87]: Wouldn't a single `!` be enough here?

src/app/detail/[task_id]/[user_type]/page.tsx
```diff
@@ -83,7 +84,9 @@ export default async function TaskDetailPage({
...
+    return <DeletedTaskRedirectPage userType={tokenPayload.internalUserId ? UserRole.IU : UserRole.Client} />
+  }
```
Comment (rrojan) [inline src/app/detail/[task_id]/[user_type]/page.tsx:89]: Why aren't we passing token as a prop directly here and instead taking a round trip to the task board selector? I think there were 2-3 issues Neil was looking over that was related to this.

Also it's just much more efficient since we are calling the DeletedTaskRedirect component from a server component directly

src/components/layouts/DeletedTaskRedirectPage.tsx
```diff
@@ -0,0 +1,62 @@
+'use client'
+
+import { UserRole } from '@/app/api/core/types/user'
```
rrojan, line 3: This is a simple static UI so there is no use to make the whole thing a client component imo.

After we remove the part to get token from taskboard selector, we are just making this a client component for the button click -> board view routing. We can simply use `Link` for that, or just make the button a client component. What do you think?

src/components/layouts/DeletedTaskRedirectPage.tsx
```diff
@@ -0,0 +1,62 @@
...
+              <Typography variant="2xl" lineHeight={'32px'}>
+                No task to show
+              </Typography>
+              <Typography variant="bodyLg" sx={{ color: (theme) => theme.color.gray[500] }}>
+                This task has been deleted. You can view your other tasks on the Tasks homepage.
+              </Typography>
+            </Stack>
+            <Box>
+              <PrimaryBtn
+                buttonText="New task"
```
rrojan, line 51: image

rrojan (changes requested):

src/app/detail/[task_id]/[user_type]/page.tsx
Comment (aatbip) [reply src/app/detail/[task_id]/[user_type]/page.tsx]: Fixed in 1b1bdd22b0f2b8e4847a14ea55319b86945f2fc0152cc13b43d57b1f3978d6c5cf00373fd4c5e70689a39d18aa04914f69695a181b726fbbecabbbd79a3fbd8da7c940c35c321774deddd92ae6e0e7785b040c81c307cd4c50a7a697ec144ac6b4b6eb2257e4a8853fff900138a8d279 commit

src/app/detail/[task_id]/[user_type]/page.tsx
Comment (aatbip) [reply src/app/detail/[task_id]/[user_type]/page.tsx]: Fixed in the commit 26a2a7b244909e7c7467cbea75c4ba08250d0dcf

src/components/layouts/DeletedTaskRedirectPage.tsx
aatbip: Fixed in the commit ce0438e80df34cd4076ef0bf16f21af2263a70ce

src/components/layouts/DeletedTaskRedirectPage.tsx
aatbip: Fixed in the commit ce0438e80df34cd4076ef0bf16f21af2263a70ce

rrojan (approved):
LGTM :trophy:

PR #387, opened 2024-10-30
OUT-884, OUT-886, OUT-982 Attachments card design, Attachments Extension Implementation

aatbip:
### Tasks

- [Attachments Extension Implementation](https://linear.app/copilotplatforms/issue/OUT-886/attachments-extension-implementation)
- [Attachments card design](https://linear.app/copilotplatforms/issue/OUT-884/attachments-card-design)
- [Loading state for file uploads](https://linear.app/copilotplatforms/issue/OUT-982/loading-state-for-file-uploads)

### Changes

- [x] Adds inline attachment layout
- [x] Worked on attachment designs
- [x] Integrated tapwrite's attachments on tasks app
- [x] handled download function of attachments, selection designs   

### Testing criteria
- [LOOM](https://www.loom.com/share/d871e27b992c4049be88fe2f6e9b334c)

### Notes
- tasks left : loading state for attachments, default icons for unrecognized filetypes and some minor polishing.


src/app/detail/ui/TaskEditor.tsx
```diff
@@ -173,6 +174,7 @@ export const TaskEditor = ({
             return t
           }}
           deleteEditorAttachments={(url) => deleteEditorAttachmentsHandler(url, token ?? '', task_id)}
+          attachmentLayout={(props) => <AttachmentLayout {...props} />}
```
makeitraina, line 177: any reason we spread the AttachmentLayout and dont directly do `attachmentLayout={AttachmentLayout}`?

src/utils/signedUrlReplacer.ts
```diff
@@ -2,12 +2,13 @@ import { Prisma, PrismaClient, StateType, WorkflowState } from '@prisma/client'
 import { DefaultArgs } from '@prisma/client/runtime/library'
 
 export async function replaceImageSrc(htmlString: string, getSignedUrl: (filePath: string) => Promise<string | undefined>) {
-  const imgTagRegex = /<img\s+[^>]*src="([^"]+)"[^>]*>/g //expression used to match all img tags in provided HTML string.
+  const imgTagRegex = /<img\s+[^>]*src="([^"]+)"[^>]*>/g
+  const attachmentTagRegex = /data-type="attachment"\s+[^>]*src="([^"]+)"[^>]*>/g //expression used to match all img tags in provided HTML string.
```
makeitraina, line 6: this comment is incorrect, this regex is different. Also where does data-type get set? We should update this comment to explain _why_ this regex works

makeitraina (commented):
very minor comments on my end that are just some suggestions. I'll let your team do the main review and testing.

src/utils/signedUrlReplacer.ts
priosshrsth: Does this regex need to match any attachment or just the image? In which case, should the src matter? Is it the case that we only have image attachment? Can we include mime type in our attachment extension to make things easier for us maybe?

priosshrsth (changes requested):
@aatbip I have mentioned this in the past, but we shouldn't be seeing all these warnings about missing dependencies in useEffect and useCallback. Please suppress them. In the future any developer would try to include the missing dependency unknowingly causing memory leak because no one would have context, whether they should be ignored or was just a mistake.

src/utils/signedUrlReplacer.ts
arpandhakal: This change needs to be reverted because I realized we do not need fresh signed urls for attachments. Unlike images, attachments needs to be fetched on demand only(download) and not every time. So, doing this was unnecessary.

src/app/detail/ui/TaskEditor.tsx
arpandhakal: This is done because we want the props to be dynamically passed from parent component(Tapwrite) to child component (Task App's task editor). We are to get props from Tapwrite like fileName, src, fileSize, fileType because file handling is done through Tapwrite.

makeitraina: What do you mean by dynamically passed?

If you do `attachmentLayout={AttachmentLayout}` the props are passed down are they not?

arpandhakal: yup. it should work, sorry I misunderstood. Applying the changes.

PR #393, opened 2024-11-07
OUT-972, OUT-984 Set up proper Supabase Storage folder structure in SupabaseService

arpandhakal:
### Tasks

- [Set up proper Supabase Storage folder structure in SupabaseService](https://linear.app/copilotplatforms/issue/OUT-972/set-up-proper-supabase-storage-folder-structure-in-supabaseservice)
- [Code cleaning for m4](https://linear.app/copilotplatforms/issue/OUT-984/code-cleaning-for-m4)

### Changes

- [x] refactored code : removed attachments code from before. (some cleaning still remaining here) .
- [x] supabase folder structure : created a clean supabase folder structure following this pattern in supabase bucket for attachments and inline images : bucketName/workspaceId/taskId/file. Before, every files were stored under bucket
- [x] scrap table name refactored and utility functions refactored which were shared by inline images and attachments. 

### Testing Criteria

Screenshot from 2024-11-07 18-00-49

### DIFF
--- a/prisma/migrations/20241107081546_scrap_attachments/migration.sql
+++ b/prisma/migrations/20241107081546_scrap_attachments/migration.sql
@@ -1,0 +1,36 @@
+/*
+  Warnings:
+
+  - You are about to drop the `ScrapImages` table. If the table is not empty, all the data it contains will be lost.
+
+*/
+-- DropForeignKey
+ALTER TABLE "ScrapImages" DROP CONSTRAINT "ScrapImages_taskId_fkey";
+
+-- DropTable
+DROP TABLE "ScrapImages";
+
+-- CreateTable
+CREATE TABLE "ScrapAttachments" (
+    "id" UUID NOT NULL DEFAULT gen_random_uuid(),
+    "filePath" TEXT NOT NULL,
+    "taskId" UUID,
+    "createdAt" TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMPTZ NOT NULL,
+    "deletedAt" TIMESTAMPTZ,
+    "workspaceId" VARCHAR(32) NOT NULL,
+
+    CONSTRAINT "ScrapAttachments_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "ScrapAttachments_filePath_key" ON "ScrapAttachments"("filePath");
+
+-- CreateIndex
+CREATE INDEX "ScrapAttachments_createdAt_idx" ON "ScrapAttachments"("createdAt");
+
+-- CreateIndex
+CREATE INDEX "ScrapAttachments_filePath_idx" ON "ScrapAttachments"("filePath");
+
+-- AddForeignKey
+ALTER TABLE "ScrapAttachments" ADD CONSTRAINT "ScrapAttachments_taskId_fkey" FOREIGN KEY ("taskId") REFERENCES "Tasks"("id") ON DELETE CASCADE ON UPDATE CASCADE;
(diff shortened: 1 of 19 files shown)


sazanrjb:
In one migration you have added workspaceId in ScrapImages table, and in another migration you have deleted ScrapImages table. I don't get this change. Also it will throw error in migrate command if there are data in the table as you have added not null.

Also, IMO we could have just used ScrapImages table for now and after the release we could have created the new table, migrated the old data, and then only deleted the ScrapImages table.

I don't see the code to migrate the data from ScrapImages table to ScrapAttachments table, is it missing?
And what if there are data in ScrapImages table, and after the deployment, it deletes the table. There will be stale data, no?

The correct steps here would be:

Deployment 1
- Create new table ScrapAttachments
- Write a migration to migrate from ScrapImages table to ScrapAttachments table
- Continue writing to both tables
- Take the data from ScrapAttachments in worker

Deployment 2
- Create migration to delete ScrapImages

So I think we can just complete the scope of this ticket with structuring folder structure, and we can refactor the table name in another PR. I think both of these tasks are not blocked for the release.

src/utils/signedUrlReplacer.ts
```diff
@@ -36,6 +33,16 @@ export function getFilePathFromUrl(url: string) {
   }
 }
 
+export function getAbsoluteFileName(filePath: string): string | null {
+  try {
+    const parts = filePath.split('/')
+    return parts.pop() || null
```
rrojan, line 39: As long as the filePath is a string `parts.pop()` should always return a value right?
The only case when `parts.pop()` will be falsy is when the parts array is empty, or the last element in array is `undefined`, `null`, `0`, `false`, `""` etc. Is there a case this can occur?

src/utils/inlineAttachment.ts
```diff
@@ -9,14 +9,20 @@ import { getFilePathFromUrl } from '@/utils/signedUrlReplacer'
...
-export const uploadImageHandler = async (file: File, token: string, task_id: string | null): Promise<string | undefined> => {
+export const uploadAttachmentHandler = async (
+  file: File,
+  token: string,
+  workspaceId: string,
+  task_id: string | null,
+): Promise<string | undefined> => {
   const supabaseActions = new SupabaseActions()
-
+  const filePath = `/${workspaceId}${task_id ? `/${task_id}` : ''}`
```
rrojan, line 19: It might be better to organize scrap attachments into a separate folder like `temp/` instead of having them pile up at the root of the workspaceId folder. 

This can also help us catch edge cases.

src/app/api/attachments/attachments.service.ts
```diff
@@ -8,6 +8,7 @@ import { supabaseBucket } from '@/config'
 import APIError from '@api/core/exceptions/api'
 import httpStatus from 'http-status'
 import { SupabaseService } from '@api/core/services/supabase.service'
+import { signedUrlTtl } from '@/types/constants'
```
rrojan, line 11: Has this been added to Vercel? We should also be using a fallback TTL for this

rrojan (changes requested):

prisma/migrations/20241107081546_scrap_attachments/migration.sql
```diff
@@ -0,0 +1,36 @@
...
+    "filePath" TEXT NOT NULL,
+    "taskId" UUID,
+    "createdAt" TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMPTZ NOT NULL,
+    "deletedAt" TIMESTAMPTZ,
+    "workspaceId" VARCHAR(32) NOT NULL,
+
+    CONSTRAINT "ScrapAttachments_pkey" PRIMARY KEY ("id")
+);
+
```
rrojan, line 25:
```suggestion
ALTER TABLE ScrapImages
RENAME TO ScrapAttachments;
```

rrojan (changes requested):

PR #402, opened 2024-11-11
OUT-1003, OUT-1004 Add 8px of padding between attachment cards for proper spacing, Update shadow on attachment card

aatbip:
### Changes

- [x] Fixes design issues in the `AttachmentLayout`

### Testing Criteria

- [x] Go to the details page
- [x] Press '/' to open the menu and click on the 'Uploads'
- [x] Upload a file then notice the file card in the editor. Upload another file after a enter press when the first file gets uploaded
- [x] Notice the margin between the cards is 8px
- [x] Notice the components doesn't flicker when hovered
- [x] Notice the overall CSS of the layout has been fixed

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "react-dom": "^18",
     "react-redux": "^9.1.0",
     "swr": "^2.2.5",
-    "tapwrite": "1.1.67",
+    "tapwrite": "1.1.70",
     "zod": "^3.23.8"
   },
   "devDependencies": {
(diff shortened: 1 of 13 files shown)


rrojan (changes requested):
There is no spacing between text and attachments. In the design, spacing should be 8px.

image

The way `margin-bottom` of 8px has been applied will not add any gap on top of the attachment if the element on top is not an attachment card

rrojan (changes requested):
The shadow seems to "cut out" on the sides, probably due to inadequate spacing for the box-shadow to blur out... it doesn't look smooth.

Editor:
image
image

Figma:
image

PR #405, opened 2024-11-12
OUT 991 When I have the formatting menu open and hit 'Escape' I want to close the formatting menu instead of being taken back to the all tasks views

aatbip:
### Changes

- [x] Adds a check if formatting menu is active in the dom before closing the modal

### Testing Criteria

[Screencast from 12-11-24 12:47:18 PM +0545.webm](https://github.com/user-attachments/assets/17ab9836-c74c-4283-9fde-8eb03f116684)

### DIFF
--- a/src/app/ui/Modal_NewTaskForm.tsx
+++ b/src/app/ui/Modal_NewTaskForm.tsx
@@ -29,8 +29,8 @@ export const ModalNewTaskForm = ({
   const { title, description, workflowStateId, assigneeId, assigneeType, attachments, dueDate, showModal } =
     useSelector(selectCreateTask)
 
-  const handleModalClose = async ({ triggeredFrom }: { triggeredFrom: 'MOUSE' | 'KEYBOARD' }) => {
-    if (triggeredFrom === 'KEYBOARD' && document.querySelector('.tippy-box')) {
+  const handleModalClose = async (isKeyboard: boolean = false) => {
+    if (isKeyboard && document.querySelector('.tippy-box')) {
       return
     }
     store.dispatch(setShowModal())
@@ -43,7 +43,7 @@ export const ModalNewTaskForm = ({
   return (
     <Modal
       open={showModal}
-      onClose={() => handleModalClose({ triggeredFrom: 'KEYBOARD' })}
+      onClose={() => handleModalClose(true)}
       aria-labelledby="create-task-modal"
       aria-describedby="add-new-task"
     >
@@ -73,7 +73,7 @@ export const ModalNewTaskForm = ({
             await handleCreateMultipleAttachments(toUploadAttachments)
           }
         }}
-        handleClose={() => handleModalClose({ triggeredFrom: 'MOUSE' })}
+        handleClose={() => handleModalClose()}
         getSignedUrlUpload={getSignedUrlUpload}
       />
     </Modal>
--- a/src/app/ui/Modal_NewTaskForm.tsx
+++ b/src/app/ui/Modal_NewTaskForm.tsx
@@ -29,16 +29,24 @@ export const ModalNewTaskForm = ({
   const { title, description, workflowStateId, assigneeId, assigneeType, attachments, dueDate, showModal } =
     useSelector(selectCreateTask)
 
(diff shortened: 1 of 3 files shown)


rrojan (changes requested):
> I want to close the formatting menu instead of being taken back to the all tasks views

The main issue here still persists.

This is expected to work in both task create modal, as well as task description.

[Screencast from 2024-11-12 13-00-03.webm](https://github.com/user-attachments/assets/dd5f5d8e-17ef-4d75-85de-138a447d1138)

src/app/ui/Modal_NewTaskForm.tsx
```diff
@@ -29,16 +29,24 @@ export const ModalNewTaskForm = ({
...
+  const handleModalClose = async ({ triggeredFrom }: { triggeredFrom: 'MOUSE' | 'KEYBOARD' }) => {
+    const hasSlashCommandMenu = document.querySelector('.tippy-box') !== null
+    if (!(triggeredFrom === 'KEYBOARD' && hasSlashCommandMenu)) {
+      store.dispatch(setShowModal())
+      store.dispatch(clearCreateTaskFields({ isFilterOn: !!filterOptions[FilterOptions.ASSIGNEE] }))
+      store.dispatch(setActiveWorkflowStateId(null))
+    }
     // NOTE: Reimplement in M3
     // await bulkRemoveAttachments(attachments)
   }
```
priosshrsth, line 41:
```suggestion
    if (triggeredFrom === 'KEYBOARD' && document.querySelector('.tippy-box')) {
      return;
    }
    
    store.dispatch(setShowModal())
      store.dispatch(clearCreateTaskFields({ isFilterOn: !!filterOptions[FilterOptions.ASSIGNEE] }))
      store.dispatch(setActiveWorkflowStateId(null))
    /* TODO: Reimplement in M3
    await bulkRemoveAttachments(attachments)
    */
  }
```

src/app/ui/NewTaskForm.tsx
```diff
@@ -107,7 +107,7 @@ export const NewTaskForm = ({ handleCreate, handleClose, getSignedUrlUpload }: N
   useEffect(() => {
     function handleEscPress(e: KeyboardEvent) {
       if (e.key === 'Escape') {
-        handleClose()
+        handleClose({ triggeredFrom: 'MOUSE' })
```
rrojan, line 110: Instead of adding `triggeredFrom: 'Mouse'` to every button and icon, we can make it the default behavior

rrojan (changes requested):

src/app/ui/NewTaskForm.tsx
aatbip: Using a default value for `triggeredFrom` can lead to implicit behavior, making it unclear what triggered the modal close if `triggeredFrom` is omitted by mistake. Explicitly specifying 'MOUSE' or 'KEYBOARD' keeps the intent clear and improves readability. This approach avoids potential debugging issues due to unintended default behavior.

priosshrsth: @aatbip I think we only care when trigged  by keyboard, isn't it? In that case, we could make it simpler by making this optional parameter. Which should've kept the changes minimal. But now that you have made the changes, I do agree with you here that it keeps intention clear. But the way it is done, looks bothersome to provide the triggeredFrom params everywhere.

aatbip: Changes added in the commit below @rrojan @priosshrsth

rrojan (changes requested):
@aatbip When i add some changes, then a command like `/up` and press escape to close the formatting menu the changes are lost...

[Screencast from 2024-11-12 19-25-39.webm](https://github.com/user-attachments/assets/d076a8e6-58d9-4f9f-a187-02a649d5c935)

rrojan:
@aatbip It seems like Realtime is having issues on Supabase again. Is it possible this is the cause? After this is resolved please look through this once again.

image

rrojan (approved):
LGTM :trophy:

PR #426, opened 2024-11-21
OUT-1050 | Add and retain selected design state to List/Board view buttons in Display drawer

arpandhakal:
### Changes

- [x] added background for list and board view icon containers according to their selected state.

### Testing Criteria

image


rrojan (approved):
LGTM :trophy:

PR #427, opened 2024-11-21
OUT-1053 | Display drawer floats away from Display button when window is resized

arpandhakal:
### Changes

- [x] fixed Display drawer floating away on resize complete. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/3204354cc27f468ba32a487528e593ab)

### Notes

- There is still some inconsistency in position of the display drawer. This is likely due to 2 display drawer being rendered in filter bar. Also, the display drawer fits into the desired position after resizing is complete. ref : https://mui.com/material-ui/react-menu/

### DIFF
--- a/src/app/_fetchers/TaskDataFetcher.tsx
+++ b/src/app/_fetchers/TaskDataFetcher.tsx
@@ -6,7 +6,7 @@ import { useCallback, useEffect } from 'react'
 import { useSelector } from 'react-redux'
 import useSWR from 'swr'
 
-export const TaskDataFetcher = ({ token }: { token: string }) => {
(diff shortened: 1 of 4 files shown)


rrojan (changes requested):
The display drawer closes when the mobile filterbar renders.

Imo the best way to do this is to make the "open" state shared and make the `anchorEl` just update new changes to it. This way the selector won't randomly close when resizing to a mobile view.

This will also be beneficial to other tasks that depend on having access to the Selector open state outside of the filterbar (e.g. https://linear.app/copilotplatforms/issue/OUT-1046/when-tasks-board-goes-from-empty-state-to-one-with-tasks-display)

rrojan (approved):
LGTM :trophy:

PR #431, opened 2024-11-25
OUT-1057 | Remove leading zero from dates with single digits

rrojan:
### Changes

- [x] Remove leading zero by changing format string from `DD` to `D`


arpandhakal (approved):
LGTM

PR #435, opened 2024-11-25
OUT-1046, OUT-1085 | When tasks board goes from empty state to one with tasks, display drawer closes unexpectedly

arpandhakal:
### Changes

- [x] removed  ```showArchived``` dependency from  ```hasNoFilter``` and introduced ```globalTasksRepo``` check on it. 
- [x] api calls optimization on taskDataFetcher. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/34d5da049caf44e0af9660ae5a431925)

### DIFF
--- a/src/app/_fetchers/TaskDataFetcher.tsx
+++ b/src/app/_fetchers/TaskDataFetcher.tsx
@@ -2,13 +2,12 @@ import { selectTaskBoard, setIsTasksLoading, setTasks } from '@/redux/features/t
 import store from '@/redux/store'
 import { ArchivedOptionsType } from '@/types/dto/viewSettings.dto'
 import { fetcher } from '@/utils/fetcher'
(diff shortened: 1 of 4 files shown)


rrojan (changes requested):
The experience feels very laggy. The toggle is slow to trigger and feels a unresponsive. We should be triggering to toggle before waiting for POST request to `/api/view-settings`, if this is what is causing the lag.

Also if we have a state storing all tasks regardless of filter, shouldn't toggling a filter off then on again be snappy... currently it looks like this - the tasks take a while to load, the assignees are initially set from skeleton to No Assignee, before finally to a valid assignee

Let's give this PR a fresh look again

[Screencast from 2024-11-26 19-32-38.webm](https://github.com/user-attachments/assets/d4e83ca5-9be6-4d42-8533-89fad6e473ac)

PR #436, opened 2024-11-25
OUT-1027  | Hover state over attachments button should be a rectangle, not a circle

arpandhakal:
### Changes

- [x] tapwrite update : change hover to rectangular from circular for attachments button 

### Testing Criteria

image


rrojan (approved):
LGTM 🏆

PR #440, opened 2024-11-27
OUT 1086 Empty states are not shown for IU and CU

aatbip:
### Changes

- [x] Fixes the empty screen in the case `task` is empty

### Testing Criteria
[Screencast from 27-11-24 10:39:45 AM +0545.webm](https://github.com/user-attachments/assets/219fb461-862b-4d00-8a2f-ca339fe426ef)

### DIFF
--- a/src/app/ui/TaskBoard.tsx
+++ b/src/app/ui/TaskBoard.tsx
@@ -88,7 +88,7 @@ export const TaskBoard = ({ mode }: TaskBoardProps) => {
 
   const isNoTasksWithFilter = tasks && !userHasNoFilter && !filteredTasks.length
 
-  if (tasks && tasks.length === 0 && userHasNoFilter && !isTasksLoading) {
+  if (tasks && tasks.length === 0 && userHasNoFilter) {
     return (
       <>
         <TaskDataFetcher token={token ?? ''} />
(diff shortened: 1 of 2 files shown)


src/app/ui/TaskBoard.tsx
```diff
@@ -88,7 +88,7 @@ export const TaskBoard = ({ mode }: TaskBoardProps) => {
 
   const isNoTasksWithFilter = tasks && !userHasNoFilter && !filteredTasks.length
 
-  if (tasks && tasks.length === 0 && userHasNoFilter && !isTasksLoading) {
+  if (tasks && tasks.length === 0 && userHasNoFilter) {
     return (
```
rrojan, line 92: Is there a reason why we are removing `isTasksLoading` check? This would cause a regression for a previous fix. Originally it was intended to remove sideeffects so we don't show "No tasks found" when loading tasks

Right now "No tasks found" state occurs even though there are valid tasks in the following cases:
- When toggling show archived / show unarchived
- Brief flicker on loading tasks board for the first time (happens for a split second but is noticable sometimes like in 0:21 in screencast)

[Screencast from 2024-11-27 13-27-15.webm](https://github.com/user-attachments/assets/15c31c9d-20f2-406f-8d46-2b6f223d3626)

rrojan (changes requested):

rrojan (approved):
LGTM :trophy: 

Please promote this deployment to production after squash and merge.

PR #442, opened 2024-11-27
OUT 1064 Update hover state for a task

aatbip:
### Changes

- [x] Updates hover state to `gray[100]` 

### Testing Criteria

- Hover on the task card and notice the hover state background color is now `#F8F9FB` i.e. `gray[100]` according to the `theme.ts` design system

image


rrojan (approved):
LGTM :trophy:

PR #444, opened 2024-11-27
OUT-1071 | Tasks should sort by due date, not created date

rrojan:
### Changes

- [x] Change backend query to sort by dueDate, then createdAt as secondary ordering criteria:
```sql
 ORDER by "tasks"."dueDate" asc nulls last, "tasks"."createdAt" DESC
```
- [x] Change frontend tasks sorter to also sort by dueDate (if defined), also made it reusable.
- [x] (OUT-1087) Remove the period from the "Archived." tooltip 
 
### Testing Criteria

- [x] Screencast
[Screencast from 2024-11-27 15-21-43.webm](https://github.com/user-attachments/assets/a12a04ca-b2d9-4678-90ef-2c5d3f48e362)

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -96,7 +96,14 @@ export class TasksService extends BaseService {
         ...filters.where,
         isArchived,
       },
-      orderBy: { createdAt: 'desc' },
+      orderBy: [
+        {
+          dueDate: { sort: 'asc', nulls: 'last' },
+        },
+        {
+          createdAt: 'desc',
+        },
+      ],
       // @ts-ignore TS support for this param is still shakey
       relationLoadStrategy: 'join',
       include: {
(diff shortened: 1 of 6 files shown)


aatbip (changes requested):
The issue at #443 is reflected here and the root of the problem exist here as well. It might not be the good idea to merge this PR without a resolution on the timezone issue. @rrojan

rrojan:
@aatbip I think this issue is unrelated with the other one since it only deals with dueDate, which does not have time data associated with it. Since we don't have a solid start on the other PR, we could merge this first, then deal with it as well.

aatbip:
> @aatbip I think this issue is unrelated with the other one since it only deals with dueDate, which does not have time data associated with it. Since we don't have a solid start on the other PR, we could merge this first, then deal with it as well.

Sounds good :)

aatbip (approved):
lgtm

PR #448, opened 2024-11-28
OUT-1085-re, OUT-1046, OUT-1067 | Issue opening tasks after showing tasks from archived/unarchived filter, globalTaskRepo state which fetches all tasks on load

arpandhakal:
### Changes

- [x] optimized existing code from OUT-1085. 
- [x] New changes includes having only an object dependency store in the global state : ```activeTask``` for components of taskDetails page. 

### Testing Criteria

- [LOOM]()


rrojan (approved):
LGTM :trophy:

PR #449, opened 2024-11-28
OUT-1093 | Tasks reshuffle on mouse hover in board view

rrojan:
### Changes

- [x] Turns out Supabase Realtime is throwing a `UPDATE` event everytime we hover over a task, even though the update payload is exactly the same. I looked into this but there doesn't seem to be anything in any docs / forums related to this.
- [x] Changed `handleRealTimeUpdates` function to treat new update event payload `createdAt` as GMT time (append with 'Z')

### Testing Criteria

- [x] Tasks don't bounce around any more. Screencast:
[Screencast from 2024-11-28 17-14-52.webm](https://github.com/user-attachments/assets/4ade102d-417c-4811-bc14-5095844bcfe4)

### DIFF
--- a/src/hoc/RealTime.tsx
+++ b/src/hoc/RealTime.tsx
@@ -53,7 +53,6 @@ export const RealTime = ({
       }
       //check if the new task in this event belongs to the same workspaceId
       if (canUserAccessTask && showUnarchived) {
-        console.log('Realtime 4')
         store.dispatch(setTasks([...tasks, { ...payload.new, createdAt: new Date(payload.new.createdAt + 'Z') }]))
         store.dispatch(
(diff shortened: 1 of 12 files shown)


src/hoc/RealTime.tsx
```diff
@@ -64,6 +64,13 @@ export const RealTime = ({
     }
     if (payload.eventType === 'UPDATE') {
       const updatedTask = payload.new
+      const isCreatedAtGMT = (updatedTask.createdAt as unknown as string).slice(-1).toLowerCase() === 'z'
+      if (!isCreatedAtGMT) {
+        // DB stores GMT timestamp without 'z', so need to append this manually
+        updatedTask.createdAt = ((updatedTask.createdAt as unknown as string) + 'Z') as unknown as Date
```
sazanrjb, line 70: A zod parse is better than casting to unknown and string..

sazanrjb (approved):

PR #450, opened 2024-12-02
OUT-1097, OUT-1096 | Fix 'Display' button alignment issue

arpandhakal:
### Changes

- [x] fix alignment issue for searchbar and display button in mobile view when the name of filter assignee is too large.
- [x] created a flexwrap to transfer searchbar and display button to another row when they are overflowing. 
- [x] added display icon for display button in mobile view  

### Testing Criteria

- [LOOM](https://www.loom.com/share/c6028bd5c77b4c329a02d3cc985ee484)
image


rrojan (changes requested):
According to OUT-1096, we should only be showing an archive icon for the button in mobile view... what was the change that was done for OUT-1096?

image

rrojan (approved):
LGTM :trophy:

PR #453, opened 2024-12-02
OUT-773 | Show 'No matches found for {search-text}' when there is no search result

arpandhakal:
### Changes

- [x] added noDataOption to be rendered when assignee filter is loaded and there is no data for the searched text

### Testing Criteria

- [LOOM](https://www.loom.com/share/a4991955e780401fb3ee8d599535d8db)


rrojan (approved):
LGTM :trophy: 
Great solution

PR #463, opened 2024-12-05
OUT-1104 | Add list UI for Templates

rrojan:
### Changes

- [x] Add UI for templates selector
- [x] Modify selector component to accept an endoption component 

### Testing Criteria

- [x] Screenshots (my system freezes when trying to record any kind of screencast / loom):
image
image
image


arpandhakal (commented):
image
Notice how the outer div of this search textfield is different from assignee selector and other selector. The border radius seems to be 0 here.

arpandhakal (commented):
image
The outer div's padding  has a padding which is causing the highlighted option not to cover the whole outer div

rrojan:
@arpandhakal I checked this before, and the issues are persisting with our existing selectors as well, this is just not visible because the background for these other selectors for assignee, workflowstate, etc is white due to them being higher up than the templates selector.

Let's deal with this in a dedicated ticket to reduce the scope of this one.

arpandhakal (approved):

PR #464, opened 2024-12-05
OUT-1102 | Add real-time functionality for Manage Page

arpandhakal:
### Changes

- [x] removed revalidation and unnecessary state updates for ```templates```  state from other sources and applied the updates from ```RealTimeTemplate``` component which handles all the changes like Insert, Update, Delete on templates in real time. 
- [x] Creation and integration of ```RealTimeTemplates``` a higher order component which applies relevant changes to the client side on api operation in templates.  

### Testing Criteria

- [LOOM](https://www.loom.com/share/37ee2e69a032437bb9c24ddf052e7d44)

### DIFF
--- a/src/app/manage-templates/page.tsx
+++ b/src/app/manage-templates/page.tsx
@@ -45,7 +45,7 @@ async function getAllTemplates(token: string): Promise<ITemplate[]> {
   return templates.data
 }
 
-export async function getTokenPayload(token: string): Promise<Token> {
+async function getTokenPayload(token: string): Promise<Token> {
   const copilotClient = new CopilotAPI(token)
   const payload = TokenSchema.parse(await copilotClient.getTokenPayload())
   return payload as Token
@@ -59,11 +59,11 @@ interface ManageTemplatesPageProps {
 
 export default async function ManageTemplatesPage({ searchParams }: ManageTemplatesPageProps) {
   const { token } = searchParams
-  const tokenPayload = await getTokenPayload(token)
-  const [workflowStates, assignee, templates] = await Promise.all([
+  const [workflowStates, assignee, templates, tokenPayload] = await Promise.all([
     await getAllWorkflowStates(token),
     addTypeToAssignee(await getAssigneeList(token)),
     await getAllTemplates(token),
+    await getTokenPayload(token),
   ])
 
   return (
--- a/src/app/manage-templates/page.tsx
+++ b/src/app/manage-templates/page.tsx
@@ -48,7 +48,7 @@ async function getAllTemplates(token: string): Promise<ITemplate[]> {
 async function getTokenPayload(token: string): Promise<Token> {
   const copilotClient = new CopilotAPI(token)
   const payload = TokenSchema.parse(await copilotClient.getTokenPayload())
-  return payload as Token
+  return payload
 }
 
 interface ManageTemplatesPageProps {
--- a/src/app/manage-templates/page.tsx
+++ b/src/app/manage-templates/page.tsx
@@ -12,6 +12,8 @@ import { ManageTemplateHeader } from './ui/Header'
 import { MAX_FETCH_ASSIGNEE_COUNT } from '@/constants/users'
 import { CopilotAPI } from '@/utils/CopilotAPI'
 import { CreateTemplateRequest, UpdateTemplateRequest } from '@/types/dto/templates.dto'
+import { RealTimeTemplates } from '@/hoc/RealtimeTemplates'
+import { Token, TokenSchema } from '@/types/common'
 
 async function getAllWorkflowStates(token: string): Promise<WorkflowStateResponse[]> {
   const res = await fetch(`${apiUrl}/api/workflow-states?token=${token}`, {
@@ -43,6 +45,12 @@ async function getAllTemplates(token: string): Promise<ITemplate[]> {
   return templates.data
 }
 
+export async function getTokenPayload(token: string): Promise<Token> {
+  const copilotClient = new CopilotAPI(token)
+  const payload = TokenSchema.parse(await copilotClient.getTokenPayload())
+  return payload as Token
+}
+
 interface ManageTemplatesPageProps {
   searchParams: {
     token: string
@@ -51,14 +59,11 @@ interface ManageTemplatesPageProps {
 
 export default async function ManageTemplatesPage({ searchParams }: ManageTemplatesPageProps) {
   const { token } = searchParams
-
-  const copilotClient = new CopilotAPI(token)
-
-  const [workflowStates, assignee, templates, tokenPayload] = await Promise.all([
+  const tokenPayload = await getTokenPayload(token)
+  const [workflowStates, assignee, templates] = await Promise.all([
     await getAllWorkflowStates(token),
     addTypeToAssignee(await getAssigneeList(token)),
     await getAllTemplates(token),
-    copilotClient.getTokenPayload(),
   ])
 
   return (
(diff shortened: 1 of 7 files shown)


src/app/manage-templates/ui/TemplateBoard.tsx
```diff
@@ -35,6 +35,7 @@ export const TemplateBoard = ({
   if (templates === undefined) {
     return null
   }
+  console.log(templates)
```
rrojan, line 38: Remove console.log

src/hoc/RealtimeTemplates.tsx
```diff
@@ -0,0 +1,123 @@
...
+}) => {
+  const { templates } = useSelector(selectCreateTemplate)
+  const { token } = useSelector(selectTaskBoard)
+
+  const userId = tokenPayload?.internalUserId || tokenPayload?.clientId
+  const userRole = tokenPayload?.internalUserId
+    ? AssigneeType.internalUser
+    : tokenPayload?.clientId
+      ? AssigneeType.client
+      : undefined
```
rrojan, line 36: Let's make it so that a client token cannot connect to the Templates realtime channel. Otherwise they may be able to broadcast changes to templates even though the API doesn't allow it.

src/types/interfaces.ts
```diff
@@ -137,6 +137,7 @@ export interface ITemplate {
   assigneeType: string
   workflowStateId: string
   createdBy: string
+  createdAt: Date
```
rrojan, line 140: We need support for updatedAt since templates are sorted by most recently modified.

src/hoc/RealtimeTemplates.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const handleTemplatesRealTimeUpdates = (payload: RealtimePostgresChangesPayload<RealTimeTemplateResponse>) => {
+    if (payload.eventType === 'INSERT') {
+      let canUserAccessTask = payload.new.workspaceId === tokenPayload?.workspaceId
+      if (userRole === AssigneeType.client) {
+        canUserAccessTask = false
+      }
+      if (canUserAccessTask) {
+        templates
+          ? store.dispatch(
+              setTemplates([...templates, { ...payload.new, createdAt: new Date(payload.new.createdAt + 'Z') }]),
```
rrojan, line 51: Woudn't this insert new records at the very bottom of the templates list?

src/hoc/RealtimeTemplates.tsx
```diff
@@ -0,0 +1,123 @@
...
+  if (!userId || !userRole) {
+    console.error(`Failed to authenticate a realtime connection for id ${userId} with role ${userRole}`)
+  }
+
+  const handleTemplatesRealTimeUpdates = (payload: RealtimePostgresChangesPayload<RealTimeTemplateResponse>) => {
+    if (payload.eventType === 'INSERT') {
+      let canUserAccessTask = payload.new.workspaceId === tokenPayload?.workspaceId
+      if (userRole === AssigneeType.client) {
+        canUserAccessTask = false
+      }
```
rrojan, line 47: Clients should not be able to access templates at all.

src/hoc/RealtimeTemplates.tsx
```diff
@@ -0,0 +1,123 @@
...
+            )
+          : store.dispatch(setTemplates([{ ...payload.new, createdAt: new Date(payload.new.createdAt + 'Z') }]))
+      }
+    }
+    if (payload.eventType === 'UPDATE') {
+      const updatedTemplate = payload.new
+      const isCreatedAtGMT = (updatedTemplate.createdAt as unknown as string).slice(-1).toLowerCase() === 'z'
+      if (!isCreatedAtGMT) {
+        // DB stores GMT timestamp without 'z', so need to append this manually
+        updatedTemplate.createdAt = ((updatedTemplate.createdAt as unknown as string) + 'Z') as unknown as Date
```
rrojan, line 61: Add support for updatedAt, since we sort by most recently modified first

src/hoc/RealtimeTemplates.tsx
```diff
@@ -0,0 +1,123 @@
...
+        // This casting is safe
+      }
+      const oldTemplate = templates && templates.find((template) => template.id == updatedTemplate.id)
+      if (payload.new.workspaceId === tokenPayload?.workspaceId) {
+        if (updatedTemplate.deletedAt) {
+          const newTemplateArr = templates && templates.filter((el) => el.id !== updatedTemplate.id)
+          store.dispatch(setTemplates(newTemplateArr))
```
rrojan, line 68: When templates are modified, they must be pushed to the top of the list

src/hoc/RealtimeTemplates.tsx
```diff
@@ -0,0 +1,123 @@
...
+      .channel('realtime templates')
+      .on(
+        'postgres_changes',
+        // Because of the way supabase realtime is architected for postgres_changes, it can only apply one filter at a time.
+        // Ref: https://github.com/supabase/realtime-js/issues/97
+        {
+          event: '*',
+          schema: 'public',
+          table: 'TaskTemplates',
+          filter: userRole === AssigneeType.internalUser ? `workspaceId=eq.${tokenPayload?.workspaceId}` : undefined,
```
rrojan, line 111: Only filter by workspaceId, once clients cannot connect to this channel.

src/app/manage-templates/page.tsx
```diff
@@ -43,6 +45,12 @@ async function getAllTemplates(token: string): Promise<ITemplate[]> {
   return templates.data
 }
 
+async function getTokenPayload(token: string): Promise<Token> {
+  const copilotClient = new CopilotAPI(token)
+  const payload = TokenSchema.parse(await copilotClient.getTokenPayload())
+  return payload as Token
```
rrojan, line 51: Wouldn't TokenSchema's output be a `Token` by default?
image

rrojan (changes requested):

src/hoc/RealtimeTemplates.tsx
rrojan: I think timestamps in templates table are stored differently than tasks table, *with* timezone info.

Could you please confirm this

rrojan: * I confirmed this, this is fixed in a recent migration to keep things consistent with tasks table

src/hoc/RealtimeTemplates.tsx
arpandhakal: I think we are removing the templates in this specific code block. Applied changes to the ```update``` operation.

src/hoc/RealtimeTemplates.tsx
```diff
@@ -28,37 +28,30 @@ export const RealTimeTemplates = ({
...
-    ? AssigneeType.internalUser
-    : tokenPayload?.clientId
-      ? AssigneeType.client
-      : undefined
+  const userId = tokenPayload?.internalUserId
+  const userRole = AssigneeType.internalUser
 
   if (!userId || !userRole) {
-    console.error(`Failed to authenticate a realtime connection for id ${userId} with role ${userRole}`)
+    console.error(`Failed to authenticate a realtime connection for id ${userId} with role ${userRole}`) //client users cannot connect to the templates real time channel
```
rrojan, line 35: Wouldn't this treat all userRoles as `AssigneeType.internalUser`

src/hoc/RealtimeTemplates.tsx
```diff
@@ -28,37 +28,30 @@ export const RealTimeTemplates = ({
...
-        canUserAccessTask = false
-      }
       if (canUserAccessTask) {
         templates
           ? store.dispatch(
-              setTemplates([...templates, { ...payload.new, createdAt: new Date(payload.new.createdAt + 'Z') }]),
+              setTemplates([{ ...payload.new, updatedAt: new Date(payload.new.updatedAt + 'Z') }, ...templates]),
             )
-          : store.dispatch(setTemplates([{ ...payload.new, createdAt: new Date(payload.new.createdAt + 'Z') }]))
+          : store.dispatch(setTemplates([{ ...payload.new, updatedAt: new Date(payload.new.updatedAt + 'Z') }]))
```
rrojan, line 46: Let's move this to a dedicated util for reuse. Something that takes in the timestamp, validates it to not have 'Z' and appends it if necessary.

src/hoc/RealtimeTemplates.tsx
```diff
@@ -28,37 +28,30 @@ export const RealTimeTemplates = ({
...
     }
     if (payload.eventType === 'UPDATE') {
       const updatedTemplate = payload.new
-      const isCreatedAtGMT = (updatedTemplate.createdAt as unknown as string).slice(-1).toLowerCase() === 'z'
-      if (!isCreatedAtGMT) {
+      const isUpdatedAtGMT = (updatedTemplate.updatedAt as unknown as string).slice(-1).toLowerCase() === 'z'
+      if (!isUpdatedAtGMT) {
         // DB stores GMT timestamp without 'z', so need to append this manually
-        updatedTemplate.createdAt = ((updatedTemplate.createdAt as unknown as string) + 'Z') as unknown as Date
+        updatedTemplate.updatedAt = ((updatedTemplate.updatedAt as unknown as string) + 'Z') as unknown as Date
```
rrojan, line 54: Let's move it to a dedicated util since we are doing this for more than one field

src/hoc/RealtimeTemplates.tsx
```diff
@@ -108,7 +101,7 @@ export const RealTimeTemplates = ({
           event: '*',
           schema: 'public',
           table: 'TaskTemplates',
-          filter: userRole === AssigneeType.internalUser ? `workspaceId=eq.${tokenPayload?.workspaceId}` : undefined,
+          filter: `workspaceId=eq.${tokenPayload?.workspaceId}`,
```
rrojan, line 104: I think we should wait for tokenPayload to be available before initializing a realtime channel

In cases when tokenPayload has not been resolved and put in redux, we would be opening a realtime channel with filter `workspaceId=eq.undefined`

src/types/interfaces.ts
```diff
@@ -137,7 +137,7 @@ export interface ITemplate {
   assigneeType: string
   workflowStateId: string
   createdBy: string
-  createdAt: Date
+  updatedAt: Date
```
rrojan, line 140: Let's keep both createdAt and updatedAt.

rrojan (changes requested):

src/hoc/RealtimeTemplates.tsx
arpandhakal: since tokenPayload is coming directly from server components and not the store, I think theres a guarantee of token payload working when if the token payload is valid.

rrojan (approved):
LGTM :trophy:

PR #475, opened 2024-12-10
OUT-1116 | Applying the same template a second time causes part of the templates menu to disappear

rrojan:
### Changes

- [x] Add prop to blur select when an option is clicked

### Testing Criteria

- [x] Screencast

[Screencast from 2024-12-10 15-31-27.webm](https://github.com/user-attachments/assets/c555cec8-5eb4-4245-b94a-c9a5b31c172d)


arpandhakal (approved):
LGTM

PR #489, opened 2024-12-12
OUT-1107 | Fix border radius issues with all Selectors

arpandhakal:
### Changes

- [x] fixed border radius issue on input field of selector component. 

### Testing Criteria
image


rrojan (changes requested):
The original design does not have rounded containers for autocomplete

image

PR #496, opened 2024-12-18
OUT-1163 | Change create task modal behavior

rrojan:
### Changes

- [x] Add support for creating tasks from client preview mode
- [x] Make assignee unchangeable in preview mode with current client / company as default

### Testing Criteria

- [x] Screencast

### DIFF
--- a/src/app/client/page.tsx
+++ b/src/app/client/page.tsx
@@ -1,6 +1,7 @@
 export const fetchCache = 'force-no-store'
 
 import { AssigneeFetcher } from '@/app/_fetchers/AssigneeFetcher'
+import { TemplatesFetcher } from '@/app/_fetchers/TemplatesFetcher'
 import { ValidateNotificationCountFetcher } from '@/app/_fetchers/ValidateNotificationCountFetcher'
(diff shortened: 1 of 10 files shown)


src/app/client/page.tsx
```diff
@@ -76,7 +76,11 @@ export default async function ClientPage({ searchParams }: { searchParams: { tok
         viewSettings={viewSettings}
       >
         <Suspense fallback={null}>
-          <AssigneeFetcher token={token} userType={UserType.CLIENT_USER} />
+          <AssigneeFetcher
+            token={token}
+            userType={UserType.CLIENT_USER}
+            isPreview={!!(tokenPayload.internalUserId && tokenPayload.clientId)}
```
rrojan, line 82: This will not work.

In company preview mode we won't have clientId, but a companyId that isn't "default" instead. This logic will not treat company preview as preview mode

arpandhakal (changes requested):
Assignee are being fetched on details page too. So the changes with assignee fetcher should also be applied in the details page. 
Right now assignee are not loading on Details page : 
image

arpandhakal (approved):
LGTM

PR #507, opened 2024-12-23
OUT-1176 | Switching between views shows blank screen momentarily

arpandhakal:
### Changes

- [x] added a check for tasks rather than filtered tasks to show NoFilteredTasksState. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/78d4f077391c45418e41ee072237827f)
- [x] Also checked for empty tasks state.

### DIFF
--- a/src/app/ui/TaskBoard.tsx
+++ b/src/app/ui/TaskBoard.tsx
@@ -119,7 +119,7 @@ export const TaskBoard = ({ mode }: TaskBoardProps) => {
           await updateViewModeSettings(z.string().parse(token), payload)
         }}
       />
-      {tasks && tasks.length === 0 && <NoFilteredTasksState />}
+      {!filteredTasks.length && <NoFilteredTasksState />}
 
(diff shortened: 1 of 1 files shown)


src/app/ui/TaskBoard.tsx
```diff
@@ -119,7 +119,7 @@ export const TaskBoard = ({ mode }: TaskBoardProps) => {
           await updateViewModeSettings(z.string().parse(token), payload)
         }}
       />
-      {isNoTasksWithFilter && <NoFilteredTasksState />}
+      {tasks && tasks.length === 0 && <NoFilteredTasksState />}
```
rrojan, line 122: When I change a filter like assignee doesn't filteredTasks change instead of tasks? So checking for `tasks.length === 0` would not be optimal.

rrojan (changes requested):

rrojan (approved):

PR #511, opened 2024-12-23
OUT-1179 | Replace 1/1 or X/X on top of each section in board and list view with just the number of tasks

rrojan:
### Changes

- [x] Remove total count from filtered columns & rows


arpandhakal (approved):
LGTM

PR #512, opened 2024-12-26
OUT-1201 | Comment timestamp and Deleted Users in an archived task

arpandhakal:
### Changes

- [x] Added a check to display 4 weeks old if activity time difference is more than 4 weeks and also less than a month. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/0985f643fec044e1aa8df817b5138c1b)

### DIFF
--- a/src/utils/getTimeDifference.ts
+++ b/src/utils/getTimeDifference.ts
@@ -32,12 +32,12 @@ export const getTimeDifference = (createdAt: string): string => {
   }
 
   const diffInWeeks = now.diff(targetTime, 'week')
-  const diffInMonths = now.diff(targetTime, 'month')
 
-  if (diffInWeeks < 4 || (diffInWeeks === 4 && diffInMonths < 1)) {
+  if (diffInWeeks < 4 || (diffInWeeks === 4 && diffInDays < 30)) {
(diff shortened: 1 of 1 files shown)


src/utils/getTimeDifference.ts
```diff
@@ -32,11 +32,12 @@ export const getTimeDifference = (createdAt: string): string => {
   }
 
   const diffInWeeks = now.diff(targetTime, 'week')
-  if (diffInWeeks < 4) {
+  const diffInMonths = now.diff(targetTime, 'month')
+
+  if (diffInWeeks < 4 || (diffInWeeks === 4 && diffInMonths < 1)) {
```
aatbip: I think it is better to mention in the comments about the exact reason for doing this.

aatbip (changes requested):

rrojan (dismissed):
LGTM :trophy: 

There is of course some inconsistency since not all months have exactly 30 days but this should be more than enough for now.

rrojan (approved):
LGTM :trophy: 
In the future let's use proper conventional commit structure (`docs(OUT-1201): added clarification for diffInWeeks logic`) for this.

image

rrojan (approved):

PR #517, opened 2024-12-26
OUT-1185, OUT-1187, OUT-1186 | Add support for Task Assigned Event

arpandhakal:
### Changes

- [x] refactored previously used utilities and flow for task assigned activity log creation.
- [x] created an activity log for task assigned if a task is created with an assignee (default as of now).
- [x] brought back activity log for task assigned event. 
- [x] enabled task assigned activity log component in the UI and made adjustment accordingly with reference to the design in figma.   

### Testing Criteria

- [LOOM](https://www.loom.com/share/76211bd1b3144b06bf6aef30924e101a)

### Notes

- After the PR has been approved please make sure to run : 
```
 DELETE FROM "ActivityLogs"
 WHERE "type" = 'TASK_ASSIGNED';
```
  The schema for Activity logs regarding task assignment has been changed, this is done to make data consistent. Removal of activity logs for task assignment does not do any repercussions because this is a new feature.

### DIFF
--- a/src/app/api/tasks/tasks.logger.ts
+++ b/src/app/api/tasks/tasks.logger.ts
@@ -27,7 +27,7 @@ export class TasksActivityLogger extends BaseService {
   async logNewTask() {
     await this.logTaskCreated()
     if (this.task.assigneeId) {
-      await this.logTaskAssigneeUpdated(this.task)
+      await this.logTaskAssigneeUpdated()
     }
   }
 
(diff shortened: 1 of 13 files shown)


rrojan (changes requested):
Please refactor this code to use the new `TaskActivityLogger` class from `OUT-1203`.

Let's target to merge that PR first, rebase those changes here, then implement it.

src/app/api/tasks/tasks.logger.ts
```diff
@@ -0,0 +1,70 @@
...
+      }),
+    )
+  }
+
+  private async logTaskAssigneeUpdated(prevTask: Task & { workflowState: WorkflowState }) {
+    await this.activityLogger.log(
+      ActivityType.TASK_ASSIGNED,
+      TaskAssignedSchema.parse({
+        assigneeId: this.task.assigneeId,
+      }),
```
rrojan, line 58: On second thought, could we do `oldValue` / `newValue` for this, where old value can be undefined / null?

This way we automatically support reassignment in this same method.

We should log "A assigned task to B" if oldValue is undefined / null and "A assigned task from B to C" if oldValue contains a valid assigneeId

rrojan (changes requested):

rrojan (approved):
LGTM :trophy:

PR #522, opened 2024-12-27
fix: show menu on IU task details

makeitraina:
### Tasks

- [Delete option missing for IU](https://linear.app/copilotplatforms/issue/OUT-1206/delete-option-missing-for-tasks-on-iu-end)

### Changes

Show the menu on task details when we see the user type is internal OR we are in preview mode (client/company details).

previously the logic we had was  
```
params.user_type === UserType.INTERNAL_USER || (!!getPreviewMode(tokenPayload) && <MenuBoxContainer />)
```
In the above is the user type is true, the `||` does not get executed and we render nothing. This is why the menu was not shown.

### Testing Criteria

- skipped, this is an attempt to follow up quickly on a bug found and I haven't setup my env fully for e2e testing.


rrojan (approved):
LGTM :trophy:

PR #525, opened 2024-12-30
OUT-1200 | Add animations for Activity Log component

arpandhakal:
### Changes

- [x] added collapse animation on ```activityWrapper``` component using ```TransitionGroup```. The animation is applied when new item is inserted and deleted.

### Testing Criteria

- [LOOM](https://www.loom.com/share/b57f08951ebf4d5fb2b6b64d00aacd59)

### DIFF
--- a/src/app/detail/ui/ActivityLog.tsx
+++ b/src/app/detail/ui/ActivityLog.tsx
@@ -95,7 +95,7 @@ export const ActivityLog = ({ log }: Prop) => {
   }
 
   return (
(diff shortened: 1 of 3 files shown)


rrojan (changes requested):
When i add an activity directly after a comment the component first flies over the comment, then disappears and only then appears with the proper transition

[Screencast from 2024-12-31 13-03-03.webm](https://github.com/user-attachments/assets/d4b5711b-5390-4556-b745-800322b7239f)

PR #530, opened 2025-01-02
OUT-1219, OUT-1225 | Double animations on comments.

arpandhakal:
### Changes

- [x] used ```tempId``` for animation key by creating a map state for optimistic data and comparing to use the ```tempId```, only for  those users who created the comment, once they refresh the page, the created comments will have their actual ```log.id``` as animation key. Also, users who are connected in realtime will have ```log.id``` as animation key.

### Testing Criteria

- [LOOM](https://www.loom.com/share/d112b003f1c1421db92a4c4d3ea59367)


rrojan (approved):
LGTM :trophy: :trophy:

PR #531, opened 2025-01-03
OUT-1228 | CU status change not added to activity feed

arpandhakal:
### Changes

- [x] added task Activity log for the method ```clientUpdateTask``` which was missing previously. 

### Testing Criteria

image


rrojan (approved):
LGTM :trophy:

PR #532, opened 2025-01-03
OUT-1229 | Vertical line does not extend all the way through each of the profile icons

arpandhakal:
### Changes

- [x] added styles in ```transitionGroup``` wrapper instead of adding separate styles in ```Comment``` and ```ActivityLog``` component. 

### Testing Criteria

image


rrojan (changes requested):
image

- Top padding of header should be 16px not 12

rrojan (approved):
LGTM

PR #534, opened 2025-01-03
OUT-1230 | Prevent passing unrecognized props to the DOM from custom components

arpandhakal:
### Changes

- [x] `basePadding`, `borderColor` and `borderLess` prop from `StyledTextField` are not forwarded to the DOM.

### Testing Criteria

- [x] No design regression are noticed on components using `StyledTextField` after applying the changes


rrojan (commented):
I see we're still passing `prop !== 'padding'` here. Could you please verify which error this solves?

rrojan (approved):

PR #535, opened 2025-01-03
OUT-1094 | Removing CU notifications for archived tasks

rrojan:
### Changes

- [x] Add logic to decrease notifications counter when task is archived, and increase when task is unarchived for both client & company tasks

### Screencasts:

[Screencast from 2025-01-06 14-05-03.webm](https://github.com/user-attachments/assets/2221ef94-949b-421c-939a-d8eb29b3fee8)


arpandhakal (approved):
LGTM

PR #537, opened 2025-01-06
OUT-1233 |  Applying the same template second time doesn't work

arpandhakal:
### Changes

- [x] Introduce a boolean prop `useClickHandler` on selector component to conditionally use click event on option boxes instead of using autocomplete's `onChange` event to bypass the feature that doesn't onChange event when same options are selected multiple times.
- [x] refactored the invocation of `applyTemplate` method in `newTaskForm` by removing the useEffect and introducing a handler to apply templates.   

### Testing Criteria

- [LOOM](https://www.loom.com/share/23c3972a3acb458c8d98082e7ad67f21)


rrojan:
Let's make this a part of M8 instead. M7 has already been merged to main :bow: 

image

rrojan (approved):

PR #540, opened 2025-01-07
OUT-1239 | Apply template should append template content instead of replacing

arpandhakal:
### Changes

- [x] created `appliedTitle` and `appliedDescription` which stores applied title and description from applied templates.
- [x] We compare those variables with our existing title and description on `createTaskModal`, if they are same then we will override all the data with the incoming template, else we append the incoming template. 

### Testing Criteria

- [LOOM](https://www.loom.com/share/e7c82e2dabe24eecad9a5e22ad5307ff)

### DIFF
--- a/src/app/ui/NewTaskForm.tsx
+++ b/src/app/ui/NewTaskForm.tsx
@@ -382,13 +382,13 @@ const NewTaskFooter = ({
     )
   }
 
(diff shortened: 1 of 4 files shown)


rrojan (approved):
LGTM :trophy:

PR #542, opened 2025-01-07
OUT-1238 | Make validate-count robust by removing duplicate notifications for a single task

rrojan:
### Changes

- [x] Add logic to disallow and remove any and all duplicate notifications for tasks.

### Testing Criteria

- [x] Screencast

[Screencast from 2025-01-07 19-15-04.webm](https://github.com/user-attachments/assets/f11ac89c-aba1-482c-a059-04f787ac92d8)


sazanrjb (approved):
Approved. But need to be careful if we add multiple notifications for the same task to the same client in the future, this will cause data to be incorrect..

PR #543, opened 2025-01-08
OUT-1244 | Activity feed not updating for archived tasks after status change

arpandhakal:
### Changes

- [x] instead of checking updates in `tasks` array which contains all the task fetched according to archived/unarchived queries, added a functionality to check updates in `activeTask` which is the task the user is currently navigating on.

### Testing Criteria

- [LOOM](https://www.loom.com/share/9d5ae4e774324e158496eb22e0ac4dff)


rrojan (approved):
LGTM :trophy:

PR #544, opened 2025-01-08
OUT-1241 | "Task created" activity log doesn't show up for few old tasks

arpandhakal:
### Changes

- [x] made `initiator` nullable in `LogResponseSchema` so that in case of deleted user, parsing the schema doesn't throw an error.

### Testing Criteria

image


src/app/api/activity-logs/services/activity-log.service.ts
```diff
@@ -69,6 +69,7 @@ export class ActivityLogService extends BaseService {
     const allReplies = await commentService.getReplies(commentIds)
 
     const logResponseData = parsedActivityLogs.map((activityLog) => {
+      const initiator = copilotUsers.find((iu) => iu.id === activityLog.userId)
```
rrojan, line 72: Not a big issue but this would make our code slightly more concise

```suggestion
      const initiator = copilotUsers.find((iu) => iu.id === activityLog.userId) || null
```

src/app/api/activity-logs/services/activity-log.service.ts
```diff
@@ -82,9 +83,7 @@ export class ActivityLogService extends BaseService {
           companies,
         ),
         createdAt: activityLog.createdAt.toISOString(),
-        initiator: {
-          ...copilotUsers.find((iu) => iu.id === activityLog.userId),
-        },
+        initiator: initiator || null,
```
rrojan, line 86: After implementing previous fix:
```suggestion
        initiator,
```

rrojan (changes requested):

rrojan (approved):

rrojan (approved):

PR #560, opened 2025-01-14
OUT-1272 | No filtered task screen showing on tasks app for a split second on first load.

arpandhakal:
## Changes

- [x] changed conditions to show no tasks on filter page and board/list page. 
- [x] transferred useFilter hook from `FilterBar` to `TaskBoard`.

## Testing Criteria

- [LOOM](https://www.loom.com/share/1a708936fba24253bfeb481d0729b11c)

### DIFF
--- a/src/app/ui/TaskBoard.tsx
+++ b/src/app/ui/TaskBoard.tsx
@@ -92,7 +92,7 @@ export const TaskBoard = ({ mode }: TaskBoardProps) => {
     archivedOptions.showUnarchived &&
     !archivedOptions.showArchived
 
-  const isNoTasksWithFilter = tasks && !userHasNoFilter && !filteredTasks.length
+  const isNoTasksWithFilter = !filteredTasks.length && (!tasks.length || !userHasNoFilter)
 
   const [hasInitialized, setHasInitialized] = useState(false)
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/app/ui/TaskBoard.tsx
```diff
@@ -92,7 +93,7 @@ export const TaskBoard = ({ mode }: TaskBoardProps) => {
     archivedOptions.showUnarchived &&
     !archivedOptions.showArchived
 
-  const isNoTasksWithFilter = tasks && !userHasNoFilter && !filteredTasks.length
+  const isNoTasksWithFilter = (!tasks.length || !userHasNoFilter) && !filteredTasks.length
```
rrojan, line 96: Can `tasks` be undefined here? Looks like we did `tasks &&` here for a reason.

Thinking about it, tasks should be defined since we fetch it in server component level, but is there a same (e.g. client side filters) where tasks could be undefined / null?

rrojan (approved):

src/app/ui/TaskBoard.tsx
arpandhakal: tasks can be empty array here when both archived filters are turned off, and they take time to load once they are turned back again.

PR #576, opened 2025-01-21
OUT-1299 | Remove '+ Create task' button from header for empty state

arpandhakal:
## Changes

- [x] added workspace in redux/authSlice.
- [x] fetched portal url from workspace from selectAuthSlice.
- [x] Create task in header removed on empty task board.

## Testing Criteria

image
image


rrojan (approved):
LGTM :trophy:

PR #582, opened 2025-01-22
OUT-1307 | Show the previously selected assignee at the top of the Assignee Selector

arpandhakal:
## Changes

- [x] added currentOption prop for selector component. If currentOption prop is provided, it renders a standalone option without any group in the autocomplete options.
- [x] used generic type for options in selector component. Some refactoring work are still left. 

## Testing Criteria

image

### DIFF
--- a/src/components/inputs/Selector.tsx
+++ b/src/components/inputs/Selector.tsx
@@ -22,7 +22,7 @@ export enum SelectorType {
   TEMPLATE_SELECTOR = 'templateSelected',
 }
 type SelectorOptionsType = {
-  [K in keyof typeof SelectorType as `${(typeof SelectorType)[K]}`]: K extends 'ASSIGNEE_SELECTOR'
+  [K in keyof typeof SelectorType as (typeof SelectorType)[K]]: K extends 'ASSIGNEE_SELECTOR'
     ? IAssigneeCombined
     : K extends 'STATUS_SELECTOR'
       ? WorkflowStateResponse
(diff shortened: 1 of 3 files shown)


src/components/inputs/Selector.tsx
```diff
@@ -98,6 +109,16 @@ export default function Selector({
...
+  const processedOptions = useMemo(() => {
+    if (!currentOption) return options
+    const filteredOptions = options.filter((option) => option.id !== currentOption.id)
+    return [currentOption, ...filteredOptions]
+  }, [currentOption, options])
+
+  const standaloneOptionIds = useMemo(() => {
+    return currentOption ? new Set([currentOption.id]) : new Set()
+  }, [currentOption])
+
```
rrojan, line 121: Looks good, but would be helpful to have some comments here for us to understand why we're doing this in the future

src/components/inputs/Selector.tsx
```diff
@@ -21,13 +21,22 @@ export enum SelectorType {
   STATUS_SELECTOR = 'statusSelector',
   TEMPLATE_SELECTOR = 'templateSelected',
 }
+type SelectorOptionsType = {
+  [K in keyof typeof SelectorType as `${(typeof SelectorType)[K]}`]: K extends 'ASSIGNEE_SELECTOR'
```
rrojan, line 25:
```suggestion
  [K in keyof typeof SelectorType as typeof SelectorType[K]]: K extends 'ASSIGNEE_SELECTOR'
```

I believe this should work as well. Can you please confirm

rrojan: It might also be a good idea to extract the type definition from `keyof` in a separate type / enum so we can simply do `[K in SelectorTypeKeys]`... but this is not required.

rrojan (changes requested):

rrojan (approved):
LGTM :trophy:

PR #583, opened 2025-01-23
OUT-1309 | Implement Confirmation Modal UI

arpandhakal:
## Changes

- [x] Implemented the new reassignment confirmation modal on reassignment of tasks.
- [x] Created ConfirmUI component, separated it from ConfirmDeleteUI for specific purposes in the future.
- [x] Implemented ConfirmUI component on each reassignment, in other PR, based on reassignment logic in PRD, confirmUI shall be invoked.

## Testing Criteria

- [LOOM](https://www.loom.com/share/fa98551f3be848429d12f982970fd87f). Tested on mobile view too.

### DIFF
--- a/src/app/detail/ui/Sidebar.tsx
+++ b/src/app/detail/ui/Sidebar.tsx
@@ -13,7 +13,7 @@ import { StyledBox, StyledModal } from './styledComponent'
 import { getAssigneeTypeCorrected } from '@/utils/getAssigneeTypeCorrected'
 import { UpdateTaskRequest } from '@/types/dto/tasks.dto'
 import { createDateFromFormattedDateString, formatDate } from '@/utils/dateHelper'
-import { selectTaskDetails, setShowConfirmAssignModal, setShowSidebar } from '@/redux/features/taskDetailsSlice'
+import { selectTaskDetails, toggleShowConfirmAssignModal, setShowSidebar } from '@/redux/features/taskDetailsSlice'
 import { ToggleButtonContainer } from './ToggleButtonContainer'
 import { NoAssignee } from '@/utils/noAssignee'
 import { WorkflowStateSelector } from '@/components/inputs/Selector-WorkflowState'
@@ -88,6 +88,17 @@ export const Sidebar = ({
   const windowWidth = useWindowWidth()
   const isMobile = windowWidth < 600 && windowWidth !== 0
 
+  const handleAssigneeChange = (assigneeValue: IAssigneeCombined) => {
+    updateAssigneeValue(assigneeValue)
+    const assigneeType = getAssigneeTypeCorrected(assigneeValue)
+    updateAssignee(assigneeType, assigneeValue?.id)
+  }
+
+  const handleConfirmAssigneeChange = (assigneeValue: IAssigneeCombined) => {
+    handleAssigneeChange(assigneeValue)
+    store.dispatch(toggleShowConfirmAssignModal())
+  }
+
   useEffect(() => {
     if (isMobile) {
       store.dispatch(setShowSidebar(false))
@@ -137,11 +148,9 @@ export const Sidebar = ({
             placeholder="Set assignee"
             getSelectedValue={(newValue) => {
               const assignee = newValue as IAssigneeCombined
-              // updateAssigneeValue(assignee)
-              // const assigneeType = getAssigneeTypeCorrected(assignee)
-              // updateAssignee(assigneeType, assignee?.id) //Conditionally open modal or just perform the action (commented part) without the modal based on the reassignment logic in the future.
+              // handleAssigneeChange(assignee) Conditionally open modal or just perform the action (commented part) without the modal based on the reassignment logic in the future.
               setSelectedAssignee(assignee)
-              store.dispatch(setShowConfirmAssignModal())
+              store.dispatch(toggleShowConfirmAssignModal())
             }}
             startIcon={
               (assigneeValue as IAssigneeCombined)?.name == 'No assignee' ? (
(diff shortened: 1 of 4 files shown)


src/components/layouts/ConfirmUI.tsx
```diff
@@ -0,0 +1,64 @@
+'use client'
+
+import { StyledBox } from '@/app/detail/ui/styledComponent'
+import { AppMargin, SizeofAppMargin } from '@/hoc/AppMargin'
+import { Box, Stack, Typography, styled } from '@mui/material'
+import { SecondaryBtn } from '../buttons/SecondaryBtn'
+import { PrimaryBtn } from '../buttons/PrimaryBtn'
+
```
rrojan, line 8: Use aliased imports

src/app/detail/ui/Sidebar.tsx
```diff
@@ -363,6 +393,30 @@ export const Sidebar = ({
...
+          handleConfirm={() => {
+            updateAssigneeValue(selectedAssignee)
+            if (selectedAssignee) {
+              const assigneeType = getAssigneeTypeCorrected(selectedAssignee)
+              updateAssignee(assigneeType, selectedAssignee?.id)
+            }
+            store.dispatch(setShowConfirmAssignModal())
+          }}
+          buttonText="Reassign"
+          description={`You're about to reassign this task from ${getAssigneeName(assigneeValue)} to ${getAssigneeName(selectedAssignee)}. This will give ${getAssigneeName(selectedAssignee)} access to all task comments and history.`}
```
rrojan, line 416: This callback is repeated twice. Let's be sure to extract this to a function and pass that to `handleConfirm` instead.

src/app/detail/ui/Sidebar.tsx
```diff
@@ -363,6 +393,30 @@ export const Sidebar = ({
           </Box>
         </Stack>
       </AppMargin>
+      <StyledModal
+        open={showConfirmAssignModal}
+        onClose={() => store.dispatch(setShowConfirmAssignModal())}
+        aria-labelledby="delete-task-modal"
+        aria-describedby="delete-task"
+      >
```
rrojan, line 401: I believe this is the reassignment modal:)

src/app/detail/ui/Sidebar.tsx
```diff
@@ -363,6 +393,30 @@ export const Sidebar = ({
...
+        <ConfirmUI
+          handleCancel={() => {
+            setSelectedAssignee(undefined)
+            store.dispatch(setShowConfirmAssignModal())
+          }}
+          handleConfirm={() => {
+            updateAssigneeValue(selectedAssignee)
+            if (selectedAssignee) {
+              const assigneeType = getAssigneeTypeCorrected(selectedAssignee)
```
rrojan, line 410: Might also want to check if the selected assignee is the same as the previously selected assignee to prevent rerenders and unnecessary computations

src/components/layouts/ConfirmUI.tsx
```diff
@@ -0,0 +1,64 @@
...
+interface Prop {
+  handleCancel: () => void
+  handleConfirm: () => void
+  buttonText: string
+  title?: string
+  description?: string
```
rrojan, line 14: It's good practice to make this a non-optional field and provide the fallback text here instead of in the JSX

src/components/layouts/ConfirmUI.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+const UIContainer = styled(Box)(({ theme }) => ({
+  margin: '0 auto',
+  position: 'absolute',
+  top: '50%',
+  left: '50%',
+  transform: 'translate(-50%, -50%)',
+  borderRadius: '4px',
+  backgroundColor: '#ffffff',
```
rrojan, line 61: I believe we changed the backdrop color a while ago. Can you confirm whether this is in-line with that change?

src/redux/features/taskDetailsSlice.ts
```diff
@@ -33,11 +35,15 @@ const taskDetailsSlice = createSlice({
     setTask: (state, action: { payload: TaskResponse }) => {
       state.task = action.payload
     },
+    setShowConfirmAssignModal: (state) => {
+      state.showConfirmAssignModal = !state.showConfirmAssignModal
```
rrojan, line 39:
```suggestion
    toggleShowConfirmAssignModal: (state) => {
      state.showConfirmAssignModal = !state.showConfirmAssignModal
```

src/components/layouts/ConfirmUI.tsx
```diff
@@ -0,0 +1,64 @@
+'use client'
+
+import { StyledBox } from '@/app/detail/ui/styledComponent'
+import { AppMargin, SizeofAppMargin } from '@/hoc/AppMargin'
+import { Box, Stack, Typography, styled } from '@mui/material'
+import { SecondaryBtn } from '../buttons/SecondaryBtn'
+import { PrimaryBtn } from '../buttons/PrimaryBtn'
+
+interface Prop {
+  handleCancel: () => void
```
rrojan, line 10:
```suggestion
interface ConfirmUIProps {
  handleCancel: () => void
```

It's good convention to name these Component name + "Props"

rrojan (changes requested):

src/app/detail/ui/Sidebar.tsx
arpandhakal: I dont think this is necessary. If the same value as previous assignee is selected for reassignment. MUI autocomplete doesn't trigger the `onChange` function, thus the callback `getSelectedValue` will not be invoked.

src/components/layouts/ConfirmUI.tsx
arpandhakal: This is the background color for the ui of the modal. The backdrop is handled by `StyledModal`. Still we should use color value from theme.

src/app/detail/ui/Sidebar.tsx
rrojan: Got it. Shouldn't be necessary then

rrojan (approved):
LGTM :trophy:

PR #586, opened 2025-01-23
OUT-1318 | Reassignment selector on sidebar issues

arpandhakal:
## Changes

- [x] added a loader and removed currentOptions assignee from selector options on sidebar if the filtered option does not contain it

## Testing Criteria

image
image


rrojan (approved):
LGTM :trophy:

PR #587, opened 2025-01-24
OUT-1321 | Reassignment modal should close after click on assignees name

arpandhakal:
## Changes

- [x] used onMouseDown instead of onClick to open the popper for selectors. 

## Testing Criteria

- [LOOM](https://www.loom.com/share/1db382973b224d1aa39ed1d8ab3fcf72)


rrojan (approved):
Unfortunate hack but this should get the job done just as well.

LGTM :trophy:

PR #593, opened 2025-01-28
OUT-1328 | Images are not loading in task description, disappearing after upload

arpandhakal:
## Changes

- [x] Added a client side fetching for updating assignee change from sidebar. This is being used instead of server actions. Server actions are of blocking behaviour which causes other consecutive server actions on hold if the previous actions takes time. handling assignee change on server actions takes too much time because of notification if an assignee is a company.


rrojan (approved):
LGTM :trophy:

PR #596, opened 2025-01-29
OUT-1333 | Total list of tasks is not being rendered for Client Access Limited IU

rrojan:
## Changes

- [x] Raise assignee fetch limit
- [x] Clean code to use reusable constants instead 

## Testing Criteria

- [x] Notice that the task count in columns for IU & Notification count (total number of tasks) for Client is the same

image

image


arpandhakal (approved):
LGTM

PR #597, opened 2025-01-29
OUT-1340 | Tasks do not disappear in real-time when task reassigned

arpandhakal:
## Changes

- [x]  removed filter on supabase channel listening function to invoke change when clientId or clients' company Id is equal to new assigneeId of updated tasks.
- [x] handled the filter separately on update changes function to also remove the tasks from client's end and redirect to board if the client is on details page.

## Testing Criteria

- [LOOM](https://www.loom.com/share/45ecf87e334342a28167abe5fbdd800b)


rrojan (approved):
LGTM :trophy: 
Great job with this one

PR #601, opened 2025-01-30
OUT-1322 | Text on task properties should not be selectable

arpandhakal:
## Changes

- [x] Disabled text selections on button-less selectors' values with `userSelect : 'none'`

## Testing Criteria
- [LOOM](https://www.loom.com/share/89cff42d34e8472cb25b56acd93af7b5)


rrojan (changes requested):
Let's disable this for display drawer too,while we're at it
image

PR #602, opened 2025-01-30
OUT-1339 | Selected assignee is not shown at the top of the Assignee Selector on page refresh

arpandhakal:
## Changes

- [x] added an absolute comparing mechanism by id instead of a reference check for checking weather currentOption is in list of option which previously returned always false in the first render. 

## Testing Criteria

- [LOOM](https://www.loom.com/share/aa00b8722f4d4fc3a380937ab5924875)


rrojan (approved):
LGTM :trophy:

PR #619, opened 2025-02-07
OUT-1377 | Task description dissapearing when opening tasks.

arpandhakal:
## Changes

- [x] added a check for updating task description to prevent it from updating to empty on first load.


rrojan (approved):
LGTM :trophy:

PR #621, opened 2025-02-10
OUT-1390 | Fix Unique Constraint issue in InternalUserNotifications table while deleting notifications

rrojan:
## Changes

- [x] The cause of the unique constraint violation was that multiple tasks were being softdeleted at the same time. We have the unique constraint on taskId + userId + deletedAt.
- [x] Hard delete these notifications since we are deleting them from Copilot, not marking them as read. 

## Testing Criteria

- [x] Screencast (The notification dispatch times are very long because of a temporary infra issue on Trigger's end)

 
[Screencast from 2025-02-10 16-59-58.webm](https://github.com/user-attachments/assets/b09f3c2d-6ad9-4754-acf1-2f035be75191)

- [x] Successful job run

image


arpandhakal (approved):
LGTM

PR #624, opened 2025-02-11
OUT-1380 | Comment box designs

arpandhakal:
## Changes

- [x] Added a mechanism to refresh srcs of images while on activity fetch.
- [x]  Added styles and effects on comment input while a file is being dragged.
- [x]  Added attachment icon alongside submit button in comment box with a new tapwrite update, version 1.1.83

## Testing Criteria

- [LOOM](https://www.loom.com/share/d990af0b4d97425aaa8303e1e4c4b013)

### DIFF
--- a/src/components/inputs/CommentInput.tsx
+++ b/src/components/inputs/CommentInput.tsx
@@ -153,7 +153,7 @@ export const CommentInput = ({ createComment, task_id }: Prop) => {
           parentContainerStyle={{
             width: '100%',
             height: '100%',
-            maxWidth: '600px',
+            maxWidth: '566px',
             display: 'flex',
             flexDirection: 'column',
           }}
--- a/src/components/inputs/CommentInput.tsx
+++ b/src/components/inputs/CommentInput.tsx
@@ -12,7 +12,7 @@ import { getMentionsList } from '@/utils/getMentionList'
 import { deleteEditorAttachmentsHandler, uploadImageHandler } from '@/utils/inlineImage'
 import { ArrowUpward } from '@mui/icons-material'
 import { IconButton, InputAdornment, Stack } from '@mui/material'
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { Tapwrite } from 'tapwrite'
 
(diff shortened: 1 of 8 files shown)


rrojan:
Please resolve conflicts here @arpandhakal

src/components/inputs/CommentInput.tsx
```diff
@@ -77,6 +80,43 @@ export const CommentInput = ({ createComment, task_id }: Prop) => {
...
   }, [detail, isListOrMenuActive]) // Depend on detail to ensure the latest state is captured
 
+  const uploadFn = token
+    ? async (file: File) => {
+        if (activeTask) {
+          const fileUrl = await uploadImageHandler(file, token ?? '', activeTask.workspaceId, task_id)
+          return fileUrl
+        }
+      }
+    : undefined
```
rrojan, line 90:
```suggestion
  const uploadFn = token && async (file: File) => {
        if (activeTask) {
          const fileUrl = await uploadImageHandler(file, token ?? '', activeTask.workspaceId, task_id)
          return fileUrl
        }
      }
```
This should work if token is string | undefined

src/components/inputs/CommentInput.tsx
```diff
@@ -77,6 +80,43 @@ export const CommentInput = ({ createComment, task_id }: Prop) => {
...
+
+  const handleDragEnter = () => {
+    dragCounter.current += 1
+    if (!isDragging) {
+      setIsDragging(true)
+    }
+  }
+
+  const handleDragLeave = () => {
+    dragCounter.current -= 1
```
rrojan, line 109: Could you please explain the need for dragCounter here... don't really catch the reason for it. Don't `isDragging` and `dragCounter` do the same thing?

rrojan (changes requested):

src/components/inputs/CommentInput.tsx
arpandhakal: dragcounter is used because there could be multiple events firing while dragging, the events could be of same type or different type, for example drag enter, drag leave or other. To track the multiple events and stop flickering. The reason there are multiple events is because of nested dom structure. Lets say, there is a parent container and multiple child containers nested inside, if the drag is left from the innermost child container, then drag leave is fired causing setIsDragging to be false but since the drag is not out of the parent container, there is flickering.

src/components/inputs/CommentInput.tsx
arpandhakal: I dont think this is possible because this is passed directly to tapwrite prop `uploadFn` which expects something like `(file : File) => promise('string' | undefined')` and not `(" " | (file : File) => promise('string' | undefined'))`

rrojan (approved):
LGTM :trophy:

PR #626, opened 2025-02-12
OUT-1396 | Assignee selector glitches up and down when active

arpandhakal:
## Changes

- [x] memoized and used callback on  the handle placement change function.
- [x] added an extra check to change the placement of the autocomplete.
- By doing the above mentioned steps, we actually update the state of the placement only when it is required and not in every render/resize which was leading to an infinite rendering.   

## Testing Criteria

- [LOOM](https://www.loom.com/share/0eeee5df463643aa8f04196dbf3f1de3)


rrojan (approved):
LGTM 🏆

PR #628, opened 2025-02-13
OUT-1404 | resetting active task in board page

arpandhakal:
## Changes

- [x] added a mechanism to reset active task when navigated back to board page


rrojan (approved):
LGTM 🏆

PR #630, opened 2025-02-14
OUT-1406 | 'Deleted User' shows in place of user name in activity log and properties

arpandhakal:
## Changes

- [x] Shifted assignee fetched to the top of parent client component so that it isnt holded by a suspense and assignee are fetched faster. 
- [x] Copilot Usernames memoized based on assignee list var change and current log's initiator.  

## Testing Criteria

- [LOOM](https://www.loom.com/share/818a6b942faf46a8ab98e36e262695e4)


rrojan (approved):
LGTM :rocket:

PR #637, opened 2025-02-18
OUT-1413 | Cannot insert a new line and type below an image

arpandhakal:
## Changes

- [x]  new tapwrite update, 1.1.86 which supports hardbreaking for images and attachments.
- [x]  validated that empty comments cannot be edited/posted and reseting the tiptap instance on cancel edit.

## Testing Criteria

- [LOOM](https://www.loom.com/share/661722727ff345578b6d7fb8dcf6b3c8)


src/components/cards/CommentCard.tsx
```diff
@@ -65,7 +65,7 @@ export const CommentCard = ({
   const windowWidth = useWindowWidth()
   const isMobile = windowWidth < 600 && windowWidth !== 0
 
-  const [editedContent, setEditedContent] = useState('')
+  const [editedContent, setEditedContent] = useState((comment.details as { content: string }).content)
```
rrojan, line 68:
```suggestion
  const content = (comment.details as { content: string }).content || ''
  const [editedContent, setEditedContent] = useState(content)
```

You don't have to do the same long type cast in three different places then.

rrojan (changes requested):

rrojan (approved):
LGTM :trophy:

PR #643, opened 2025-02-19
OUT-1424 | Remove console log for 'Comment cannot be empty'

arpandhakal:
## Changes

- [x] handled enter key events on comment input only when the comment input is focused. Previously, even if we were editing description then they enter key event was triggered for comment input. 
- [x] removed the log.


rrojan (approved):
LGTM :rocket:

PR #655, opened 2025-02-24
OUT-1429 | Split first and last name in activity feed to avoid new lines

rrojan:
## Changes

- [x] Make activity log `display: block` and corresponding elements `inline` instead of using flexbox

## Testing Criteria

- [x] Screenshot

image


sazanrjb (approved):

src/app/detail/ui/ActivityLog.tsx
sazanrjb, line 1: lets look into these errors in possible as well..

PR #661, opened 2025-02-28
OUT-1467 | Tasks reposition on hover in board view

arpandhakal:
## Changes

- [x] the main reason for this issue was our sorting method for tasks while rendering them in a board. Tasks are sorted according to due dates, if due dates are not there, then they are sorted by createdAt timestamp. The issue was arising because of createdAt timestamp becoming equal for certain tasks.
- [x] So added another condition for two task having same value of createdAt, sorted them by id alphabetically.

## Testing Criteria

- [LOOM](https://www.loom.com/share/50bb5072308940eba8a7282e275f0800?focus_title=1&muted=1&from_recorder=1)

### DIFF
--- a/src/utils/sortTask.ts
+++ b/src/utils/sortTask.ts
@@ -1,6 +1,7 @@
 interface Sortable {
   dueDate?: string
   createdAt: Date
+  id: string
 }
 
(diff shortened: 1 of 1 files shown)


rrojan (approved):
LGTM 🏆

PR #663, opened 2025-03-03
OUT-1457 | Add endpoint to get comments with replies

rrojan:
## Changes

- [x] Add query to return latest 3 replies in GET activity-logs endpoint itself
- [x] Add support for `expandComments` URL param. Comment IDs passed to this param will return all replies, not just the latest 3.
- [x] Clean code and add util to get intersection of arrays  

## Testing Criteria

- [x] Screencast

https://drive.google.com/file/d/1n8gSL5pgA-SakoufzQzQ3r8OLouI7i6p/view?usp=sharing


sazanrjb (approved):
When a user expands a comment, is there a use case to send multiple limited replies, and all replies in a single request?

I feel like building an API which fetches all the replies of that parent comment is more simple and reusable instead of handling 3 replies and all replies using the same endpoint.

```
getCommentsReplies(commentIds: string[], limit = 3);

getCommentReplies(commentId: string);
```

I may not know the complete use case though..

PR #677, opened 2025-03-11
HOTFIX : overlapping avatars design issue

arpandhakal:
## Changes

- [x] fixed messed up design in collapsible reply card when there is no icon image url for a user.

## Testing Criteria

image


rrojan (approved):
Thanks for the quick fix 🚀

PR #680, opened 2025-03-12
OUT-1509 | Deleted parent comment should be removed from UI if all replies are deleted

rrojan:
## Changes

- [x] Add auto-deletion of deleted parent comment when last reply is deleted

## Testing Criteria

- [x] Screencast

https://github.com/user-attachments/assets/cffc54a8-27f3-4d4f-82ea-5375cf2289c9

- Comment deletion behavior (thoroughly tested to avoid regressions)


arpandhakal (approved):
LGTM!

PR #682, opened 2025-03-12
OUT-1511 | text and avatar alignment on replies

arpandhakal:
## Changes

- [x] avatar and text alignment fix on reply card and collapsible reply card.

## Testing Criteria

image


rrojan (approved):
LGTM 🏅

PR #685, opened 2025-03-14
OUT-1519 | Reply content is not vertically centered within contianer

arpandhakal:
## Changes

- [x]  reply content vertically centered within the reply card. 

## Testing Criteria

image


rrojan (approved):
LGTM 🏆

PR #691, opened 2025-03-17
OUT-1530 | TypeError: crypto.randomUUID is not a function

rrojan:
## Changes

- [x] Use `uuid` package since `crypto` module is unstable to be run on client code unless using `window.crypto.*`. (however using `window` would make this util non-reusable for backend code, so swapped to use the main package instead)

## Testing Criteria

- [x] Screencast

https://github.com/user-attachments/assets/46d246cd-2d50-4150-9641-737ff3730a7d

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "supabase": "^2.1.1",
     "swr": "^2.2.5",
     "tapwrite": "1.1.91",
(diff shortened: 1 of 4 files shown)


arpandhakal (approved):
LGTM

PR #700, opened 2025-03-19
OUT-1468 | Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

arpandhakal:
## Changes

- [x] added shalowcompare before updating tasks which is a possible fix for multiple rerenders.


rrojan (approved):
LGTM 🏆

rrojan (approved):
LGTM 🏅

PR #719, opened 2025-03-31
OUT-1584 | Fix Realtime Behavior

arpandhakal:
## Changes

- fixed issue of sub tasks popping up in task board for iu and clients.
- only those tasks whose parent id are un-accesible to the client/iu are made to be popped up in the task board in real time.

## Testing Criteria

- [LOOM](https://www.loom.com/share/129d9a1a0a60456282a81a84789eee5a)


rrojan (commented):
There is a flaw in this logic in the sense that we only check the immediate parent of a task to check if we need to shortcircuit our insert / update event.

Here you can see that tasks nested > 1 layer will always appear on insert or update as a parent task on the board.


https://github.com/user-attachments/assets/0357a4da-650b-4652-81ce-5a3ac67a0b0e

PR #730, opened 2025-04-04
OUT-1614 | Stale task data(description, title.. others) while navigating to parent task from a sub-task using back button from browser.

arpandhakal:
## Changes

- [x] added a mechanism to re-render the server component of details page with router.refresh() method.
- [x] added a cleanup function in detail state update hoc to unset an active task when the detail page component is unmounted.

## Testing Criteria

- [LOOM](https://www.loom.com/share/37af6a10c7604cb297da2cebb7cfaf2b)


src/app/detail/ui/TaskEditor.tsx
```diff
@@ -72,14 +73,20 @@ export const TaskEditor = ({
...
         setUpdateTitle(currentTask.title || '')
         setUpdateDetail(currentTask.body ?? '')
       }
     }
   }, [activeTask, task_id, isUserTyping, activeUploads])
 
+  const router = useRouter()
+  useEffect(() => {
+    router.refresh()
+  }, [])
```
rrojan, line 88: This will cause double-loading, even on normal click redirections...

rrojan (changes requested):

PR #734, opened 2025-04-09
OUT-1620 | Bug: Tasks with inaccessible subtasks appear in card should not show '0' as subtask count

rrojan:
## Changes

- [x] Fix subtaskCount meta item rendering logic to not show count when count is 0, even for tasks without inaccessible children

## Testing Criteria

- [x] Screenshot

image

Previously this used to falsely show "0"


arpandhakal (approved):
LGTM

PR #749, opened 2025-04-16
OUT-1650 | Exclude subtasks from CU notification counter

rrojan:
## Changes

- [x] Only show notifications to CU for thats that are either (i) Parent / standalone tasks (ii) Disjoint tasks (Client cannot access parent task)
- [x] Fix Validate Count job to use this criteria to validate notification count for clients on each refresh. So when new clients first load the tasks app, the count is fixed for them automatically
- [x] QOL: TempleteFetcher was being used which caused a 401 error to be thrown everytime client board was loaded, it has been removed  

## Testing Criteria

- [x] Screencast

https://github.com/user-attachments/assets/d5dfcb48-afd6-436c-ad60-dcefa31a5a80

Sometimes the notification counter hangs and is fixed by refresh


arpandhakal (approved):
LGTM!

PR #756, opened 2025-04-18
OUT-1648 | Filter limited access IU from assignee dropdown in subtasks

arpandhakal:
## Changes

- [x] added a separate optional param in users api called clientCompanyId. The param is passed from details page if the current task is assigned to a client or company.
- [x] filtered out limited access internal uses from assignee dropdown list in subtask creation and subtask list if the parent task is assigned to client/company out of scope for the iu.
- [x] created a separate redux state to store the filtered out assignee list array for a task detail page.

## Testing Criteria

- [LOOM](https://www.loom.com/share/278e41a440d741bda123559bff69078f)

## Notes

- User can still reassign the parent task to a client out of scope for an iu. And if the subtasks of the parent task is assigned to limited access IU, then there is no way for the IU to access the sub task once the parent is reassigned to a client.

### DIFF
--- a/src/app/_fetchers/AssigneeFetcher.tsx
+++ b/src/app/_fetchers/AssigneeFetcher.tsx
@@ -54,10 +54,12 @@ export const AssigneeFetcher = async ({
   task,
   clientCompanyId,
 }: AssigneeFetcherProps) => {
-  const assignableUsersWithType = addTypeToAssignee(await fetchAssignee(token, userType, isPreview))
-  const assignableUsersWithTypeForLimitedTask = clientCompanyId
-    ? addTypeToAssignee(await fetchAssignee(token, userType, isPreview, clientCompanyId))
-    : []
+  const [assignableUsersWithType, assignableUsersWithTypeForLimitedTask] = await Promise.all([
+    fetchAssignee(token, userType, isPreview).then(addTypeToAssignee),
+    clientCompanyId
+      ? fetchAssignee(token, userType, isPreview, clientCompanyId).then(addTypeToAssignee)
+      : Promise.resolve([]),
+  ])
   return (
     <ClientSideStateUpdate
       assignee={assignableUsersWithType}
--- a/src/app/_fetchers/AssigneeFetcher.tsx
+++ b/src/app/_fetchers/AssigneeFetcher.tsx
@@ -16,9 +16,15 @@ interface AssigneeFetcherProps extends PropsWithToken {
   userType?: UserType
   isPreview?: boolean
   task?: TaskResponse
+  clientCompanyId?: string
 }
 
-const fetchAssignee = async (token: string, userType?: UserType, isPreview?: boolean): Promise<IAssignee> => {
+const fetchAssignee = async (
+  token: string,
+  userType?: UserType,
+  isPreview?: boolean,
+  clientCompanyId?: string,
+): Promise<IAssignee> => {
   if (userType === UserType.CLIENT_USER && !isPreview) {
     const res = await fetchWithRetry(`${apiUrl}/api/users/client?token=${token}&limit=${MAX_FETCH_ASSIGNEE_COUNT}`, {
       next: { tags: ['getAssigneeList'] },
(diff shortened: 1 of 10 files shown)


src/app/_fetchers/AssigneeFetcher.tsx
```diff
@@ -31,16 +37,34 @@ const fetchAssignee = async (token: string, userType?: UserType, isPreview?: boo
...
+  userType,
+  viewSettings,
+  isPreview,
+  task,
+  clientCompanyId,
+}: AssigneeFetcherProps) => {
   const assignableUsersWithType = addTypeToAssignee(await fetchAssignee(token, userType, isPreview))
+  const assignableUsersWithTypeForLimitedTask = clientCompanyId
+    ? addTypeToAssignee(await fetchAssignee(token, userType, isPreview, clientCompanyId))
+    : []
```
rrojan, line 60: `/api/users` is very slow due to client access filtering and calling endpoints synchronously will double the already long time we get stuck with loading skeletons.

Please use Promise.all here

src/app/api/users/users.service.ts
```diff
@@ -65,10 +65,32 @@ class UsersService extends BaseService {
...
+    let filteredIUs = internalUsers
+    if (companyId) {
+      filteredIUs = filteredIUs.filter((iu) => {
+        if (iu.isClientAccessLimited) {
+          return iu.companyAccessList?.includes(companyId)
+        } else {
+          return true
+        }
+      })
+    }
```
rrojan, line 89:
```suggestion
    if (companyId) {
  filteredIUs = filteredIUs.filter(iu =>
    !iu.isClientAccessLimited || iu.companyAccessList?.includes(companyId)
  );
}
```

src/app/detail/ui/NewTaskCard.tsx
```diff
@@ -53,6 +54,8 @@ export const NewTaskCard = ({
   handleSubTaskCreation: (payload: CreateTaskRequest) => void
 }) => {
   const { workflowStates, assignee, token, filterOptions, previewMode, activeTask } = useSelector(selectTaskBoard)
+  const { assigneeListForLimitedTasks } = useSelector(selectTaskDetails)
```
rrojan, line 57:
```suggestion
  const { activeTaskAssignees } = useSelector(selectTaskDetails)
```
Might be a better name

src/app/detail/ui/TaskCardList.tsx
```diff
@@ -101,6 +106,13 @@ export const TaskCardList = ({ task, variant, workflowState, mode, handleUpdate
     store.dispatch(setConfirmAssigneeModalId(undefined))
   }
 
+  const getClientCompanyId = () => {
+    if (variant == 'subtask' && activeTask) {
+      return activeTask.assigneeType !== 'internalUser' ? activeTask.assigneeId : undefined
```
rrojan, line 111: Use `AssigneeType` or related enum instead of hardcoding string values here

src/app/api/users/users.controller.ts
```diff
@@ -12,10 +12,12 @@ export const getUsers = async (req: NextRequest) => {
   const userType = req.nextUrl.searchParams.get('userType') || undefined
   const limit = rawLimit ? +rawLimit : undefined
   const nextToken = req.nextUrl.searchParams.get('nextToken') || undefined
+  const clientCompanyId = req.nextUrl.searchParams.get('clientCompanyId') || undefined
```
rrojan, line 15: Is there a reason why this is called `clientCompanyId` and not `companyId`? Filtering is done on the basis of `companyId` regardless of if it is a client task or company task right?

src/utils/users.ts
```diff
@@ -51,14 +51,21 @@ export const setDebouncedFilteredAssignees = (
...
     clearTimeout(activeDebounceTimeoutId)
   }
   const newTimeoutId = setTimeout(async () => {
     setLoading(true)
-
-    const newAssignees = await getAssigneeList(z.string().parse(token), newInputValue, 10000, '0', filterOptions)
+    const newAssignees = await getAssigneeList(
+      z.string().parse(token),
+      newInputValue,
+      10000,
```
rrojan, line 64: Use MAX_FETCH_ASSIGNEE_COUNT here

rrojan (changes requested):

src/app/api/users/users.controller.ts
arpandhakal: This is called clientCompanyId because we are sending client or company's id to the route as a param. And in its corresponding service, we find if its a client id(if so, we extract companyId from the client), else its a companyId.

rrojan (approved):
LGTM 🏆 
Great job with this one

PR #764, opened 2025-04-24
OUT-1673 | Add support for exposing Delete Task API

arpandhakal:
## Changes

- [x] created a delete route for public api to delete a task.
- [x] applied recursive params and its consequent functionality as per PRD.
- [x] verified response structure as per PRD.

## Testing Criteria

- [TBD]()


src/app/api/tasks/public/public.controller.ts
```diff
@@ -50,3 +50,12 @@ export const getOneTaskPublic = async (req: NextRequest, { params: { id } }: IdP
   const task = await tasksService.getOneTask(id)
   return NextResponse.json({ ...PublicTaskSerializer.serialize(task) })
 }
+
+export const deleteOneTaskPublic = async (req: NextRequest, { params: { id } }: IdParams) => {
+  const recursive = req.nextUrl.searchParams.get('recursive')
+  const parsedRecursive = recursive === 'true'
```
rrojan, line 56: Use `z.coerce` instead. Even when not stated, we should be able to parse all truthy values like 'true', 1,etc as true, and 'false', '0', etc as false

rrojan (changes requested):

src/app/api/tasks/public/public.controller.ts
```diff
@@ -50,3 +51,11 @@ export const getOneTaskPublic = async (req: NextRequest, { params: { id } }: IdP
   const task = await tasksService.getOneTask(id)
   return NextResponse.json({ ...PublicTaskSerializer.serialize(task) })
 }
+
+export const deleteOneTaskPublic = async (req: NextRequest, { params: { id } }: IdParams) => {
+  const recursive = req.nextUrl.searchParams.get('recursive')
+  const user = await authenticate(req)
+  const tasksService = new TasksService(user)
+  const task = await tasksService.deleteOneTask(id, z.coerce.boolean().parse(recursive))
+  return NextResponse.json({ ...PublicTaskSerializer.serialize(task) })
```
rrojan, line 60: After successful deletion we should be returning 204

rrojan: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204

src/app/api/tasks/tasks.service.ts
```diff
@@ -317,14 +317,19 @@ export class TasksService extends BaseService {
 
     if (!task) throw new APIError(httpStatus.NOT_FOUND, 'The requested task to delete was not found')
 
+    if (!recursive && task) {
```
rrojan, line 320: Task is sure to exist since we would throw an error otherwise

rrojan (changes requested):

src/app/api/tasks/public/public.controller.ts
arpandhakal: The PRD specifies to return 200 though. 
image

rrojan (approved):

PR #774, opened 2025-04-30
OUT-1697 | Implement deletedBy column in Tasks

arpandhakal:
## Changes

- [x] filled the new field deletedBy in task whenever the task is deleted.
- [x] exposed deletedBy on public endpoints LIST+GET+DELETE.
- [x] validated that deletedBy must exist whenever a DELETE request on a particular task is recieved.

## Testing Criteria

image

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -343,15 +343,7 @@ export class TasksService extends BaseService {
     const policyGate = new PoliciesService(this.user)
     policyGate.authorize(UserAction.Delete, Resource.Tasks)
     const deletedBy = this.user.internalUserId
-    if (!deletedBy) {
-      throw new APIError(httpStatus.BAD_REQUEST, 'Cannot delete task. could not find user.') //validation for deletedBy
-    }
-    await this.db.task.update({
-      where: { id, workspaceId: this.user.workspaceId },
-      data: {
-        deletedBy,
-      },
-    })
+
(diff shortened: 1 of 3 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -342,7 +342,16 @@ export class TasksService extends BaseService {
   async deleteOneTask(id: string, recursive: boolean = true) {
     const policyGate = new PoliciesService(this.user)
     policyGate.authorize(UserAction.Delete, Resource.Tasks)
-
+    const deletedBy = this.user.internalUserId
+    if (!deletedBy) {
+      throw new Error('Cannot delete task. could not find user.') //validation for deletedBy
+    }
+    await this.db.task.update({
```
SandipBajracharya, line 349: Can we move this update part inside the transaction same just after the task deletion?

arpandhakal: Moving the task into the transaction requires an extra fetching of the particular task for the response to public api without mutating the response. Doing this will avoid the extra call.

rrojan: @arpandhakal Since we are overriding the prisma `delete` method to update and add `deletedAt`, we can use `update instead to add `deletedAt` and `deletedBy` at the same time at line 376

rrojan (changes requested):

rrojan (approved):

PR #793, opened 2025-05-09
OUT-1734 | do not require name when templateId is present when creating a task

arpandhakal:
## Changes

- [x] name should not be required when templateId is present when creating a task from public api
- [x] moved applying template name and body to task from service to deseriazer of public create task dto separating concerns on task creation api from web and public.
- [x] applied name to be an optional body props when creating a template from public api if tempalteId is present.

## Testing Criteria

- [LOOM](https://www.loom.com/share/84ff656d4a0243bf9227702c58a08e90)

### DIFF
--- a/src/app/api/tasks/public/public.serializer.ts
+++ b/src/app/api/tasks/public/public.serializer.ts
@@ -1,3 +1,4 @@
+import APIError from '@/app/api/core/exceptions/api'
 import DBClient from '@/lib/db'
 import { RFC3339DateSchema } from '@/types/common'
 import {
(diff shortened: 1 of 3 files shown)


src/app/api/tasks/public/public.serializer.ts
```diff
@@ -77,14 +79,45 @@ export class PublicTaskSerializer {
...
+        id: templateId,
+        workspaceId: workspaceId,
+      },
+    })
+
+    if (!template) {
+      throw new APIError(httpStatus.NOT_FOUND, 'The requested template was not found')
+    }
+
+    const title = (existingTitle ?? '') + (existingTitle ? ' ' : '') + template.title
```
SandipBajracharya, line 100:
```suggestion
    const title = (existingTitle ? `${existingTitle} ` : '') + template.title
```

SandipBajracharya (approved):

PR #794, opened 2025-05-09
OUT-1736 | createdBy in body is ignored when creating a task

arpandhakal:
## Changes

- [x] added a functionality to cater createdBy prop from create task (public) body.
- [x] fetched a list of internal user and matched createdBy with error handling, and created task with the createdBy value if provided.

## Testing Criteria

image

image

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -183,7 +183,14 @@ export class TasksService extends BaseService {
     // NOTE: This block strictly doesn't allow clients to create tasks
     let createdById = z.string().parse(this.user.internalUserId)
 
-    //todo : implement data.createdBy functionality if it is provided from public api here.
+    if (data.createdById && opts?.isPublicApi) {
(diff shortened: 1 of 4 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -183,7 +183,14 @@ export class TasksService extends BaseService {
     // NOTE: This block strictly doesn't allow clients to create tasks
     let createdById = z.string().parse(this.user.internalUserId)
 
-    //todo : implement data.createdBy functionality if it is provided from public api here.
+    if (data.createdById && opts?.isPublicApi) {
+      const internalUsers = await copilot.getInternalUsers({ limit: MAX_FETCH_ASSIGNEE_COUNT }) // there are 2 of these api running on this same serive. need to think about this later.
+      const createdBy = internalUsers?.data?.find((iu) => iu.id === data.createdById)
```
SandipBajracharya, line 188: Can we use `await copilot.getInternalUser(userId)` function directly here?

arpandhakal: A deleted user could be provided for createdById, where the `await copilot.getInternalUser(userId)` gives a normal response. Apparently there is not flags like `isDeleted` to check that. So there is no way of knowing if the user is deleted.

SandipBajracharya (approved):

PR #795, opened 2025-05-12
OUT-1716 | Intermittent "Something Went Wrong" Error on Notifications Page for Task Notifications

rrojan:
## Changes

- [x] Bump NextJS to latest v14.2.28 since a fix for `undici` was patched to v14.2.20. Many people have reported that the issue still persists ([ref](https://github.com/vercel/next.js/discussions/69635)), however i was unable to repro the issue after trying for a long time after this patch (both in-product + email notifications)

- [x] After latest v14.2.x fixed the issue, there was no need to downgrade Node to 18 anymore so I switched the node version back to 20
image

## Notes

- Downgrading Node to 18 seems to be another way to fix the issue, however do to a chain of peer dependencies we cannot downgrade, the export of Next bundle on node18 in Vercel fails.

image

- Patching global `fetch` with a patched version of undici doesn't seem feasible since NextJS has its own gymnastics on top of standard fetch

## Testing Criteria

- 15 mins+ of testing across notifications without any issues
image

### Refs:
- https://github.com/vercel/next.js/pull/70082 (Author backported this fix later to 14.2.20)
- https://github.com/vercel/next.js/discussions/48192
- https://github.com/vercel/next.js/discussions/69635


arpandhakal (approved):
LGTM

PR #797, opened 2025-05-12
OUT-1736 | createdBy in body is ignored when creating a task

arpandhakal:
## Changes

- [x] Added a support on task creation service to pass createdBy and createdBy role to activity logger for logging new task creation activity.
- [x] Previously, it the userId of the activityLog would be the token user, but since createdBy can be manually passed, we needed to change it.

## Testing Criteria

- [LOOM](https://www.loom.com/share/5772aac12cb246489bc3e6eee8aeeba7)


rrojan (approved):
LGTM 🏆

PR #799, opened 2025-05-13
OUT-1742 | You can reassign a parent task with an IU assignee and subtasks for an IU to a CU or Company

rrojan:
## Changes

- [x] Make sidebar consider `clientCompanyId` and `activeTaskAssignees`

## Testing Criteria

- [] TBD


arpandhakal (approved):
LGTM. But after introducing public api feature. I think filtering assignee list in the ui dropdown partially solves the problem. We still need to validate the assignee in api methods.

PR #802, opened 2025-05-14
OUT-1751 | Implement new iu + client + company ids in Public API

arpandhakal:
## Changes

- [x] removed `assigneeId` and `assigneeType` on public tasks response on all operations. Public api will only get `internalUserId`, `clientId` and `companyId` as of now.
- [x] added support for serializing proper response of userId for `getAllTasks` and `getOneTask` methods for public API.
- [x] added support for desializing proper request of userIds for createTask and updateTask methods for public API.
- [x] added support to update userIds : `internalUserIds`, `clientIds` and `companyIds` on `createTask` and `updateOneTask` method for public API.
- [x] added validation on create task dto from public api with 2 different rules. i.e. `InternalUserId` and `(clientId | companyId)` are exclusive and `companyId` should be required when `clientId` is provided.
- [x] added a feature flag called dual assignee mode, when enabled we run functions to update userIds if create update calls are coming from web app, and another function to update `assigneeId/assigneeType` when create update calls are coming from publicAPI.
- [x] added companyId validation for clientId on create/upate task to use `client.companyIds[] `or `client.companyId` based on `dualAssigneeMode` 

## Testing Criteria

- [LOOM](https://www.loom.com/share/eaa493179b864585a6db89375bb3a8af)
- [updatedLOOM](https://www.loom.com/share/1b8673d4241c4cd2be0c1780dcdb3db1), update : fixed and tested the last part where task was seen as deleted while updating.


src/app/api/label-mapping/label-mapping.service.ts
```diff
@@ -34,6 +34,7 @@ export class LabelMappingService extends BaseService {
     const copilotUtils = new CopilotAPI(this.user.token, this.customApiKey)
 
     if (assigneeType === AssigneeType.internalUser) {
+      await copilotUtils.getInternalUser(assigneeId)
```
rrojan, line 37: Why are we fetching the current internal user but not saving it anywhere?

src/app/api/tasks/public/public.serializer.ts
```diff
@@ -139,11 +141,12 @@ export class PublicTaskSerializer {
...
       title: payload.name,
       body: payload.description,
-      assigneeId: payload.assigneeId,
-      assigneeType: payload.assigneeType,
       dueDate: rfc3339ToDateString(payload.dueDate),
       isArchived: payload.isArchived,
       workflowStateId,
+      internalUserId: payload.internalUserId ?? null,
+      clientId: payload.clientId ?? null,
+      companyId: payload.companyId ?? null,
```
rrojan, line 149: Wouldn't this update all userId fields to null if data for those fields hasn't been provided though

src/types/dto/tasks.dto.ts
```diff
@@ -29,8 +29,12 @@ export const CreateTaskRequestSchema = z
     parentId: z.string().uuid().nullish(),
     templateId: z.string().uuid().nullish(),
     createdById: z.string().uuid().optional(),
+    internalUserId: z.string().uuid().nullable().optional(),
+    clientId: z.string().uuid().nullable().optional(),
+    companyId: z.string().uuid().nullable().optional(),
```
rrojan, line 34:
```suggestion
    internalUserId: z.string().uuid().nullish(),
    clientId: z.string().uuid().nullish(),
    companyId: z.string().uuid().nullish(),
```

src/types/dto/tasks.dto.ts
```diff
@@ -42,6 +46,9 @@ export const UpdateTaskRequestSchema = z
     workflowStateId: z.string().uuid().optional(),
     dueDate: DateStringSchema.nullish(),
     isArchived: z.boolean().optional(),
+    internalUserId: z.string().uuid().nullable().optional(),
```
rrojan, line 49:
```suggestion
    internalUserId: z.string().uuid().nullish(),
```

src/types/common.ts
```diff
@@ -69,6 +69,7 @@ export const ClientResponseSchema = z.object({
   familyName: z.string(),
   email: z.string(),
   companyId: z.string(),
+  companyIds: z.array(z.string()).optional(),
```
rrojan, line 72:
```suggestion
  companyIds: z.array(z.string().uuid()).optional(),
```
Company id's are uuid's right

rrojan (changes requested):

src/app/api/tasks/public/public.serializer.ts
arpandhakal: The dto expects at least one of these ids. Also there are certain other rules applied. Like an `internalUserId` and `clientId/companyId ` cannot be provided at the same time, `companyId` must be provided if clientId is provided...

src/app/api/label-mapping/label-mapping.service.ts
arpandhakal: Currently this is required while validating assigneeId/assigneeType from web app. This shall be removed while introducing userIds in the webapp API.

rrojan (approved):
LGTM :trophy:

PR #804, opened 2025-05-19
Out 1765

arpandhakal:
## Changes

- [x] removed the outdated props `assigneeId` and `assigneeType` and introduced `internalUserId`, `clientId` and `companyId` with validations on task creation schema.
- [x] removed calculations and updates done for syncing `assigneeId`/`assigneeType` and userIds because from this feature onwards, both public API and the webApp share the same assignee format for creating a task.
- [x] created a separate utility function for userIds validation since 2 dtos from public and web can reuse them.
- [x] made necessary changes, `assignee`/`assigneeType` calculations for creating label in a task.

## Testing Criteria

- [LOOM]()


rrojan (approved):
LGTM 🏆

PR #813, opened 2025-05-28
OUT-1812 | Add DB constraint to enforce AT LEAST ONE userId column to be present for non-null assigneeID

rrojan:
… mapping

## Changes

- [x] Add constraint with following cases:

-- If assigneeType is internalUser, enforce assigneeId + internalUserId
-- If assigneeType is client, enforce assigneeId + clientId + companyId
-- If assigneeType is company, enforce assigneeId + companyId

## Testing Criteria

- [x] Screencast

https://github.com/user-attachments/assets/aea024d9-1de8-49de-b64d-d9ef2a1ebee5


arpandhakal (approved):
Nice work.

PR #822, opened 2025-06-04
OUT-1815 | There should only be one separator line in the task template dropdown

arpandhakal:
## Changes

- [x] double border on manage templates end option when on empty template options - template selector. 

## Testing Criteria

image


rrojan (approved):
LGTM 🏆

PR #824, opened 2025-06-06
OUT-1827 | Add support to implement initial values in all selectors using UserCompanySelector

arpandhakal:
## Changes

- [x] added initial values in sidebar assignee selector.
- [x] added initial values in filter by assignee button.
- [x] added initial values for create task form's assignee selection. If assignee is filtered in filter by assignee button or my tasks tab, then initial value of those assignee are added on createTasksField too.
- [x] added a mechanism to remove clients, companies and iu group from options in selector component when certain filters are applied. Only in filter by assignee button.

## Testing Criteria

- [LOOM]()


rrojan (approved):
LGTM :trophy:

PR #830, opened 2025-06-12
OUT-1872 | Tasks are not being filtered from the api for 2 different companies on client portal

arpandhakal:
## Changes

- [x] Changed getClientorCompanyAssigneeFilter() to support userIds instead of old assigneeId which was outdated for the multi companies portal.
- [x] Returned the provided companyId instead of client.companyId while validating user Ids on task creation.

## Testing Criteria

- [LOOM](https://www.loom.com/share/d8d4f7c41d3445c1ad39a00bdaa3a8be)

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -496,17 +496,16 @@ export class TasksService extends BaseService {
 
     const clientId = parsedClientId.data
     const parsedCompanyId = z.string().safeParse(this.user.companyId)
-
     if (!parsedCompanyId.data) {
       return {
-        OR: [{ assigneeId: clientId, assigneeType: 'client' }],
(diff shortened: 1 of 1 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -876,14 +875,18 @@ export class TasksService extends BaseService {
         throw new APIError(httpStatus.BAD_REQUEST, `Invalid clientId`)
       }
 
+      if (!companyId) {
+        throw new APIError(httpStatus.BAD_REQUEST, `CompanyId is required for client tasks`)
+      }
+
```
rrojan, line 881: We are doing data validation on DTO level so this should not be necessary
image

rrojan (changes requested):

PR #832, opened 2025-06-13
OUT-1888 | Label wrong for a client task with different companyId.

arpandhakal:
## Changes

- [x] changed label mapping service getLabel method parameters to use userIds instead of assigneeId/assigneeType
- [x] applied relevant changes on createTask and updateTask to create labels. the changes being arguments for the getLabel method.

## Testing Criteria

image


rrojan (approved):
LGTM 🏆

PR #835, opened 2025-06-16
TRYFIX OUT-1897 | 'Something went wrong' error in production workspace after clicking on a task in the Tasks App as an IU

arpandhakal:
## Note

Problem that was noticed : userIds rolling back to null after updating due date/ workflowstate from details page.

## Changes

- [x] modified a condition where the update task function strictly updates userIds only when task is reassigned and not when due date and workflowstates are changed.


rrojan (approved):
LGTM :trophy:

PR #843, opened 2025-06-19
OUT-1910 | Sub task list's assignee is change but the avatar is not changing on real time.

arpandhakal:
## Changes

- [x] AssigneeValue was not being cached and updated in TaskCardList which was effecting list view and sub task list cards. The assignee was not being changed in realtime in avatar only.
- [x] Also fixed the issue in sidebar.

## Testing Criteria

- [LOOM](https://www.loom.com/share/c494536ff6e7413592e8bebc9d6f844a)


rrojan (approved):
LGTM 🏆

PR #850, opened 2025-06-23
OUT-1935 | Activity logs shows Company users as undefined when task is unassigned from the company Users.

arpandhakal:
## Changes

- [x] added getAssigneeName for task unassignment log instead of using given name and family name. Companies donot have given name and family names.
- [x] added undefined checks for updated userIds in updateOneTask service. Before this fix, the issue was task being unassigned on every task update, even when assignee was not being updated.

## Testing Criteria

image


rrojan (approved):
LGTM 🏆 
Let's make sure to test other parts of the application to see what new changes we have to do to accomodate `companyIds` as well

PR #853, opened 2025-06-23
OUT-1934 | Add implementation for new companyIds[] response from copilot sdk's getClients()

arpandhakal:
## Changes

- [x] flattened client list according to companyIds array in clients for selector options.
- [x] added support for incorporating companyIds[] instead of companyId for limited access IU. Fetched users for limited access IU on the basis of companyAccessList and client's companyIds[].
- [x] added a fix to support no assignee tasks for limited access ius. 

## Testing Criteria

- Coming soon..


rrojan (approved):
LGTM :trophy:

PR #855, opened 2025-06-24
OUT-1943 | Error in client details

arpandhakal:
## Changes

- [x] fixed zod error by setting userIds in filterOptions.Assignee instead of string in PreviewMode.ts' handlePreviewMode
- [x] changed logic for IUs on disjointTaskFilters and getParentIdFilter adding a check for user.clientId. If the logged in user is an IU and has user.clientId in payload, then its a preview mode login and should act as a client user instead of an IU.

## Testing Criteria

Preview Mode : 
image

Normal Mode from client : 
image


rrojan (approved):
LGTM 🏆

PR #866, opened 2025-07-03
OUT-1996 | Tasks app issue in Company details view

arpandhakal:
## Changes

- [x] removed the possiblity of users ending up on IU route in Company details view in CRM. This was caused while navigating to a subtask, the task app routed the user to iu due to a slight logic mismatch.
- [x] added block on navigation to a subtask from sub tasks list when the subtasks is still in optimistic update state with a tempId.

## Testing 

- [LOOM](https://www.loom.com/share/39df49d174a54198a4a9f961ab9eb67f)

### DIFF
--- a/src/app/detail/ui/TaskCardList.tsx
+++ b/src/app/detail/ui/TaskCardList.tsx
@@ -47,9 +47,10 @@ interface TaskCardListProps {
   mode: UserRole
   assignee?: IAssigneeCombined
   handleUpdate?: (taskId: string, changes: Partial<TaskResponse>, updater: () => Promise<void>) => Promise<void>
+  isTemp?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


src/app/detail/ui/TaskCardList.tsx
```diff
@@ -164,52 +165,100 @@ export const TaskCardList = ({ task, variant, workflowState, mode, handleUpdate
...
-            width: '100%',
-          }}
-        >
-          <Stack
-            direction="row"
-            sx={{
-              gap: '8px',
+        {isTemp ? (
+          <div
+            key={task.id}
```
rrojan, line 170: Could you please add a loom for these changes

rrojan (approved):
LGTM 🏆

PR #869, opened 2025-07-07
OUT-2002 | Performance optimization on tasks api from TasksDataFetcher

arpandhakal:
## Changes

- [x] removed unnecessary Tasks api fetching in the `TasksDataFetcher` component. Instead of adding refetch mechanism in an effect with archived filters as dependency, SWR refetches data on key changes where key contains the value of archived filters. So, the whole refetching mechanism we implemented was faulty and expensive. 
- [x] Added a condition to fetch subtasks on realtime changes. Only when activeTasks.subTasksCount and subTasks.length are different. Before, it was refetching many times on useless updates. 

## Testing Criteria

- [LOOM](https://www.loom.com/share/8fa7e997730447d9ac205a74f4a35f0d)

### DIFF
--- a/src/app/_fetchers/TaskDataFetcher.tsx
+++ b/src/app/_fetchers/TaskDataFetcher.tsx
@@ -8,7 +8,7 @@ import { useSelector } from 'react-redux'
 import useSWR from 'swr'
 
 export const TaskDataFetcher = ({ token }: PropsWithToken) => {
-  const { showArchived, showUnarchived } = useSelector(selectTaskBoard)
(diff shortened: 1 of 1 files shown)


rrojan (changes requested):
https://github.com/user-attachments/assets/48c1def4-5789-4957-b4d2-1d010885bd43

There is still last remaining duplicate request for archive filters on the IU side. Let's squeeze out every last bit of performance we can!

rrojan (approved):
LGTM :trophy:

PR #876, opened 2025-07-10
OUT-2010 |  Further Optimization on rate limits.

arpandhakal:
## Changes

- [x] removed the revalidation of getOneTask through revalidateTag which was re-rending the server component multiple times on each navigation. This was triggering multiple instance of assignee fetcher causing rate limites.
- [x] implemented client side invalidation of task in details page using SWR. refetched the api to invalidate cache using OneTaskDataFetcher component. This doesnt re-render any components and the assignee api is fetched only once. Before it was fetched 6-7 times on each mount.
- [x] solved hydration error in details page.

## Testing Criteria

- [LOOM]()


rrojan (approved):
LGTM 🏆

PR #877, opened 2025-07-10
OUT-1995: Change activity log for when task is unassigned

SandipBajracharya:
## Changes

- [X] When a task is unassigned, the activity log reads "{unassigner} removed assignee."

## Testing Criteria

image


arpandhakal (commented):
I think removing only the case is not enough here and the whole TASK_UNASSIGNED logic used in the app should be removed because it would only be bloating the app's codebase. 

image

We might be needing the TASK_UNASSIGNED enum and its cases in the future so lets tread on this carefully.

rrojan (approved):

PR #880, opened 2025-07-14
OUT-2020 | Tasks App errors out (autosave feature on viewSettings) 

arpandhakal:
## Changes

- [x] prevent viewsettings to mutate from reducers. Provided  a deep copy of viewsettings instead of actual reference to the reducer because viewsettings has nested object for filterOptions. 

## Testing Criteria

- [LOOM]()


rrojan (approved):
LGTM 🏆

PR #881, opened 2025-07-15
OUT-2022 | Investigate and apply Activity Log Optimization

arpandhakal:
## Changes

- [x] removed the functionality of attaching initiator property, full objects of users on activityLogs (incl comments) response.
- [x] applied relevant changes on the client side, removed initiator from response schema. Looked up initiator from the global assignee state using userId from the logs.
- [x] removed copilotAPI for user fetching for internal users. And only clients and companies are fetched for limited access user and clients for filtering activity logs for limited access.
- [x] returnerd only ids from api instead of whole user objects for firstInitiators in a comment.
- [x] added relevent changes to lookup users objects from global assignee state in the client side.
- [x] added duduping interval of 5 seconds for activity log and subtasks api call from swr to prevent double fetching of these apis on component mount. 

## Testing Criteria

- coming soon...


rrojan (approved):
LGTM :trophy:

PR #882, opened 2025-07-16
OUT-2036 | OUT-2037 | Enforce strict no-duplication mechanism for notification dispatch

rrojan:
## Changes

- [x] Implement no-duplication for Validate Task Count job
- [x] Implement no-duplication for Notification create 
- [x] Implement no-duplication for Notification bulk create

## Testing Criteria

Previously these use cases were causing multiple duplicate notifications to be dispatched for a single task operation

https://github.com/user-attachments/assets/9df154dd-f99f-40fb-9ac1-58453b155d39

https://github.com/user-attachments/assets/07b04a07-4368-4afc-a84f-de15af9283ef

### DIFF
--- a/src/cmd/queries/remove-duplicate-notifications.sql
+++ b/src/cmd/queries/remove-duplicate-notifications.sql
@@ -1,0 +1,11 @@
+DELETE FROM "ClientNotifications" AS cn
+    WHERE cn."deletedAt" IS NULL
+    AND EXISTS (
+        SELECT 1
+        FROM "ClientNotifications" AS sub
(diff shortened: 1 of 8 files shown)


src/cmd/queries/remove-duplicate-notifications.sql
```diff
@@ -0,0 +1,11 @@
+DELETE FROM "ClientNotifications" AS cn
```
SandipBajracharya, line 1: Correct me if I'm wrong but I think this query deletes all the records from `ClientNotifications` table for matching conditions. We only want to delete duplicate records right?

PR #889, opened 2025-07-23
OUT-1985 | OUT-2058 | Support `companyIds` in webhook event Notifications

rrojan:
## Changes

- [x] Implement `companyIds` with new notification create fields in Webhook events
- [x] Optimize bottleneck for higher prod ratelimits 

## Testing Criteria:

- [x] `client.created` - Create client with company

https://github.com/user-attachments/assets/51849d2a-1041-47be-a6cf-6349c89ed808

- [x] `client.updated` - Assign one / multiple companies

NOTE: The copilot multi-companies notifications are very buggy. Sometimes the notification chip disappears entirely for no reason.

https://github.com/user-attachments/assets/7b03a406-17e5-4407-a79b-3b17b4d7ff50

- [x] `client.updated` - unassign companies 

https://github.com/user-attachments/assets/5eefe46e-a6b4-4617-842e-429989585ef0

The count cannot be verified from the UI since access to the company is revoked, if if this company is reassigned, the notification count start again from 0, due to the calls to `markNotificationAsRead` when companies are unassigned.

- [x] `client.deleted`
No changes

### DIFF
--- a/src/app/api/webhook/webhook.service.ts
+++ b/src/app/api/webhook/webhook.service.ts
@@ -1,18 +1,19 @@
-import { ClientUpdatedEventData, HANDLEABLE_EVENT, WebhookEntitySchema, WebhookEvent, WebhookSchema } from '@/types/webhook'
-import { BaseService } from '@api/core/services/base.service'
-import { NextRequest } from 'next/server'
-import APIError from '@api/core/exceptions/api'
-import httpStatus from 'http-status'
-import { AssigneeType, StateType } from '@prisma/client'
-import { CopilotAPI } from '@/utils/CopilotAPI'
-import { TasksService } from '@api/tasks/tasks.service'
-import Bottleneck from 'bottleneck'
-import { getInProductNotificationDetails } from '@api/notification/notification.helpers'
-import { NotificationTaskActions } from '@api/core/types/tasks'
-import { NotificationService } from '@api/notification/notification.service'
-import User from '@api/core/models/User.model'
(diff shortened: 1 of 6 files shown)


src/app/api/webhook/webhook.service.ts
```diff
@@ -56,18 +57,25 @@ class WebhookService extends BaseService {
...
+    const client = await this.copilot.getClient(clientId)
     let company
     try {
       // Accomodate indididual client's company id behavior for copilot
-      company = await this.copilot.getCompany(client.companyId)
+      // Note: When a client is created, only one company can be assigned to it.
+      // This is why we don't need to check for multiple companies here.
+      //
+      // NOTE: forCompanyId is used when a client is created for a company that is not the first company in the client's companyIds array
+      const companyId = forCompanyId || client.companyIds?.[0] || client.companyId
```
arpandhakal, line 70: client.companyIds?.[0] and client.companyId is the same. Furthermore, client.companyId might be deprecated in the future.
```suggestion
      const companyId = forCompanyId || client.companyIds?.[0] 
```

arpandhakal (approved):

src/app/api/webhook/webhook.service.ts
rrojan: @arpandhakal This is for backwards compatibility, which is a good practice to have when making a switch like this.

From left -> right we have the most high priority choice -> low priority one

PR #891, opened 2025-07-24
OUT-2076 | Optimizations to the User's API

rrojan:
## Changes

- [x] Optimize the flow of `getGroupedUsers` function to avoid unnecessary filters
- [x] Slice array only when `limit` is provided (saves us from 3 slices, minor optimization)
- [x] Apply Copilot API's new `isPlaceholder` filter instead of filtering placeholder companies manually

## Testing Criteria:

Response time improvement in my local machine with a production build across 10 requests averaged on OUT-2076 vs `feature/M15`: **13.5% response time improvement**

image

 …ng clients (you can never be too safe)

### DIFF
--- a/src/constants/users.ts
+++ b/src/constants/users.ts
@@ -1,1 +1,1 @@
-export const MAX_FETCH_ASSIGNEE_COUNT = 10_000
+export const MAX_FETCH_ASSIGNEE_COUNT = 15_000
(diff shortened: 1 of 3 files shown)


src/constants/users.ts
```diff
@@ -1 +1 @@
-export const MAX_FETCH_ASSIGNEE_COUNT = 10_000
+export const MAX_FETCH_ASSIGNEE_COUNT = 15_000
```
arpandhakal, line 1: The CopilotAPI doesn't support fetching users with limit even upto 10,000. Is there a reason we are increasing this to 15,000?

PR #903, opened 2025-07-29
OUT-2097 | Lazy load assignee data in chunks for workspaces with a high amount of assignees

arpandhakal:
## Changes

- [x] fetched clients in batches of max 5000 limit until the desired limit is reached. `Max_LIMIT_CLIENT_COUNT` is set to 5000. This is done because copilot API does not support querying for massive amount of clients >~5000. So the basic idea applied in this PR is that, if a workspace has <5000 clients, we only hit the api once, if it has >5000 and <10000, we hit the client apis twice, with the use of `nextToken` and so on...


rrojan (approved):
Works like a charm. Great PR!

Image

PR #908, opened 2025-08-05
OUT-2122 | Add support for CU Mark as Done Button

arpandhakal:
## Changes

- [x] created a ClientDetailsAppBridge component to support the mark as done button in app bridge. This feature is used for Cus only so created a separate component.
- [x] Added relevant conditions to show the button: should be a client user and the task should not be completed yet.

## Testing Criteria

- WIP


src/app/detail/ui/ClientDetailAppBridge.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface DetailAppBridgeProps {
+  handleTaskComplete: () => void
+  isTaskCompleted: boolean
+  portalUrl?: string
+}
+
+export const ClientDetailAppBridge = ({ handleTaskComplete, isTaskCompleted, portalUrl }: DetailAppBridgeProps) => {
+  const [awake, setAwake] = useState(false)
+
+  useEffect(() => {
```
SandipBajracharya, line 20: Awesome solution. Why do you think there is a hydration issue here in the first place?

SandipBajracharya (approved):
LGTM :raised_hands:

PR #909, opened 2025-08-06
OUT-2104 | Filterbar components are very slow on load testing app

arpandhakal:
## Changes

- [x] when there are lots of tasks in the app and tasks state is changed, the app froze for 4-5 seconds.
- [x] applied virtualization for listing tasks on board and list.
- [x] optimized filter by keyword functionality.
- [ ] Some optimization on list view remaining.

## Testing Criteria

- [LOOM]()

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@reduxjs/toolkit": "^2.2.3",
     "@sentry/nextjs": "^8",
     "@supabase/supabase-js": "^2.47.5",
+    "@tanstack/react-virtual": "^3.13.12",
     "@trigger.dev/sdk": "3.3.17",
     "@types/date-fns": "^2.6.0",
     "@types/deep-equal": "^1.0.4",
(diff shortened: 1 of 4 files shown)


rrojan (approved):
Very rarely, a task moved from one workflow state to the next reverts back to the original one for a split second before moving to the intended new workflow state, however this doesn't seem to be an issue with the changes made in the PR.

LGTM, amazing work 🔥

src/app/globals.css
```diff
@@ -40,3 +40,29 @@ a {
...
+.List::-webkit-scrollbar-thumb {
+  background-color: rgba(0, 0, 0, 0.5);
+  border-radius: 20px;
+  margin-right: 1px;
+  transition: background-color 0.15s ease-in-out;
+}
+
+.List::-webkit-scrollbar-thumb:hover {
+  background-color: rgba(0, 0, 0, 0.5);
+}
```
rrojan, line 68: Do we need to do the same for `moz`? Can you check the custom scrollbar on firefox once to verify this?

PR #911, opened 2025-08-08
OUT-2140 | Reassignment is not possible

arpandhakal:
## Changes

- [x] closed the popper and called the update(reassignment) function if some value is selected.
- [x] In case of value being removed(unassigned), prevented calling the update function and closing the popper immediately, instead listened to the close state of popper and then called the update function when popper is closed and no assignee is selected.

## Testing Criteria

- [LOOM](https://www.loom.com/share/8d7db0f8e1b4403fb5e7bc63792c2cdb)


SandipBajracharya (approved):
LGTM 🙌

PR #915, opened 2025-08-13
OUT-2131, OUT-2136, OUT-2134 | Add support for tooltips on editable fields on hover for task cards.

arpandhakal:
## Changes

- [x] created CopilotTooltip component wrapping Tooltip component from copilot-design-system.
- [x] added necessary css to the tooltip component, balancing offset based on children's position.
- [x] applied tooltip to all interactive components on list view and board view.
- [x] reduces padding of list view on mobile from 20px to 12px as per requirement.
- [x] changed focus effect to focus-visible effect so that the styles are active only on keyboard navigation and not on drag-and-drop.

## Testing Criteria

- [LOOM](https://www.loom.com/share/97f5078ae4f448f1ad38983bd0f54002)

### DIFF
--- a/src/components/inputs/CopilotSelector.tsx
+++ b/src/components/inputs/CopilotSelector.tsx
@@ -146,6 +146,9 @@ export const CopilotPopSelector = ({
             width: 'fit-content',
           }}
           placement="bottom-start"
+          onClick={(e) => {
+            e.stopPropagation()
+          }}
         >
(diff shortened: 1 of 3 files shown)


rrojan (changes requested):
This tooltip seems to be hugging the left edge of the viewport.. can we add a little bit of separation between them?

Image

PR #929, opened 2025-08-21
OUT-2203 | Tasks completed and task-assigned notifications do not appear in the notification center to IUs

rrojan:
## Changes

- [x] Add backwards compatibility for workspaces where multi-companies is not enabled. I had no way to check if a workspace had this enabled at all, so i had to go for an error message matching approach.

## Testing Criteria

- [x] Screencast:

https://github.com/user-attachments/assets/5431cf15-e1b2-468b-b279-f4a236d68bed

### DIFF
--- a/src/app/api/notification/notification.service.ts
+++ b/src/app/api/notification/notification.service.ts
@@ -504,6 +504,7 @@ export class NotificationService extends BaseService {
         senderCompanyId: undefined,
       })
     } else if (e instanceof Error) {
+      console.error('Error when handling sender companyId:')
       throw e
     } else {
       console.error(e)


src/app/api/notification/notification.service.ts
```diff
@@ -480,6 +494,23 @@ export class NotificationService extends BaseService {
     return await this.db.clientNotification.findMany({ where: { taskId: { in: taskIds } } })
   }
 
+  private async handleIfSenderCompanyIdError(e: unknown, copilot: CopilotAPI, notificationDetails: NotificationRequestBody) {
+    // Account for workspaces that don't have multi-companies enabled, thus don't support the senderCompanyId key
+    // Yes, this is hacky. No, I don't have a choice (I can't find out if workspace has single/multi company at all from the Copilot API)
+    if (isMessagableError(e) && e.body?.message === 'sender company ID is invalid based on sender') {
```
sajxraj, line 500: this is breakable. I understand its the only way currently but lets work with the Copilot team for proper handling

sajxraj (approved):

src/app/api/notification/notification.service.ts
rrojan: @sajxraj This situation is being caused because they decided to adopt a rolling release where 10% of workspaces are included in this new change while the rest are in the old one. Will be a non-issue by coming Tuesday

PR #944, opened 2025-09-02
OUT-2266: Add support for new "Client visibility" button on Task and Subtask creation form.

SandipBajracharya:
## Changes

- [X] popper button with UserCompanySelector in task and subtask creation form with label "Client visibility"
- [X] the button is visible if the assignee of the task is IU
- [X] send viewers id array while task and subtask creation
- [X] remove client visibility when assignee is changed from IU to client or company
- [X] include new "Person" svg icon

## Testing Criteria

[Loom](https://www.loom.com/share/b27195e926bc468998b1e8633ae12ecc?sid=b07607bf-08ba-4218-9251-8e1dab8be1c4)

- [X] verified that viewers id are stored in DB

image


arpandhakal (approved):
Left a comment on designs. other than that LGTM. nice work. 🏆

PR #951, opened 2025-09-04
OUT-2303 | Tasks App broken in production

arpandhakal:
## Changes

- [x] applied upsert and skipDuplicates flag to prevent constraint violation on creating new viewSettings and WorkflowStates


rrojan (approved):
LGTM 🏆 
Great solution

PR #952, opened 2025-09-04
OUT-2263 | Viewers schema and dto changes

arpandhakal:
## Changes

- [x] schema change and added relevant changes to core api to support viewers in tasks api.


SandipBajracharya (approved):
LGTM 🙌

PR #962, opened 2025-09-10
OUT-2314 | Adding viewer to a client to a parent task where subtasks are accessible to the client is not reflecting in realtime

arpandhakal:
## Changes

- [x] solved an error in handle task insert realtime logic which was causing the realtime handler to throw false on insert event from client user even though the client was added as a viewer.


SandipBajracharya (approved):
LGTM 🙌

PR #963, opened 2025-09-10
OUT-2318 | OUT-2323: revoke permission to change task status by viewer 

SandipBajracharya:
## Changes

- [X] FE implement to disable workflow status dropdown
- [X] fix: allow company as viewer
- [X] minor UI fix for sidebar client visibility
- [X] flag is passed to control viewers filter. We dont want to retrieve task for update when the current user is viewer
- [X] fix: make tasks visible to client when the client's company is provided with visibility

## Testing Criteria

[Loom](https://www.loom.com/share/cad71f05e2de40f8a7b48305c816c709?sid=80013a08-cf75-4cee-986d-88176e9aaf4f)

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -40,7 +40,7 @@ export class TasksService extends BaseService {
    * If user is a client, return filter for just the tasks assigned to this clientId.
    * If user is a client and has a companyId, return filter for just the tasks assigned to this clientId `OR` to this companyId
    */
-  private buildTaskPermissions(id?: string) {
+  private buildTaskPermissions(id?: string, includeViewer: boolean = true) {
     const user = this.user
 
     // Default filters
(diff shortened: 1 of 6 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -532,16 +532,16 @@ export class TasksService extends BaseService {
...
       )
+      if (includeViewer)
+        filters.push(
+          // Get tasks that includes the client as a viewer
+          {
+            viewers: {
+              hasSome: [{ clientId, companyId }, { companyId }],
+            },
+          },
+        )
```
rrojan, line 544: `includeViewer` seems redundant... For clients we always want the viewership query to run.. sort of like:

```sql
select * from "tasks"
where ("clientId" = .. and companyId = ..) OR (viewers <has some [{ clientId, companyId }, { companyId }] > )
```

rrojan (changes requested):

src/app/api/tasks/tasks.service.ts
SandipBajracharya: I'm not sure if I got this. Main object to do this is to implement a safe side or precaution in BE when task update by viewers. Viewers cannot update the workflow status but clients can. I have already disabled the workflow status dropdown for viewers in UI so this scenario might not come.

PR #973, opened 2025-09-16
HOTFIX: validate notification count errors 

arpandhakal:
## Changes

- [x] validate notification count failing through manual fetch api for notifications because of invalid workspaceId passed (undefined)

### DIFF
--- a/src/utils/CopilotAPI.ts
+++ b/src/utils/CopilotAPI.ts
@@ -266,12 +266,16 @@ export class CopilotAPI {
     } = { limit: 100 },
   ) {
     console.info('CopilotAPI#_getClientNotifications', this.token)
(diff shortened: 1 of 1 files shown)


src/utils/CopilotAPI.ts
```diff
@@ -266,12 +266,16 @@ export class CopilotAPI {
...
+      'notifications',
+      {
+        recipientClientId,
+        recipientCompanyId,
+        limit: `${opts.limit}`,
+        workspaceId,
+      },
       workspaceId,
-    })
+    )
```
rrojan, line 278: Shouldn't we be extracting workspaceId from the 3rd argument though

rrojan (changes requested):

PR #975, opened 2025-09-16
OUT-2354 | update public task api with viewers bug

arpandhakal:
## Changes

- [x] fixed tasks update api logic to not reset viewers if internalUser is not provided in the payload and the previous assignee is an IU


SandipBajracharya (approved):
Approved 🙌

PR #1115, opened 2026-01-29
FIX: client portal crashing on beta app when validate count function takes more than 15 seconds. 

arpandhakal:
## Changes

- [x] increased the max duration of client page just in case where validate-notificaiton job takes more than 15 seconds


src/app/client/page.tsx
```diff
@@ -27,6 +27,8 @@ import { UserRole } from '@api/core/types/user'
 import { Suspense } from 'react'
 import { z } from 'zod'
 
+export const maxDuration = 60 //just to be safe. the validate count job might take longer that 15 seconds(default max duration of server components) which will make the app crash. Increasing the duration to 60.
```
SandipBajracharya, line 30: Are we sure this gets resolved in a min? Otherwise we can actually update it to more than a minute if needed.

SandipBajracharya (dismissed):
lgtm

priosshrsth (approved):

PR #1117, opened 2026-02-02
OUT-2998 | Bad comment/task ID returns 400 instead of 404

arpandhakal:
## Changes

- [x] added a validation to check valid uuids for in public controllers. If the uuid is invalid.
- [x] applied short circuit to throw 404 error

## Testing Criteria

image
image


SandipBajracharya (dismissed):
LGTM

priosshrsth (approved):

PR #1123, opened 2026-02-11
OUT-3073: support task association in task detail page

SandipBajracharya:
## Changes

- [x] optimistic UI update and change in task update api
- [x] allow to select association when assignee is IU or no assignee
- [x] share a task with association when assignee is IU
- [x] activity log when a task is shared/stopped sharing with association
- [x] UI implementation

## Testing Criteria

[Loom](https://www.loom.com/share/ac384df7e9fc42edaa551714cfef4bac)

### DIFF
--- a/src/app/ui/NewTaskForm.tsx
+++ b/src/app/ui/NewTaskForm.tsx
@@ -435,6 +435,7 @@ export const NewTaskForm = ({ handleCreate, handleClose }: NewTaskFormProps) => 
                   )
                 }
                 checked={store.getState().createTask.isShared}
+                className="p-1.5 py-2" // px-1.5 is not working
               />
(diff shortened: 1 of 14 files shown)


src/app/ui/NewTaskForm.tsx
```diff
@@ -435,6 +435,7 @@ export const NewTaskForm = ({ handleCreate, handleClose }: NewTaskFormProps) =>
                   )
                 }
                 checked={store.getState().createTask.isShared}
+                className="p-1.5 py-2" // px-1.5 is not working
```
arpandhakal, line 438: will there be any repercussions from this?

arpandhakal (approved):

PR #1130, opened 2026-02-17
fix: await async assemblyApi() factory with lazy init pattern

foleyatwork:
## Summary
- The `@assembly-js/node-sdk` `assemblyApi()` factory is `async` and returns `Promise<AssemblyAPI>`, but was being assigned synchronously in the `CopilotAPI` constructor — causing all SDK method calls to operate on a Promise object instead of the resolved client.
- Uses lazy init pattern: store the promise in the constructor, `await` it in each method. Zero changes to external call sites (`new CopilotAPI(token)` still works as before).
- Renamed internal `this.copilot` references to `assembly` to reflect the company rename.

## Test plan
- [ ] Deploy to a preview environment and verify SDK calls (token payload, client retrieval, notifications, etc.) work correctly
- [ ] Verify no regressions in task creation, comment notifications, and webhook dispatch

### DIFF
--- a/src/utils/CopilotAPI.ts
+++ b/src/utils/CopilotAPI.ts
@@ -42,13 +42,13 @@ import { assemblyApi } from '@assembly-js/node-sdk'
 import { z } from 'zod'
 
 export class CopilotAPI {
-  copilot: SDK
+  private assemblyPromise: Promise<SDK>
 
   constructor(
     private token: string,
     customApiKey?: string,
   ) {
-    this.copilot = assemblyApi({ apiKey: customApiKey ?? apiKey, token })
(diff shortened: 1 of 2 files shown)


foleyatwork:
@rrojan @arpandhakal My apologies, I was so focused on the token logic I completely forgot about this other necessary change to make `assemblyApi` an async function. I published a migration guide here.
https://www.notion.so/copilotplatforms/Migrating-from-copilot-node-sdk-to-assembly-js-node-sdk-30a5ddd81ef4819e82ebdfd6272ef6ff

aschwartz91 (approved):

PR #1134, opened 2026-02-19
OUT-3170 | Changing association to another client/company when the task is shared removes the association all together.

arpandhakal:
## Changes

- [x] Fixed association being reset while changing the association when task is shared. 
- [x] Managed a separate state for selected association value and updated the task with it while chaning association.  

## Testing Criteria

- [LOOM](https://www.loom.com/share/5685ab6a5f7d4fcc90deadea4a830720)


rrojan (approved):
LGTM 🏆

PR #1136, opened 2026-02-20
OUT-3179 | Resetting association from public API on task update endpoint does not work

arpandhakal:
## Changes

- [x] kept the association logic consistent on public task update service. 
- [x] fixed the logic for validating isShared value on task update apis : both web and public. 

## Testing Criteria

- Case 1 : when all value aligns with requirement. 
image

-Case 2 : when association is empty but isShared is on.
image

-Case 3 : when internalUserId is null but isShared is on.
image

-Case 4 : removing association with isShared off. 
image

### DIFF
--- a/src/app/api/tasks/public/public.service.ts
+++ b/src/app/api/tasks/public/public.service.ts
@@ -332,10 +332,25 @@ export class PublicTasksService extends TasksSharedService {
       companyId: validatedIds?.companyId ?? null,
     })
 
-    const associations: Associations = await this.getValidatedAssociations({
-      prevAssociations: prevTask.associations,
-      associationsResetCondition: shouldUpdateUserIds ? !!clientId || !!companyId : !prevTask.internalUserId,
-    })
+    // const associations: Associations = await this.getValidatedAssociations({
+    //   prevAssociations: prevTask.associations,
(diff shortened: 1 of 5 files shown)


src/app/api/tasks/public/public.service.ts
```diff
@@ -332,10 +332,25 @@ export class PublicTasksService extends TasksSharedService {
...
+      : prevTask.clientId || prevTask.companyId
+
+    if (data.associations) {
+      // only update of associations attribute is available. No associations in payload attribute means the data remains as it is in DB.
+      if (associationsResetCondition || !data.associations?.length) {
+        associations = [] // reset associations to [] if task is not reassigned to IU.
+      } else if (data.associations?.length) {
+        associations = await this.validateAssociations(data.associations)
+      }
+    }
```
priosshrsth: @arpandhakal This could be simplified or organized to make it more understandable.
eg create a function that initally checks the condition in line 342. If like 342 is true, it returns empty array.

It then validate associatsions agains schema, after that if result is empty, it again returns empty array. Otherwise. It returns this.validateAssociations(...)

priosshrsth (changes requested):

priosshrsth (approved):
lgtm

PR #1138, opened 2026-02-25
OUT-3202 | Viewers list is not backwards compatible with client association

arpandhakal:
## Changes

- [x] accepted `viewers` in create/update task request from public api alongside `association` and `isShared`.
- [x] `viewers` and `association`/`isShared` is mutually exclusive. which means viewers cannot be used together with association or isShared. Use either viewers alone, or association/isShared together.
- [x] if `viewers` is used in payload for task create/update from public api. the value is transformed i.e `association` becomes `viewers`  and `isShared` becomes true. 

## Testing Criteria

image
Screenshot 2026-02-25 at 18 09 59
image

### DIFF
--- a/src/app/api/tasks/public/public.dto.ts
+++ b/src/app/api/tasks/public/public.dto.ts
@@ -48,6 +48,32 @@ export const PublicTaskDtoSchema = z.object({
 })
 export type PublicTaskDto = z.infer<typeof PublicTaskDtoSchema>
 
+const viewersAssociationExclusivitySchema = z
+  .object({
+    viewers: AssociationsSchema.optional(),
+    association: AssociationsSchema.optional(),
+    isShared: z.boolean().optional(),
+  })
+  .superRefine((val, ctx) => {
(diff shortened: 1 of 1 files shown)


src/app/api/tasks/public/public.dto.ts
```diff
@@ -48,6 +48,32 @@ export const PublicTaskDtoSchema = z.object({
...
+const viewersAssociationExclusivitySchema = z
+  .object({
+    viewers: AssociationsSchema.optional(),
+    association: AssociationsSchema.optional(),
+    isShared: z.boolean().optional(),
+  })
+  .superRefine((val, ctx) => {
+    const hasViewers = val.viewers !== undefined
+    const hasAssociation = val.association !== undefined
+    const hasIsShared = val.isShared !== undefined
```
rrojan, line 60: Since viewers and associations are objects we can simply do `!val.viewers` / `!val.association`, etc but this should be fine as well

rrojan (approved):
LGTM 🏆

PR #1141, opened 2026-02-26
OUT-3208: remove 'shared with' filter option

SandipBajracharya:
## Changes

- [x] remove secondary filters related to shared with

## Testing Criteria

[Loom](https://www.loom.com/share/21e921593eb44761a592d3815043c50c)

### DIFF
--- a/src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
+++ b/src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
@@ -20,7 +20,6 @@ export const filterOptionsMap = {
   [FilterType.Assignee]: FilterOptions.ASSIGNEE,
   [FilterType.Creator]: FilterOptions.CREATOR,
   [FilterType.Association]: FilterOptions.ASSOCIATION,
-  [FilterType.IsShared]: FilterOptions.IS_SHARED,
 }
 
 export const FilterAssigneeSection = ({ filterMode, setAnchorEl }: FilterAssigneeSectionProps) => {
@@ -36,7 +35,7 @@ export const FilterAssigneeSection = ({ filterMode, setAnchorEl }: FilterAssigne
   const hideClientsAndCompanies =
     filterMode === FilterType.Creator || (type === FilterOptionsKeywords.TEAM && filterMode === FilterType.Assignee)
   const hideIus =
-    [FilterType.Association, FilterType.IsShared].includes(filterMode) ||
+    [FilterType.Association].includes(filterMode) ||
     (type === FilterOptionsKeywords.CLIENTS && filterMode === FilterType.Assignee)
 
   const handleChange = (inputValue: InputValue[]) => {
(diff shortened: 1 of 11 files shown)


src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
```diff
@@ -36,7 +35,7 @@ export const FilterAssigneeSection = ({ filterMode, setAnchorEl }: FilterAssigne
   const hideClientsAndCompanies =
     filterMode === FilterType.Creator || (type === FilterOptionsKeywords.TEAM && filterMode === FilterType.Assignee)
   const hideIus =
-    [FilterType.Association, FilterType.IsShared].includes(filterMode) ||
+    [FilterType.Association].includes(filterMode) ||
```
priosshrsth: Better to do `===` check?

src/components/inputs/FilterSelector/FilterTypeSection.tsx
```diff
@@ -16,7 +16,7 @@ export const FilterTypeSection = ({ setFilterMode, filterModes }: FilterTypeSect
     filterOptions: { type },
   } = useSelector(selectTaskBoard)
 
-  const disabled = type === FilterOptionsKeywords.CLIENTS || FilterOptionsKeywords.UNASSIGNED ? [FilterType.IsShared] : []
+  const disabled = type === FilterOptionsKeywords.CLIENTS ? [FilterType.Association] : []
```
priosshrsth: I'd always recommend set if you have to check presence in an array especially inside another loop.

priosshrsth (changes requested):

src/components/inputs/FilterSelector/FilterTypeSection.tsx
```diff
@@ -16,8 +16,8 @@ export const FilterTypeSection = ({ setFilterMode, filterModes }: FilterTypeSect
     filterOptions: { type },
   } = useSelector(selectTaskBoard)
 
-  const disabled = type === FilterOptionsKeywords.CLIENTS || FilterOptionsKeywords.UNASSIGNED ? [FilterType.IsShared] : []
-  const removed = type.length > 20 ? [FilterType.Assignee] : []
+  const disabled = type === FilterOptionsKeywords.CLIENTS && new Set([FilterType.Association])
+  const removed = type.length > 20 && new Set([FilterType.Assignee])
```
priosshrsth:
```ts
  const filtersToDisable = new Set<typeof FilterType>();
  if (type === FilterOptionsKeywords.CLIENTS) {
   filtersToDisable.add(FilterType.Association);
   }
   
  const filtersToRemove = new Set<typeof FilterType>();
   if (type.length > 20) {
   filtersToRemove.add(FilterType.Assignee);
   }
```
Something like this? Makes it more readable imo. But I might be biased.

priosshrsth (changes requested):

priosshrsth (approved):

PR #1148, opened 2026-03-02
feat: add token refresh via @assembly-js/app-bridge

foleyatwork:
**Kevin's Notes:**
I forgot to add the app bridge last time, hopefully this solves any expiring token issues.

Eventually you guys can replace all your postMessages with function calls from this new package.

---

**AI Description:**

## Summary
- Adds `@assembly-js/app-bridge` (v1.1.0) to subscribe to automatic token refresh events from the parent Copilot dashboard
- Creates `useTokenRefresh` hook that configures the bridge with the portal URL and dispatches updated tokens to Redux
- Wires the hook into `ClientSideStateUpdate` so it runs on every page

## Test plan
- [x] `npx tsc --noEmit` passes with no type errors
- [x] In browser dev tools, verify the `postMessage` listener is active after page load
- [x] When the parent dashboard sends a `sessionToken.update` message, confirm Redux state `taskBoard.token` updates

### DIFF
--- a/src/hooks/app-bridge/useTokenRefresh.ts
+++ b/src/hooks/app-bridge/useTokenRefresh.ts
@@ -19,6 +19,7 @@ export function useTokenRefresh(portalUrl?: string) {
 
   useEffect(() => {
     const unsubscribe = AssemblyBridge.sessionToken.onTokenUpdate((data) => {
+      console.info('#onTokenUpdate active', data)
       store.dispatch(setToken(data.token))
     })
     return unsubscribe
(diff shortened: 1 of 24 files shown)


src/hooks/app-bridge/useTokenRefresh.ts
```diff
@@ -0,0 +1,26 @@
...
+  useEffect(() => {
+    if (portalUrl) {
+      AssemblyBridge.configure({ additionalOrigins: [ensureHttps(portalUrl)] })
+    }
+  }, [portalUrl])
+
+  useEffect(() => {
+    const unsubscribe = AssemblyBridge.sessionToken.onTokenUpdate((data) => {
+      store.dispatch(setToken(data.token))
+    })
```
rrojan, line 24: @foleyatwork Assuming a long user session, does onTokenUpdate fire events for multiple token refreshes?

foleyatwork: Yes, I think it's a pretty self-explanatory implementation, any time the frame receives a new token that callback is fired.

PR #1150, opened 2026-03-09
fix(OUT-3297): subtasks inherit parent assignee from template

aschwartz91:
## Summary
- Removed `previewMode` conditional in `createSubtasksFromTemplate` so assignee fields (`internalUserId`, `clientId`, `companyId`, `associations`) are always passed to subtasks created from templates
- Previously subtasks only inherited the parent assignee when creating from CRM preview mode; now they inherit it from the main app as well

## Test plan
- [x] As an internal user, create a task from a template with subtasks and assign the parent — verify subtasks inherit the assignee
- [x] As a client/company user (CRM preview), create a task from a template — verify subtasks still inherit the assignee as before
- [x] Create a task from a template without assigning anyone — verify subtasks are also unassigned

https://github.com/user-attachments/assets/d71672c9-d0c5-4b10-a7d9-d830868b73a0

Linear: https://linear.app/assemblycom/issue/OUT-3297


aschwartz91:
@greptileai review

PR #1151, opened 2026-03-24
OUT-3439: support dynamic fields and its conversion on task template title

SandipBajracharya:
## Changes

- [x] Typing { in the template title opens a dropdown with 4 dynamic field options: Current Week, Current Month, Current Quarter, Current Year
- [x] Selecting a field inserts {{Field Name}} token into the title at cursor position
- [x] Dynamic field tokens display with a light gray border (#D0D4DA), background (#F5F5F5), and text color (#6B6F76) — visually distinct from normal text
- [x] Tokens behave as atomic units — backspace deletes the entire token, not individual characters
- [x] Dismissing the dropdown (click outside / Escape) removes the trigger { character from the field
- [x] Typing after { filters the dropdown options; closes if no match
- [x] Dynamic fields work in both template editing (/manage-templates) and template creation modal
- [x] When creating a task from a template, tokens resolve to current values: {{Current Month}} → "March", {{Current Quarter}} → "Q1", {{Current Week}} → "Week of March 24", {{Current Year}} → "2026"
- [x] Resolution applies to both task creation and subtask creation from templates 

#### What cursor position does in this PR?                                                                                                                                                                                
Cursor position tracking (getCursorOffset / restoreCursorOffset) solves a fundamental problem with contentEditable + React:                                                                                  

The problem: When we update innerHTML (e.g., after token insertion or brace normalization), the browser resets the cursor to the beginning. The user would lose their place every time.

### DIFF
--- a/src/app/manage-templates/ui/TemplateDetails.tsx
+++ b/src/app/manage-templates/ui/TemplateDetails.tsx
@@ -2,7 +2,7 @@
 
 import { StyledModal } from '@/app/detail/ui/styledComponent'
 import AttachmentLayout from '@/components/AttachmentLayout'
-import { StyledTextField } from '@/components/inputs/TextField'
+import { TokenizedInput, restoreCursorOffset } from '@/components/inputs/TokenizedInput'
 import { ConfirmDeleteUI } from '@/components/layouts/ConfirmDeleteUI'
 import { MAX_UPLOAD_LIMIT } from '@/constants/attachments'
 import { useDebounce, useDebounceWithCancel } from '@/hooks/useDebounce'
@@ -42,6 +42,7 @@ export default function TemplateDetails({
   const { activeTemplate, targetTemplateId, taskName } = useSelector(selectCreateTemplate)
   const [isUserTyping, setIsUserTyping] = useState(false)
   const [activeUploads, setActiveUploads] = useState(0)
+  const titleRef = useRef<HTMLDivElement>(null)
 
   const { showConfirmDeleteModal } = useSelector(selectTaskDetails)
 
@@ -73,8 +74,7 @@ export default function TemplateDetails({
   const [debouncedResetTypingFlag, _cancelDebouncedResetTypingFlag] = useDebounceWithCancel(resetTypingFlag, 1500)
   const [debouncedResetTypingFlagTitle, cancelDebouncedResetTypingFlagTitle] = useDebounceWithCancel(resetTypingFlag, 2500)
 
-  const handleTitleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const newTitle = e.target.value
+  const handleTitleChange = (newTitle: string) => {
     setUpdateTitle(newTitle)
     if (newTitle.trim() == '') {
       cancelTitleUpdateDebounced()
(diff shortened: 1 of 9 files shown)


src/app/manage-templates/ui/TemplateDetails.tsx
priosshrsth, line 62: @SandipBajracharya You have to address missing dependency warning here: activeTemplate

src/components/inputs/TokenizedInput.tsx
```diff
@@ -0,0 +1,227 @@
...
+      const range = document.createRange()
+      range.setStart(pos.node, pos.offset)
+      range.collapse(true)
+      sel.removeAllRanges()
+      sel.addRange(range)
+    }
+  }
+}
+
+export const TokenizedInput = forwardRef<HTMLDivElement, TokenizedInputProps>(
```
priosshrsth: Let's not use forwardRef. Now react can accept ref as normal props.

priosshrsth (changes requested):

src/utils/dynamicFields.ts
```diff
@@ -0,0 +1,82 @@
...
+    return resolveDynamicField(key, now)
+  })
+}
+
+function escapeHtml(text: string): string {
+  return text.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;')
+}
+
+const TOKEN_SPAN_STYLE =
+  'border:1px solid #D0D4DA;color:#6B6F76;background:#F5F5F5;border-radius:4px;padding:0 4px;white-space:nowrap;'
```
priosshrsth: You can just use this style by using data-token selector in css instead. Or you can make it more unique or add class as well.

src/utils/dynamicFields.ts
```diff
@@ -0,0 +1,82 @@
...
+  { key: 'Current Quarter', label: 'Current Quarter' },
+  { key: 'Current Year', label: 'Current Year' },
+]
+
+function getQuarter(date: dayjs.Dayjs): string {
+  const month = date.month() // 0-indexed
+  return `Q${Math.floor(month / 3) + 1}`
+}
+
+export function resolveDynamicField(key: string, now: dayjs.Dayjs = dayjs()): string {
```
priosshrsth: Could key be enum?

src/hooks/useDynamicFieldTrigger.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  const handleChange = useCallback(
+    (newValue: string) => {
+      const cursorPos = getCursorPos()
+
+      if (!open) {
+        if (cursorPos > 0 && newValue[cursorPos - 1] === '{') {
+          triggerCursorPos.current = cursorPos - 1
+          setFilterText('')
+          setOpen(true)
```
priosshrsth, line 96: Wondering if blurring and focusing would be easier. I am assuming you are trying to set the cursor position to end of the closing of curly braces.

priosshrsth (changes requested):

src/hooks/useDynamicFieldTrigger.ts
SandipBajracharya: Cursor position is mainly used here for contentEditable used in textfield. When we update innerHTML (e.g., after tokenToHtml()), the browser resets the cursor to the beginning. The user would lose their place every time.

priosshrsth (approved):
lgtm

PR #1166, opened 2026-03-27
Fix/firefox text cursor

arpandhakal:


priosshrsth (approved):
Nice one. Good to know.

PR #1167, opened 2026-03-27
OUT-3478: support dynamic fields in sub-template creation and display

SandipBajracharya:
## Changes

- Replace plain text input with TitleEditor in NewTemplateCard so sub-template titles support dynamic field tokens ({{Current Month}}, etc.)
- Resolve dynamic field tokens in subtask titles and body when tasks are created from templates via the internal and public API (tasksShared.service.ts)
- Add DynamicFieldTitle component to render {{token}} patterns as styled inline chips in the sub-templates list on the template detail page

## Testing Criteria

[Loom UI](https://www.loom.com/share/fc11bf2eeb1849c5a6d5fe785f657bbf)

[Loom public api)(https://www.loom.com/share/7e7cec47b3a04cb1bb54ef6c2edb0a9c)

### DIFF
--- a/src/components/atoms/DynamicFieldTitle.tsx
+++ b/src/components/atoms/DynamicFieldTitle.tsx
@@ -1,0 +1,98 @@
+import { DYNAMIC_FIELDS } from '@/utils/dynamicFields'
+import { Typography, Box, SxProps, Theme } from '@mui/material'
+import React from 'react'
+
+type DynamicFieldTitleVariant = 'subtasks' | 'card'
+
+const variantStyles: Record<DynamicFieldTitleVariant, { typographyVariant: 'bodySm' | 'bodyMd'; sx: SxProps<Theme> }> = {
+  subtasks: {
+    typographyVariant: 'bodySm',
+    sx: {
+      lineHeight: '21px',
+      fontSize: '13px',
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+      whiteSpace: 'nowrap',
+      flexShrink: 1,
+      flexGrow: 0,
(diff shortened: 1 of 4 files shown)


src/components/atoms/DynamicFieldTitle.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function parseDynamicFieldTokens(text: string): { text: string; isDynamic: boolean }[] {
+  const parts: { text: string; isDynamic: boolean }[] = []
+  const regex = /\{\{([^}]+)\}\}/g
+  let lastIndex = 0
+  let match
+
+  while ((match = regex.exec(text)) !== null) {
+    if (match.index > lastIndex) {
```
priosshrsth, line 78: @SandipBajracharya I think you already have create a functionality to convert string with curly brackets to html string. Can we not use that?

priosshrsth (changes requested):

src/components/atoms/DynamicFieldTitle.tsx
SandipBajracharya: The use-case of this function is to return the title string in formatted form:

Example:
for title: Report {{Current Month}} summary

Output:
 [                                                                                                                                                                                                                                                                            
    { text: 'Report ', isDynamic: false },
    { text: 'Current Month', isDynamic: true },                                                                                                                                                                                                                                
    { text: ' summary', isDynamic: false },                                                                                                                                                                                                                                    
  ]

This way we get to know which is dynamic field

priosshrsth (approved):

PR #1170, opened 2026-03-30
OUT-3480: prevent layout shift when inserting dynamic field in template title

SandipBajracharya:
## Summary

- The autofill chip renders inside an inline-flex node view. Its border (1px top + 1px bottom) causes the flex item's box to exceed the paragraph line-height, expanding the line box by 2px on every chip insertion.

- Replace the chip's border with an outline scoped to the title editor context. Outline is cosmetically identical (same color, follows border-radius in modern browsers) but has no effect on layout, eliminating the shift without changing the chip's size or padding.

## Testing Criteria

[Loom](https://www.loom.com/share/917b654d357c497daaab95accc048ca6)

### DIFF
--- a/src/components/inputs/tiptap/TitleEditor.tsx
+++ b/src/components/inputs/tiptap/TitleEditor.tsx
@@ -57,4 +57,12 @@ const StyledEditorWrapper = styled(Box)(({ theme }) => ({
     float: 'left',
     height: 0,
   },
+  // The chip border (1px top + 1px bottom) makes the inline-flex node view taller than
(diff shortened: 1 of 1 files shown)


priosshrsth (approved):
I wonder if setting box-sizing to border-box or something would have solved the issue. But if it solves the issue, I am not against it.

SandipBajracharya:
> I wonder if setting box-sizing to border-box or something would have solved the issue. But if it solves the issue, I am not against it.

`border-box` requires explicit height. In our case we should have height = line-height. Since line height is dynamic (sent as props in component), the current approach would be simpler

PR #1171, opened 2026-03-30
OUT-2668, OUT-3181 | Show subtasks IU is assigned to in "My Tasks"

arpandhakal:
## Summary
- Subtasks now appear as standalone items in filtered views (e.g. "My Tasks") when they match the active filter but their parent task does not
- Refactored `useFilter` hook to reuse filter logic across both parent tasks and subtasks via `applyFilters`/`applyOneFilter` helpers
- Added `accessibleTasks` to the `useEffect` dependency array so realtime subtask changes trigger re-filtering
- Rendered subtasks on task board according to its archived, unarchived states. 

## Test plan
- [x] Apply "My Tasks" filter → verify subtasks assigned to you appear standalone when their parent is assigned to someone else
- [x] Apply other filters (assignee, creator, association, type, keyword) → verify subtasks are promoted to standalone when matching
- [x] Verify parent tasks with matching subtasks still show those subtasks nested (not duplicated as standalone)
- [x] Verify realtime: another user assigns a subtask to you → it appears in your "My Tasks" view without refresh
- [x] Verify no filter active → board behaves as before (no standalone subtasks injected)

## Testing Criteria 

- [LOOM](https://www.loom.com/share/d3b9a501c7df43789599546c7cdfb388?focus_title=1&muted=1&from_recorder=1)

- OUT-3181 : [LOOM](https://www.loom.com/share/bdee8a617a0b45d7ae6c08b38ed1405f)

### DIFF
--- a/src/hooks/useFilter.tsx
+++ b/src/hooks/useFilter.tsx
@@ -149,31 +149,49 @@ export const useFilter = (filterOptions: IFilterOptions, isPreviewMode: boolean)
   const { tasks, accessibleTasks, assignee } = useSelector(selectTaskBoard)
   const [_, startTransition] = useTransition()
 
-  function applyFilter(tasks: TaskResponse[], filterOptions: IFilterOptions) {
+  function applyFilters(tasks: TaskResponse[], filterOptions: IFilterOptions) {
     let filteredTasks = [...tasks]
     for (const [filterType, filterValue] of Object.entries(filterOptions)) {
       if (!filterValue) continue
-      if (filterType === FilterOptions.ASSIGNEE && !isPreviewMode) {
-        // there is no filter by assignee in preview mode
-        const assigneeFilterValue = UserIdsSchema.parse(filterValue)
-        filteredTasks = FilterFunctions[FilterOptions.ASSIGNEE](filteredTasks, assigneeFilterValue)
-      }
-      if (filterType === FilterOptions.CREATOR || filterType === FilterOptions.ASSOCIATION) {
-        const assigneeFilterValue = UserIdsSchema.parse(filterValue)
-        filteredTasks = FilterFunctions[filterType](filteredTasks, assigneeFilterValue)
-      }
-      if (filterType === FilterOptions.KEYWORD) {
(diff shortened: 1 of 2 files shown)


src/hooks/useFilter.tsx
```diff
@@ -149,39 +149,57 @@ export const useFilter = (filterOptions: IFilterOptions, isPreviewMode: boolean)
...
     }
+    return tasks
+  }
+
+  function applyFilter(tasks: TaskResponse[], filterOptions: IFilterOptions) {
+    const filteredParentTasks = applyFilters(tasks, filterOptions)
+    const filteredParentIds = new Set(filteredParentTasks.map((t) => t.id))
+
+    // Find subtasks that match all filters but whose parent didn't
+    const hasActiveFilter = Object.values(filterOptions).some((v) => !!v)
```
priosshrsth, line 184: @arpandhakal Why do we need this check only for sub tasks?

src/hooks/useFilter.tsx
```diff
@@ -149,39 +149,57 @@ export const useFilter = (filterOptions: IFilterOptions, isPreviewMode: boolean)
...
+    let standaloneSubtasks: TaskResponse[] = []
+
+    if (hasActiveFilter) {
+      const subtasks = accessibleTasks.filter((t) => !!t.parentId)
+      const matchingSubtasks = applyFilters(subtasks, filterOptions)
+      standaloneSubtasks = matchingSubtasks.filter((t) => !filteredParentIds.has(t.parentId!))
+    }
+
+    const filteredTasks = [...filteredParentTasks, ...standaloneSubtasks]
```
priosshrsth, line 193: Her's what I understand:
1. `filteredParentTasks `only holds the parent tasks that matches our filter.
2. `standAloneSubTasks` only holds the sub tasks that matches our filter.

So what happens if Parent Task 1 is not in `filteredParentTasks`, but one of it's `subtask` is in `standAloneSubTasks`?

priosshrsth (changes requested):

priosshrsth (approved):
lgtm

PR #1174, opened 2026-04-02
OUT-3501: fix deleted parent comment avatar size mismatch

SandipBajracharya:
## Changes

The CopilotAvatar icon branch rendered a default MUI Avatar (40x40px) while normal comment avatars use AssemblyAvatar xs size (20x20px). This caused deleted parent comments to appear out of place. Map the size prop to correct pixel dimensions for the icon avatar variant.

## Testing Criteria

image


arpandhakal (approved):
LGTM

PR #1177, opened 2026-04-15
OUT-3593 | Forward metadata headers when creating task audit log events

arpandhakal:
## Changes

- [x] Captured and forwarded the header metadatas on task related webhook event. 

## Testing criteria 

Headers on Webhook on 'Task Created' from Production app: 
image

Headers on Webhook on 'Task Created' from assembly api: 
image

### DIFF
--- a/src/utils/CopilotAPI.ts
+++ b/src/utils/CopilotAPI.ts
@@ -3,6 +3,7 @@ import { withRetry } from '@/app/api/core/utils/withRetry'
 import { copilotAPIKey as apiKey, APP_ID, assemblyApiDomain } from '@/config'
 import { MAX_LIMIT_CLIENT_COUNT } from '@/constants/users'
 import {
+  AssemblyMetadata,
   ClientRequest,
   ClientResponse,
   ClientResponseSchema,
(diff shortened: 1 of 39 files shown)


src/utils/CopilotAPI.ts
```diff
@@ -311,18 +312,33 @@ export class CopilotAPI {
...
     console.info('CopilotAPI#dispatchWebhook | Dispatching webhook to ', url, 'with payload', payload ?? null)
 
+    const headers: Record<string, string> = {
+      'Content-Type': 'application/json',
+      'x-api-key': `${workspaceId}/${apiKey}`,
+    }
+
+    if (assemblyMetadata?.source) headers['X-Assembly-Source'] = assemblyMetadata.source
+    if (assemblyMetadata?.clientIp) headers['X-Assembly-Client-IP'] = assemblyMetadata.clientIp
+    if (assemblyMetadata?.userAgent) headers['X-Assembly-User-Agent'] = assemblyMetadata.userAgent
```
priosshrsth, line 334: @arpandhakal So this is not correct it seems. Looks like we were confused. So for proxy we can expect these headers. But when dispatching events, we are supposed to send it as 'x-real-ip' and 'user-agent' header

priosshrsth (changes requested):

priosshrsth (dismissed):

SandipBajracharya (approved):
lgtm

PR #1180, opened 2026-04-16
OUT-3491 | Tasks App loading state in notification center is bad

arpandhakal:
## Changes

- [x] There was a delay on rendering sidebar skeleton on notification center redirection because we we're fetching the `fromNotificationCenter` variable from redux which caused the delay. To fix this, directly pass the value of `fromNotificationsCenter` from server component which doesn't cause any delays. 

## Testing Criteria

- [LOOM](https://www.loom.com/share/22ad1c37578c4dbfa8b6e4debf153a2b)

### DIFF
--- a/src/app/detail/ui/Sidebar.tsx
+++ b/src/app/detail/ui/Sidebar.tsx
@@ -71,6 +71,7 @@ export const Sidebar = ({
   workflowDisabled = false,
   userType,
   portalUrl,
+  fromNotificationCenter: fromNotificationCenterProp,
 }: {
   task_id: string
   selectedWorkflowState: WorkflowStateResponse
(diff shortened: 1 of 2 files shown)


src/app/detail/ui/Sidebar.tsx
```diff
@@ -761,14 +763,20 @@ export const SidebarSkeleton = () => {
...
         direction="row"
         columnGap={'8px'}
         rowGap={'8px'}
         position="relative"
-        sx={{ flexWrap: 'wrap', padding: '12px 18px' }}
+        sx={{
+          flexWrap: 'wrap',
+          padding: '12px 18px',
+          maxWidth: fromNotificationCenter ? '654px' : undefined,
+          width: fromNotificationCenter ? '654px' : 'auto',
```
priosshrsth, line 777: @arpandhakal Is this correct? what happens in smaller screen sizes? maybe width need to be minmax of auto and 654px?

priosshrsth (changes requested):

priosshrsth (approved):

PR #1186, opened 2026-04-16
fix(OUT-3450): auto-focus assignee search + fix placeholder size

aschwartz91:
## Summary

Targeting two issues in the Assignee/Creator/Association filter selector:

1. **Having to click twice when searching for an assignee** — when the user clicks **Filters → Assignee** (or Creator/Association), the search input now auto-focuses so they can start typing immediately. Uses a ref + `input.focus({ preventScroll: true })` instead of react-select's `autoFocus` prop, because the latter scrolls the popover into view on mount.
2. **Placeholder text size** — the search placeholder and input were rendering at 14px. Per the Figma design-system spec they should be **13px / line-height 20px**. Both are updated.

Also tidied up a couple of related visual details while the area was open:
- Lightened the placeholder color to `#9B9FA3` so it matches the lighter gray used by the top-bar search placeholder.
- Removed the blue `focus-within` / hover border on the control; the border stays `#EFF1F4` regardless of state.

The remaining design-system work (unifying the search box + options into a single seamless container, and making 13px the default in the DS) is tracked separately in **POR-19462**.

## Test plan
- [x] Open Filters → Assignee: search input is focused immediately (can type without clicking), and the page does not scroll
- [x] Same for Filters → Creator and Filters → Association
- [x] Placeholder renders at 13px and uses the lighter gray color
- [x] Selecting an assignee/creator/association still works as before and applies the filter

**Before**
image

**After**
image

### DIFF
--- a/src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
+++ b/src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
@@ -10,11 +10,12 @@ import { getSelectedUserIds } from '@/utils/selector'
 import { Box } from '@mui/material'
 import { Dispatch, SetStateAction, useEffect, useRef } from 'react'
 import { useSelector } from 'react-redux'
-import { userCompanySelectorStyles } from '../UserCompanySelectorStyles'
+import { userCompanySelectorStyles } from '@/components/inputs/UserCompanySelectorStyles'
 
 interface FilterAssigneeSectionProps {
   filterMode: FilterType
   setAnchorEl: Dispatch<SetStateAction<HTMLElement | null>>
+  autoFocus?: boolean
 }
 
 export const filterOptionsMap = {
@@ -23,7 +24,7 @@ export const filterOptionsMap = {
   [FilterType.Association]: FilterOptions.ASSOCIATION,
 }
 
-export const FilterAssigneeSection = ({ filterMode, setAnchorEl }: FilterAssigneeSectionProps) => {
+export const FilterAssigneeSection = ({ filterMode, setAnchorEl, autoFocus = false }: FilterAssigneeSectionProps) => {
   const {
     assignee: assignees,
     filterOptions: { type },
@@ -46,9 +47,10 @@ export const FilterAssigneeSection = ({ filterMode, setAnchorEl }: FilterAssigne
 
   const containerRef = useRef<HTMLDivElement | null>(null)
   useEffect(() => {
+    if (!autoFocus) return
     const input = containerRef.current?.querySelector<HTMLInputElement>('input')
     input?.focus({ preventScroll: true })
-  }, [])
+  }, [autoFocus])
 
   return (
     <Box ref={containerRef}>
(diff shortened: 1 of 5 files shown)


src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
```diff
@@ -8,8 +8,9 @@ import { parseAssigneeToSelectorOption } from '@/utils/addTypeToAssignee'
 import { getWorkspaceLabels } from '@/utils/getWorkspaceLabels'
 import { getSelectedUserIds } from '@/utils/selector'
 import { Box } from '@mui/material'
-import { Dispatch, SetStateAction } from 'react'
+import { Dispatch, SetStateAction, useEffect, useRef } from 'react'
 import { useSelector } from 'react-redux'
+import { userCompanySelectorStyles } from '../UserCompanySelectorStyles'
```
priosshrsth: Can we change the relative import to use path alias?

src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
```diff
@@ -43,8 +44,14 @@ export const FilterAssigneeSection = ({ filterMode, setAnchorEl }: FilterAssigne
     setAnchorEl(null)
   }
 
+  const containerRef = useRef<HTMLDivElement | null>(null)
+  useEffect(() => {
+    const input = containerRef.current?.querySelector<HTMLInputElement>('input')
+    input?.focus({ preventScroll: true })
+  }, [])
```
priosshrsth: Since this runs on mount, will this cause regression, if this is component is used inline anywhere and not as a popover? cc @arpan.dhakal

priosshrsth (changes requested):
Minor nitpick and awaiting Arpan's feedback on one possible regression?

src/components/inputs/UserCompanySelectorStyles.ts
```diff
@@ -0,0 +1,48 @@
...
+export const userCompanySelectorStyles: UserCompanySelectorStyles = {
+  control: (base) => ({
+    ...base,
+    borderColor: '#EFF1F4',
+    backgroundColor: '#FFFFFF',
+    boxShadow: 'none',
+    '&:focus-within': {
+      borderColor: '#EFF1F4',
+    },
+  }),
```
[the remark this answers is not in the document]
aschwartz91: Good catch — added an explicit `'&:hover'` override in d4d2f0c7 so the border stays `#EFF1F4` regardless of hover state.

src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
aschwartz91: Done in d4d2f0c7 — switched to `@/components/inputs/UserCompanySelectorStyles`.

src/components/inputs/FilterSelector/FilterAssigneeSection.tsx
aschwartz91: Fair concern. `FilterAssigneeSection` is only rendered inside the `Popper` in `FilterSelector/index.tsx` today, so there's no current regression, but to future-proof I made the auto-focus opt-in via an `autoFocus` prop (default `false`) in d4d2f0c7. The popper caller passes `autoFocus`, so behavior is unchanged for the current use — any future inline usage won't steal focus on mount unless it explicitly opts in.

aschwartz91:
@greptileai PTAL

PR #1189, opened 2026-04-21
fix(OUT-3492): flush pending title/description saves during fast interactions

arpandhakal:
## Summary

- Debounced title (1.5s) and description (0.5s) updates on the task detail page and template editor were being lost when the user navigated away or closed the tab before the debounce timer fired. Root cause: `useDebounce`'s unmount cleanup called `clearTimeout`, which cancels the pending save instead of running it.
- `useDebounce` / `useDebounceWithCancel` now flush pending calls on unmount. `useDebounceWithCancel` additionally exposes a `flush()` as a third tuple slot (backward-compatible with existing positional destructures).
- `TaskEditor` and `TemplateDetails` flush on title blur, description blur, `beforeunload`, and when the `task_id` / `template_id` changes — so the previous task's pending save fires against its own callback before new props settle.

## Why this is safe across task/template switches

Pending args **and** the callback that scheduled them are captured at call-time (not via a live ref that follows every render). If the parent re-renders with a new save function mid-debounce, the flushed timer still fires the callback that was live when the user typed — no cross-task contamination.

Linear: https://linear.app/assemblycom/issue/OUT-3492/fix-title-saving-during-fast-interactions

## Test plan

- [x] Task detail page: type a new title rapidly, click the back button within 1.5s, reload — title persists.
- [x] Task detail page: type a new description rapidly, navigate within 0.5s, reload — description persists (this is the primary loss vector today).
- [x] Task detail page: type, then close the tab within the debounce window, reopen — persists (best-effort; subject to browser `beforeunload` quirks).
- [x] Task detail page: type, then click/tab out of the title input — save fires immediately (no 1.5s wait).
- [x] Task detail page: same blur check on the description.
- [x] Task templates (`/manage-templates`): repeat all of the above.
- [x] Switch between two tasks rapidly, each with unsaved edits — edits from Task A must not overwrite Task B.
- [x] Type a title, clear it to empty, blur — no empty-string save is sent (existing cancel-on-empty path still works).

## TEST

- [LOOM](https://www.loom.com/share/5a6476db84344798aed541d743924f44), also tested for template details. 

### DIFF
--- a/src/hooks/useDebounce.ts
+++ b/src/hooks/useDebounce.ts
@@ -1,4 +1,4 @@
-import { useRef, useEffect } from 'react'
+import { useRef, useEffect, useCallback } from 'react'
 
 type Timer = ReturnType<typeof setTimeout>
 type SomeFunction = (...args: any[]) => void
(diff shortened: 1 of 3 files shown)


src/hooks/useDebounce.ts
```diff
@@ -11,47 +11,85 @@ type SomeFunction = (...args: any[]) => void
...
+      const fn = pendingFunc.current
+      const a = pendingArgs.current
+      pendingFunc.current = null
+      pendingArgs.current = null
+      if (fn && a) fn(...a)
+    }, delay)
+  }) as Func
 
-  return [debouncedFunction, cancel] as const
+  return [debouncedFunction, cancel, flush] as const
```
priosshrsth: Will it be a good idea to return object instead?

priosshrsth (changes requested):

priosshrsth (approved):
lgtm

PR #1190, opened 2026-04-21
OUT-3495 | Inherit parent fields for new subtasks

arpandhakal:
## Summary

When creating a subtask from the task details page, the form now pre-fills **status**, **assignee**, **associations**, and **isShared** from the parent task. Values are captured once on mount — later parent updates (websocket, another tab, optimistic) won't clobber the user's in-progress edits. previewMode branches are left on their existing code paths.

Linear: https://linear.app/assemblycom/issue/OUT-3495/subtasks-should-inherit-fields-from-parent-issue-on-creation

## Implementation notes

- `defaultWorkflowState` is wrapped in `useState(() => …)` so its reference is stable across re-renders. This prevents `useHandleSelectorComponent`'s internal `setRenderingItem(item)` effect from overwriting the user's picked status if `activeTask` changes mid-flow.
- The old `useEffect` that forced `workflowStateId` to `todoWorkflowState.id` on mount is removed — the initial value is now seeded directly from `defaultWorkflowState?.id`.
- Preview-mode inheritance is intentionally gated: `workflowStateId`, `associations`, `isShared`, `assigneeValue`, and `taskAssociationValue` all fall back to the original preview-mode behavior when `previewMode` is truthy.
- Supersedes #1182 — same Linear issue, different approach (this version avoids the sync `useEffect` that would overwrite in-progress user edits on parent updates).

## Test plan

- [x] Parent with IU assignee → open subtask form → status, assignee pre-filled from parent.
- [x] Parent with client assignee (client + company) → open subtask form → assignee pre-filled, association selector hidden (as expected for client assignees).
- [x] Parent with IU + association + `isShared=true` → open subtask form → all three inherited, Share toggle on.
- [x] Unassigned parent → subtask form opens with no assignee pre-filled.
- [x] Change assignee/status in the form, leave the form open, update the parent task from another tab → form values do **not** revert.
- [ ] previewMode: open subtask form while previewing as a client → assignee/association defaults to the previewed client/company (unchanged from before); status defaults to `todo` (unchanged).
- [x] Create subtask → server accepts payload, subtask appears in list with inherited values.

### DIFF
--- a/src/app/detail/ui/NewTaskCard.tsx
+++ b/src/app/detail/ui/NewTaskCard.tsx
@@ -30,12 +30,14 @@ import {
   getSelectedViewerIds,
   getSelectorAssignee,
   getSelectorAssigneeFromFilterOptions,
+  getSelectorAssigneeFromTask,
+  getSelectorAssociationFromTask,
 } from '@/utils/selector'
 import { resolveAutofillTags, resolveDynamicFields } from '@/utils/dynamicFields'
 import { trimAllTags } from '@/utils/trimTags'
 import { Box, Stack, Typography } from '@mui/material'
 import dayjs from 'dayjs'
-import { useCallback, useEffect, useRef, useState } from 'react'
+import { useCallback, useRef, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { Tapwrite } from 'tapwrite'
 import { useAssociationLabelForWorkspace } from '@/hooks/useWorkspaceLabel'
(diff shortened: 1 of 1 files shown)


src/app/detail/ui/NewTaskCard.tsx
```diff
@@ -246,16 +263,16 @@ export const NewTaskCard = ({
...
+    if (inputValue.length && !isIU) {
       setTaskAssociationValue(null)
       handleFieldChange('associations', [])
     }
-    if (inputValue.length) {
-      setIsShared(false)
-      handleFieldChange('isShared', false)
-    }
+    setIsShared(false)
+    handleFieldChange('isShared', false)
```
priosshrsth, line 273: @arpandhakal Is this intentional?

src/app/detail/ui/NewTaskCard.tsx
```diff
@@ -145,16 +158,20 @@ export const NewTaskCard = ({
...
         assignee,
         { internalUserId: null, ...previewClientCompany }, // if preview mode, default select the respective client/company as viewer
       ) ?? null)
     : null
-  const [taskAssociationValue, setTaskAssociationValue] = useState<IAssigneeCombined | null>(previewTaskAssociation)
-  const [isShared, setIsShared] = useState(!!previewTaskAssociation)
+  const [taskAssociationValue, setTaskAssociationValue] = useState<IAssigneeCombined | null>(
+    previewTaskAssociation || (activeTask ? (getSelectorAssociationFromTask(assignee, activeTask) ?? null) : null),
+  )
+  const [isShared, setIsShared] = useState(previewTaskAssociation ? true : !!activeTask?.isShared)
```
priosshrsth, line 174:
```suggestion
  const [isShared, setIsShared] = useState(!!previewTaskAssociation || !!activeTask?.isShared)
```

priosshrsth (approved):
@arpandhakal The PR looks good to me. Left one questions and one optional nitpick. But it's ok to merge the PR without them.

src/app/detail/ui/NewTaskCard.tsx
arpandhakal: yes, the `isShared` value should be always false under certain circumstances, for example, when there are no associations, we just get constraint error on the api side so this is required. On second thought, we could just superrefine our payload through zod too

PR #1192, opened 2026-04-23
OUT-2563 | Reply notification failing when reply is created from CU

arpandhakal:
## Summary
- Fixes reply notifications that were silently dropped for cross-type replies (IU → CU's comment and CU → IU's comment) and 400'd for any client-originated reply in a multi-company workspace.
- `senderType` is now derived from the actual sender's user role (not the recipient's `initiatorType`), and `senderCompanyId` is passed when the sender is a client. Mirrors `NotificationService.buildNotificationDetails` / `handleIfSenderCompanyIdError`, including the single-company-workspace fallback that retries without `senderCompanyId` on Copilot's `'sender company ID is invalid based on sender'` error.
- Also flips an inverted `deletedAt` check that gated the parent-comment-author notification on the parent being *deleted* instead of *live* — so parent authors never received reply notifications, even in same-type cases.

## Root cause
`getInitiatorNotificationPromises` set `senderType: NotificationSenderSchema.parse(initiator.initiatorType)` — i.e. derived from the **recipient**, not the sender. Same-type replies (IU→IU, CU→CU) happened to work because recipient type coincidentally matched sender type. Cross-type replies sent a `senderId` / `senderType` mismatch, and client senders additionally lacked `senderCompanyId`, causing Copilot to 400 (confirmed via Trigger.dev run trace).

## Test plan
- [x] Deploy to Trigger staging: `yarn trigger:deploy-staging`
- [x] IU replies to CU's comment on a task → CU receives email notification
- [x] CU replies to IU's comment on a task → IU receives in-product notification
- [x] IU replies to IU's comment → IU receives in-product notification (regression check)
- [x] CU replies to CU's comment → CU receives email (regression check)
- [x] Reply to own comment → no self-notification
- [x] Top-level comment notifications still work (regression check on sibling job)
- [x] Single-company workspace: client-originated reply still succeeds via fallback branch

## Test

image

CU replies to IU comment notification

### DIFF
--- a/src/jobs/notifications/send-reply-create-notifications.ts
+++ b/src/jobs/notifications/send-reply-create-notifications.ts
@@ -87,7 +87,7 @@ export const sendReplyCreateNotifications = task({
         (initiator) => initiator.initiatorId === parentComment.initiatorId,
       )
       if (!isParentCommentDeleted && !parentInitiatorIsCurrentUser && !isNotificationAlreadySent) {
-        let promise = getInitiatorNotificationPromises(
+        const typedPromise = getInitiatorNotificationPromises(
           copilot,
           parentComment,
           senderId,
@@ -97,8 +97,9 @@ export const sendReplyCreateNotifications = task({
           payload.task.companyId || undefined,
         )
         // If there is no "initiatorType" for parentComment we have to be slightly creative (coughhackycough)
-        if (!promise) {
-          promise = getNotificationToUntypedInitiator(
+        const promise =
+          typedPromise ??
+          getNotificationToUntypedInitiator(
             copilot,
             parentComment,
             payload.task,
@@ -107,7 +108,6 @@ export const sendReplyCreateNotifications = task({
             senderCompanyId,
             deliveryTargets,
           )
-        }
         queueNotificationPromise(promise)
       }
     }
@@ -232,5 +232,8 @@ const getNotificationToUntypedInitiator = async (
     deliveryTargets,
     task.companyId || undefined,
     CommentInitiator.client,
-  )!
+  )!.catch((e) => {
+    console.error(e)
+    return undefined
+  })
 }
--- a/src/jobs/notifications/send-reply-create-notifications.ts
+++ b/src/jobs/notifications/send-reply-create-notifications.ts
@@ -1,7 +1,8 @@
-import { NotificationSender, NotificationSenderSchema } from '@/types/common'
+import { NotificationRequestBody, NotificationSender } from '@/types/common'
 import { getAssigneeName } from '@/utils/assignee'
 import { copilotBottleneck } from '@/utils/bottleneck'
 import { CopilotAPI } from '@/utils/CopilotAPI'
+import { isMessagableError } from '@/utils/copilotError'
 import { CommentRepository } from '@/app/api/comments/comment.repository'
 import { CommentService } from '@/app/api/comments/comment.service'
 import User from '@api/core/models/User.model'
@@ -39,6 +40,9 @@ export const sendReplyCreateNotifications = task({
       .string()
       .uuid()
       .parse(user.internalUserId || user.clientId)
+    const senderType: NotificationSender = user.internalUserId ? CommentInitiator.internalUser : CommentInitiator.client
+    // Copilot requires senderCompanyId when the sender is a client in a multi-company workspace
+    const senderCompanyId = senderType === CommentInitiator.client ? user.companyId : undefined
 
     const deliveryTargets = await getNotificationDetails(copilot, user, comment)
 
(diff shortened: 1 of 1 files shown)


src/jobs/notifications/send-reply-create-notifications.ts
```diff
@@ -75,7 +81,7 @@ export const sendReplyCreateNotifications = task({
       // - Parent Comment hasn't been deleted yet
       // - Parent Comment initiatorId isn't this current user
       // - Parent comment hasn't been already sent a notification through a reply
-      const isParentCommentDeleted = !parentComment.deletedAt
+      const isParentCommentDeleted = !!parentComment.deletedAt
```
priosshrsth, line 84: Was this really wrong?

src/jobs/notifications/send-reply-create-notifications.ts
```diff
@@ -137,36 +153,57 @@ const getInitiatorNotificationPromises = async (
...
 const getNotificationToUntypedInitiator = async (
   copilot: CopilotAPI,
   parentComment: Comment,
   task: Task,
   senderId: string,
+  senderType: NotificationSender,
+  senderCompanyId: string | undefined,
   deliveryTargets: { inProduct: Record<'title', any>; email: object },
 ) => {
   let promise
```
priosshrsth: remove this.

src/jobs/notifications/send-reply-create-notifications.ts
priosshrsth:
```suggestion
   return getInitiatorNotificationPromises(
```

priosshrsth, line 207: Are initiator always internal User?

src/jobs/notifications/send-reply-create-notifications.ts
```diff
@@ -137,36 +153,57 @@ const getInitiatorNotificationPromises = async (
...
+    return await copilot.createNotification(body)
+  } catch (e) {
+    if (isMessagableError(e) && e.body?.message === 'sender company ID is invalid based on sender') {
+      return await copilot.createNotification({ ...body, senderCompanyId: undefined })
+    }
+    throw e
+  }
 }
 
 const getNotificationToUntypedInitiator = async (
```
priosshrsth, line 197: @arpandhakal I think this function can be simplified with early returns without having nested try catch block.

priosshrsth (changes requested):

src/jobs/notifications/send-reply-create-notifications.ts
arpandhakal: yes because !parentComment.deletedAt is true when deletedAt is null which is not correct.

src/jobs/notifications/send-reply-create-notifications.ts
arpandhakal: No initiators can be either client or an IU. This function only runs when initiator.initiatorType is null (legacy comments predating the column). The function getNotificationToUntypedInitiator refers to this 'unknown type' cases.

src/jobs/notifications/send-reply-create-notifications.ts
```diff
@@ -132,68 +148,89 @@ const getNotificationDetails = async (copilot: CopilotAPI, user: User, comment:
...
+  return getInitiatorNotificationPromises(
+    copilot,
+    parentComment,
+    senderId,
+    senderType,
+    senderCompanyId,
+    deliveryTargets,
+    task.companyId || undefined,
+    CommentInitiator.client,
+  )!
```
priosshrsth:
```suggestion
  return getInitiatorNotificationPromises(
    copilot,
    parentComment,
    senderId,
    senderType,
    senderCompanyId,
    deliveryTargets,
    task.companyId || undefined,
    CommentInitiator.client,
  ).catch(() => undefined)
```

src/jobs/notifications/send-reply-create-notifications.ts
```diff
@@ -132,68 +148,89 @@ const getNotificationDetails = async (copilot: CopilotAPI, user: User, comment:
...
+    body = {
       senderId,
-      senderType: assume || NotificationSenderSchema.parse(initiator.initiatorType),
+      senderType,
+      senderCompanyId,
       recipientClientId: initiator.initiatorId,
       recipientCompanyId: initiatorCompanyId,
       deliveryTargets: { email: deliveryTargets.email },
-    })
+    }
```
priosshrsth, line 177:
```suggestion
  const body: NotificationRequestBody = {senderId,
      senderType,
      senderCompanyId,};
if (initiator.initiatorType === CommentInitiator.internalUser || assume === CommentInitiator.internalUser) {
    Object.assign(body, {
recipientInternalUserId: initiator.initiatorId,
deliveryTargets: { inProduct: deliveryTargets.inProduct },
    });
} else if (initiator.initiatorType === CommentInitiator.client || assume === CommentInitiator.client) {
    Object.asign(body,  {
recipientClientId: initiator.initiatorId,
recipientCompanyId: initiatorCompanyId,
deliveryTargets: { email: deliveryTargets.email },
    })
```

@arpandhakal This is optional. But wdyt about this?

priosshrsth (changes requested):

priosshrsth (approved):
lgtm

PR #1193, opened 2026-04-24
OUT-3648 | Tasks created in the webapp should not have the user agent as "node"

arpandhakal:
## Changes

- [x] Forwarded x-assembly-user-agent through server actions. 

## Testing Criteria

image

### DIFF
--- a/src/app/(home)/actions.ts
+++ b/src/app/(home)/actions.ts
@@ -4,6 +4,13 @@ import { apiUrl } from '@/config'
 import { CreateAttachmentRequest } from '@/types/dto/attachments.dto'
 import { CreateTaskRequest, UpdateTaskRequest } from '@/types/dto/tasks.dto'
 import { CreateViewSettingsDTO } from '@/types/dto/viewSettings.dto'
+import { headers } from 'next/headers'
+
+const getForwardedAssemblyHeaders = async (): Promise<Record<string, string>> => {
+  const h = await headers()
+  const userAgent = h.get('user-agent')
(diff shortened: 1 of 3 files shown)


src/app/(home)/actions.ts
```diff
@@ -4,6 +4,13 @@ import { apiUrl } from '@/config'
 import { CreateAttachmentRequest } from '@/types/dto/attachments.dto'
 import { CreateTaskRequest, UpdateTaskRequest } from '@/types/dto/tasks.dto'
 import { CreateViewSettingsDTO } from '@/types/dto/viewSettings.dto'
+import { headers } from 'next/headers'
+
+const getForwardedAssemblyHeaders = async (): Promise<Record<string, string>> => {
+  const h = await headers()
+  const userAgent = h.get('user-agent')
+  return userAgent ? { 'x-assembly-user-agent': userAgent } : {}
+}
```
priosshrsth: @arpandhakal This might not be enough. What about ip?

priosshrsth (changes requested):

src/app/(home)/actions.ts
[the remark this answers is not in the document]
priosshrsth: @arpandhakal Adressing this might be a good idea.

priosshrsth: @arpandhakal Let's use `||` instead of `??`

priosshrsth (changes requested):

priosshrsth (approved):

PR #1199, opened 2026-04-30
OUT-3174 | Use task title for breadcrumbs instead of task id

arpandhakal:
## Summary
- Breadcrumbs on the task details page now show the task **title** (truncated to 25 chars) instead of the task id, on both internal-user (`iu`) and client (`cu`) views.
- Drive-by fix: `useWindowWidth` was reading `window.outerWidth`, which doesn't shrink under Chrome dev-tools device emulation (and reports the parent window inside iframes). Switched to `window.innerWidth` so the mobile sidebar/description stacking layout fires correctly in dev tools and matches real-device behavior.

Linear: https://linear.app/assemblycom/issue/OUT-3174

## Test plan
- [x] Open a task with a long title — breadcrumb shows the title truncated to 25 chars with an ellipsis.
- [x] Open a nested subtask — each ancestor in the breadcrumb shows its (truncated) title and links to the correct task.
- [x] Verify on both an internal-user portal and a client portal.
- [x] In Chrome dev tools, toggle mobile device emulation on the details page — sidebar collapses and description stacks vertically (previously stayed side-by-side).
- [x] On a real mobile device, layout still stacks vertically as before.
- [x] Verify the embedded portal view still renders correctly at its normal width.

## Test 

image

An issue I found with the platform header: 
- [LOOM](https://www.loom.com/share/da6a16b8146644e4b822323813121449)


SandipBajracharya (dismissed):

priosshrsth (approved):
lgtm

PR #1205, opened 2026-05-06
OUT-2898 | Show tasks assigned to clients within a company on company details page

arpandhakal:
## Summary
- When an internal user enters CRM preview mode for a company, also include tasks assigned to clients of that company (TEAM TASKS) and IU tasks shared with those clients (SHARED WITH TEAM). Previously only tasks assigned to the company itself and IU tasks shared with the company were visible.
- Mirrored the same access rule in `CommentService` so comments stay aligned with task visibility (and fixed a latent strict `parse(clientId)` that would have thrown for any company-only auth context).
- The shared task/comment filter is now async; `copilot.getCompanyClients` is only called in the new IU-company-preview branch and the result is memoized per request to avoid duplicate calls.

## Context
Linear: [OUT-2898](https://linear.app/assemblycom/issue/OUT-2898/show-tasks-assigned-to-clients-within-a-company-on-company-details)

Today the company-details CRM view shows:
1. Tasks assigned to the company
2. IU tasks shared with the company

This PR adds:
3. Tasks assigned to clients within the company
4. IU tasks shared with those clients

Other flows (client preview, CU portal, normal IU listing) skip the new branch entirely, so they incur no extra I/O — only an `await` on a synchronously resolved promise.

## Test plan
- [x] As an IU, enter CRM preview for a company that has clients with their own tasks → confirm those tasks now appear in the company task list
- [x] In the same view, confirm IU-assigned tasks shared with one of the company's clients appear (SHARED WITH TEAM)
- [x] Existing visibility unchanged: tasks assigned to the company and IU tasks shared with the company still appear
- [x] Open one of the new tasks and confirm comments load (CommentService access path)
- [x] Client preview (an actual CU) still sees only their own + company-shared tasks — no leakage of other clients' tasks
- [x] CU portal user (no internalUserId) does not gain access to other clients' tasks
- [x] Normal IU board listing performance unchanged (no extra Copilot calls in non-preview flows)

## Testing 

image
Task assigned to a client of the company is also visible for company CRM. 

image
Task associated with a client of the company is also visible for company CRM.


src/app/api/comments/comment.service.ts
```diff
@@ -503,7 +504,33 @@ export class CommentService extends BaseService {
...
   } //Repeated twice because taskSharedService is an abstract class.
+
+  private _companyClientIdsPromise?: Promise<string[]>
+
+  private async getCompanyClientIds(companyId: string): Promise<string[]> {
+    if (!this._companyClientIdsPromise) {
+      this._companyClientIdsPromise = this.copilot.getCompanyClients(companyId).then((clients) => clients.map((c) => c.id))
+    }
+    return this._companyClientIdsPromise
+  }
```
priosshrsth: @arpandhakal Why are we doing it this way? This feels wrong. the function explicitly required companyId and `_companyClientIdsPromise?` doesn't account for this. Do we even need to do this? And again if we need then react cache is the best way probably to do these things.

priosshrsth (changes requested):

priosshrsth (approved):
lgtm

PR #1207, opened 2026-05-06
feat: workspace fetch optimization + parent tasks fetching

priosshrsth:
## Summary

resolves OUT-3705 [OUT-3705: Optimize workspace to be fetched once per session](https://linear.app/assemblycom/issue/OUT-3705/optimize-workspace-to-be-fetched-once-per-session)

- [x] Fetch workspace on client side and once per session
- [x] Update rest of the codebase to use portalUrl from store rather than props.
- [x] For subtasks we get the details of both tasks with single query instead of multiple 
- [x] resolves OUT-3695 [OUT-3695: Remove customFields from clients response](https://linear.app/assemblycom/issue/OUT-3695/remove-customfields-from-clients-response)

## Test plan

- [x] App boots, workspace data appears in redux after first paint
- [x] Detail page: sidebar + header breadcrumbs render correctly with `portalUrl`-dependent CTAs
- [x] Home / client board: app bridges (CTAs, breadcrumbs) work
- [x] Manage templates list + detail: app bridges work
- [x] First-paint flash on `portalUrl`-dependent UI is not visually disruptive
- [x] Subtask detail page: ancestor breadcrumbs render in correct root → leaf order (parent tasks ordering fix)

https://github.com/user-attachments/assets/42d86a99-eb06-4711-a61d-655282705d09

### DIFF
--- a/src/app/api/tasks/tasks.service.ts
+++ b/src/app/api/tasks/tasks.service.ts
@@ -704,7 +704,6 @@ export class TasksService extends TasksSharedService {
   }
 
   async getTraversalPath(id: string): Promise<AncestorTaskResponse[]> {
-    const pathQueryStartedAt = performance.now()
     const taskWithPath = (
       await this.db.$queryRaw<{ path: string }[]>`
       SELECT "path" from "Tasks"
@@ -712,7 +711,7 @@ export class TasksService extends TasksSharedService {
       LIMIT 1
     `
     )?.[0]
-    console.info(`[perf] getTraversalPath path-query id=${id} took ${(performance.now() - pathQueryStartedAt).toFixed(2)}ms`)
+
     if (!taskWithPath) {
       throw new APIError(httpStatus.NOT_FOUND, 'The requested task was not found')
     }
@@ -720,7 +719,7 @@ export class TasksService extends TasksSharedService {
     const parentIds = getIdsFromLtreePath(taskWithPath.path)
     const parentTasksStartedAt = performance.now()
 
-    const parentTasks = (await this.db.task.findMany({
+    const fetchedParents = (await this.db.task.findMany({
       where: {
         id: {
           in: parentIds,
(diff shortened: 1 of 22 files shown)


src/app/api/tasks/tasks.service.ts
```diff
@@ -711,28 +711,41 @@ export class TasksService extends TasksSharedService {
...
-            companyId: true,
-            internalUserId: true,
-            associations: true,
-            isShared: true,
-          },
-        }),
-      ) as Promise<AncestorTaskResponse>[],
-    )
+    const parentIds = getIdsFromLtreePath(taskWithPath.path)
+    const parentTasksStartedAt = performance.now()
```
arpandhakal: is this necessary to be in prod?

src/types/common.ts
```diff
@@ -94,9 +94,9 @@ export const ClientResponseSchema = z.object({
   companyIds: z.array(z.string().uuid()).optional(),
   status: z.string(),
   avatarImageUrl: z.string().nullable(),
-  customFields: z
-    .record(z.string(), z.union([z.string().nullable(), z.array(z.string()).nullable(), z.object({}).nullable()]))
-    .nullable(),
+  // customFields: z
+  //   .record(z.string(), z.union([z.string().nullable(), z.array(z.string()).nullable(), z.object({}).nullable()]))
+  //   .nullable(),
```
arpandhakal: lets remove the commented code or add an extra comment on why we commented this out.

arpandhakal (approved):

arpandhakal:
Lets consider testing in the platform itself in a custom app. That way we can ensure everything works including the app bridge.

src/app/api/tasks/tasks.service.ts
priosshrsth: @arpandhakal Addressed.

src/types/common.ts
priosshrsth: @arpandhakal Addressed.

PR #1208, opened 2026-05-07
OUT-3661 | Limited access IUs shouldn't see tasks that are associated with clients/companies beyond their access list

arpandhakal:
## Summary
Closes [OUT-3661](https://linear.app/assemblycom/issue/OUT-3661/limited-access-ius-shouldnt-see-tasks-that-are-associated-with).

Limited-access IUs were filtered only by task **assignee** company. Tasks merely **associated** with an out-of-scope client/company still leaked through. This PR closes that gap across the API, cascade SQL, and realtime paths.

To make the SQL filter clean and avoid expensive Copilot SDK enumeration of all clients/companies, `Tasks.associations` is migrated from `jsonb[]` to `jsonb` (still holding an array). That unlocks Prisma's `array_contains` (`@>` containment), which lets us match associations by `companyId` directly using `companyAccessList`.

## Schema + migration
- `Tasks.associations` column: `jsonb[]` → `jsonb`
- Single in-place migration with `USING coalesce(to_jsonb(associations), '[]'::jsonb)` — Postgres rewrites every row during `ALTER`. No separate app-side backfill. JS shape (`[{clientId?, companyId}]`) is preserved across the migration.
- AccessExclusiveLock duration ~ a few seconds at ~80k rows.

## Core fix (`tasksShared.service.ts`)
- `filterTasksByClientAccess` now also rejects tasks whose `associations` reference a company outside `companyAccessList` — this is the actual user-visible fix for the read path.
- `getAccessFilterForTasks` ANDs the new `getAccessibleAssociationsFilter` for limited IUs, so cascade workflow updates and subtask counts stay in sync with the read filter.
- `getDisjointTasksFilter` (limited IU branch) treats a parent as inaccessible when its associations are out of scope. Combined into one `parent: { OR: [...] }` to avoid Prisma double-joining the parent relation.
- New helper `getAccessibleAssociationsFilter` uses `array_contains` (Postgres `@>`) for partial-match — `{companyId: cid}` matches both stored `{companyId}` and `{clientId, companyId}` shapes without needing to enumerate clientIds.
- Existing `hasSome` filters (which relied on jsonb exact-element equality) converted to `OR: [{equals: ...}, ...]`. Behavior is identical for `max=1` arrays — semantics-preserving.
- `validateTaskShare` updated for the new `Json` (single) shape with an `Array.isArray` guard.

## `hasSome` → `equals` conversions in adjacent files
Mechanical, semantics-preserving:
- `src/app/api/tasks/tasks.service.ts` — `getIncompleteTasksForCompany`, `resetAllSharedTasks`
- `src/app/api/comments/comment.service.ts` — three CU/CRM filter sites
- `src/app/api/webhook/webhook.service.ts` — `handleCompanyUnassignment`
- OUT-2898's IU-company-preview filters in `tasksShared` and `comment` services

## Realtime sync (`src/lib/realtime.ts`)
Realtime guardrails for limited IUs only checked `task.companyId`. Now also check `associations` so a task with `companyId = null` but an out-of-scope association can no longer slip into the store via realtime.
- New private helper `hasInaccessibleAssociation` mirrors the API filter.
- Four call sites updated: `isSubtaskAccessible`, `handleRealtimeTaskInsert` guardrail, `isReassignedOutOfLimitedIUScope`, `isReassignedIntoLimitedIUScope`.

## Disjoint duplicate fix (`src/hooks/useFilter.tsx`)
Pre-existing latent bug: limited-IU disjoint-promoted subtasks rendered twice in the board after AllTasksFetcher Suspense resolved (the "second card appears with a delay" symptom). The standalone-subtasks dedup checked only `parentId`. Now also checks the subtask's own `id`, so a task that's already in `filteredParentTasks` (via disjoint promotion) isn't re-added as a standalone.

## Type tightening
- `src/app/api/tasks/subtasks.service.ts`: `JsonValue[]` → `JsonValue`
- `src/app/api/tasks/public/public.service.ts`: `Prisma.JsonValue[]` → `Prisma.JsonValue`

## Pre-existing oddities preserved (NOT introduced or worsened by this PR — worth a follow-up ticket)
- `getIncompleteTasksForCompany` filters for `{clientId: null, companyId}` but stored objects don't store `null` clientIds (the key is either present with a UUID or absent) — this branch likely never matches.
- `resetAllSharedTasks` filters for `{clientId: assigneeId}` (no companyId) but `companyId` is required by `AssociationSchema` — that half of the OR likely never matches.

Both behaviors carried over identically.

## Test plan
- [x] As a limited-access IU, view the board with a task assigned to a teammate IU but associated with an out-of-scope CU — should NOT appear (was the OUT-3661 repro).
- [x] As a limited-access IU, view a parent task that is inaccessible (out-of-scope companyId or out-of-scope association) with an accessible subtask — the subtask appears at root level **exactly once** (was duplicated).
- [x] As a limited-access IU, watch the board while another session creates an inaccessible-via-association task — does NOT appear via realtime.
- [x] As a limited-access IU, when a task on your board is updated to add an out-of-scope association — task is removed from the board via realtime.
- [x] As a CU, verify you do NOT see tasks associated with another client at the same company (security-critical: `equals` preserves the prior `hasSome` strict-match semantics).
- [x] As a CU, verify you DO see tasks shared with you (`{clientId, companyId}` association) and tasks associated company-wide (`{companyId}` association).
- [x] CRM IU previewing a client — tasks for that client appear, including shared tasks.
- [x] CRM IU previewing a company — tasks for that company AND tasks for clients within that company appear (OUT-2898 behavior preserved).
- [x] Full-access IU — board behavior unchanged.

## Testing 

- [LOOM](https://www.loom.com/share/337cff7473ba442d97bc3f082f405da5)

### DIFF
--- a/src/app/api/comments/comment.service.ts
+++ b/src/app/api/comments/comment.service.ts
@@ -2,7 +2,7 @@ import { AttachmentsService } from '@/app/api/attachments/attachments.service'
 import { PublicCommentSerializer } from '@/app/api/comments/public/public.serializer'
 import { sendCommentCreateNotifications } from '@/jobs/notifications'
 import { sendReplyCreateNotifications } from '@/jobs/notifications/send-reply-create-notifications'
-import { InitiatedEntity, TempClientFilter } from '@/types/common'
+import { InitiatedEntity } from '@/types/common'
 import { CreateAttachmentRequestSchema } from '@/types/dto/attachments.dto'
 import { CommentsPublicFilterType, CommentWithAttachments, CreateComment, UpdateComment } from '@/types/dto/comment.dto'
 import { DISPATCHABLE_EVENT } from '@/types/webhook'
@@ -479,14 +479,17 @@ export class CommentService extends BaseService {
         { companyId, clientId: null },
       )
       if (includeAssociatedTask) {
-        const tempClientFilter: TempClientFilter = {
-          associations: {
-            hasSome: [{ clientId, companyId }, { companyId }],
-          },
+        // Match tasks associated with this specific client+company OR with the company at large.
+        // `equals` preserves the prior exact-match semantics.
+        const associationVariants: Prisma.TaskWhereInput[] = [
+          { associations: { equals: [{ clientId, companyId }] } },
+          { associations: { equals: [{ companyId }] } },
+        ]
+        if (isCuPortal) {
+          filters.push({ OR: associationVariants, isShared: true })
+        } else {
+          filters.push(...associationVariants)
         }
-        if (isCuPortal) tempClientFilter.isShared = true
-        // Get tasks that includes the client as a association
-        filters.push(tempClientFilter)
       }
     } else if (companyId) {
       filters.push(
@@ -494,15 +497,16 @@ export class CommentService extends BaseService {
         { clientId: null, companyId },
       )
 
-      // Get tasks that includes the company as a association
+      // Get tasks that include the company as an association
       if (includeAssociatedTask) {
-        const tempCompanyFilter: TempClientFilter = {
-          associations: {
-            hasSome: [{ companyId }],
-          },
+        const companyAssociationFilter: Prisma.TaskWhereInput = {
+          associations: { equals: [{ companyId }] },
         }
-        if (isCuPortal) tempCompanyFilter.isShared = true
-        filters.push(tempCompanyFilter)
+        if (isCuPortal) {
+          filters.push({ ...companyAssociationFilter, isShared: true })
+        } else {
+          filters.push(companyAssociationFilter)
+        }
       }
 
       // OUT-2898: When an IU is previewing a company in the CRM, also include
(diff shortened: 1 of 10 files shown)


priosshrsth (approved):

src/app/api/comments/comment.service.ts
```diff
@@ -479,30 +479,34 @@ export class CommentService extends BaseService {
...
       )
       if (includeAssociatedTask) {
-        const tempClientFilter: TempClientFilter = {
-          associations: {
-            hasSome: [{ clientId, companyId }, { companyId }],
-          },
+        // Match tasks associated with this specific client+company OR with the company at large.
+        // `equals` preserves the prior exact-match semantics.
+        const associationVariants: Prisma.TaskWhereInput[] = [
+          { associations: { equals: [{ clientId, companyId }] } },
```
SandipBajracharya, line 485: Nice upgrade. We can also explore `array_contains` by Prisma.  For equals, the order should be exact or else it will fail to meet the condition even if the values are correct. `array_contains` solves this problem. Plus with `array_contains`, we can also support multiple associations, not a requirement now but if that can be a thing in future.

priosshrsth (approved):
lgtm.

SandipBajracharya (approved):

src/app/api/comments/comment.service.ts
SandipBajracharya: No, what I meant is {company1, client1} is not equals {client1, company1}.

SandipBajracharya: Regarding the ordering, order matters when there are multiple objects in the array. Since we have only one object, this should be good.

PR #1211, opened 2026-05-08
OUT-3494 | Improve default sorting for subtasks

arpandhakal:
## Summary
- Adds `sortSubtasksByPriority` — sorts subtasks by status (`started` → `unstarted` → `completed` → `backlog` → `cancelled`), then `dueDate` ASC nulls last, then `createdAt` ASC (oldest first), with `id` as a stability tiebreaker.
- Applies the new sort on the details page (`Subtasks.tsx`, including the optimistic-insert path) and on the board/list views via `subtasksByTaskId` in `TaskBoard.tsx`. List view inherits the order through the same dictionary.
- Top-level task sort (`sortTaskByDescendingOrder`) is unchanged. Templates (`Subtemplates.tsx`) and the backend orderBy are not touched — every web subtask render path re-sorts client-side, so the DB order is irrelevant for what users see.

Linear: https://linear.app/assemblycom/issue/OUT-3494/improve-default-sorting-for-subtasks

## Test plan
- [x] Open a task with subtasks in mixed states — verify `In Progress` items appear first, `Todo` next, `Done` after, `Backlog`/`Cancelled` last
- [x] Within a single status, verify subtasks with due dates sort earliest-first and subtasks without a due date appear after those with one
- [x] Within a single status with all-equal due dates (or none), verify subtasks created earlier appear above newer ones
- [x] Create three subtasks back-to-back in the details page — confirm they appear in creation order (optimistic insert respects the new sort)
- [x] Verify the board view and list view both reflect the new subtask ordering under each parent
- [x] Verify top-level task lists (board columns, list view sections) are unaffected


priosshrsth (approved):
lgtm

PR #1215, opened 2026-05-11
perf(detail): react.cache Copilot calls + skip client refire

priosshrsth:
Ref: [OUT-3701](https://linear.app/assemblycom/issue/OUT-3701/optimize-task-detail-page)

Foundation for the detail-page perf work. Two low-risk changes that the deeper refactor in the stacked PR builds on.

## Changes
- **`CopilotAPI`** — wrap `getInternalUser` + `getWorkspace` in `react.cache`. Per-request memoization deduplicates Copilot calls within a single RSC / route handler invocation. Falls back to no-op in non-request contexts (jobs, CLI).
- **`OneTaskDataFetcher`** — new `useFallback` prop that seeds SWR with the SSR-rendered task and disables `revalidateOnMount`. Default behavior unchanged; the detail page opts in via the stacked PR.

## Measured impact
Validated dedup: 4 call sites of `getInternalUser(self)` collapse to 1 actual network fetch on a typical detail-page load.

## Test plan
- [x] Detail page loads for IU + client tokens
- [x] Mutations to task (title, body, assignee, workflow state) still work
- [x] Non-detail pages (home, configure) unaffected

## Stacked PR
The actual page handler refactor lives in #1214 (against this branch).


src/utils/CopilotAPI.ts
```diff
@@ -377,3 +378,25 @@ export class CopilotAPI {
...
+    return withRetry(copilot._getInternalUser.bind(copilot), [id])
+  },
+)
+
+const cachedFetchWorkspace = cache(
+  async (token: string, customApiKey: string | undefined): Promise<WorkspaceResponse> => {
+    const copilot = new CopilotAPI(token, customApiKey)
+    return withRetry(copilot._getWorkspace.bind(copilot), [])
+  },
+)
```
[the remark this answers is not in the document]
priosshrsth: If the request even after 3 retries failed then another call to the function will 99.99% fail too.

src/app/_fetchers/OneTaskDataFetcher.tsx
```diff
@@ -7,25 +7,52 @@ import { PropsWithToken } from '@/types/interfaces'
...
+  useFallback?: boolean
 }
 
-export const OneTaskDataFetcher = ({ token, task_id, initialTask }: OneTaskDataFetcherProps & PropsWithToken) => {
+export const OneTaskDataFetcher = ({
+  token,
+  task_id,
+  initialTask,
+  useFallback,
+}: OneTaskDataFetcherProps & PropsWithToken) => {
```
SandipBajracharya: Do we need to do `& PropsWithToken` here? I see we did `OneTaskDataFetcherProps extends PropsWithToken` on line 12.

src/utils/CopilotAPI.ts
```diff
@@ -169,7 +170,7 @@ export class CopilotAPI {
 
   async _deleteClient(id: string) {
     console.info('CopilotAPI#_deleteClient', this.token)
-    return await this.copilot.deleteClient({ id })
+    return this.copilot.deleteClient({ id })
```
SandipBajracharya, line 173: Is the removal of `await` intentional?

priosshrsth: Yes @SandipBajracharya This is intentional. The await in this was useless.

src/app/_fetchers/OneTaskDataFetcher.tsx
priosshrsth: @SandipBajracharya Good catch. Updated.

SandipBajracharya (approved):
LGTM

PR #1216, opened 2026-05-11
OUT-3497 | Custom status labels: allow workspaces to rename Todo, In Progress, and Done

arpandhakal:
## Summary

- Workspace admins can rename the three default workflow state labels (To Do / In Progress / Done) from a new **Status Customization** section on the Configure Tasks App page, inline per row with Cancel/Save.
- Renames persist directly to `WorkflowState.name`, so every surface that already reads it (board view, list view, detail selector, activity log, template board) picks up the custom label with zero new render-time override logic.
- The public API is unaffected — `PublicTaskSerializer` maps `workflowState.type` through a hardcoded `statusMap`, so the public `status` field continues to return the default values (`todo` / `inProgress` / `completed`).

## Implementation notes

- New `PATCH /api/workflow-states/[id]` endpoint, scoped by `workspaceId`, gated by `UserAction.Update, Resource.Tasks`. Cross-workspace IDs return 404 by design.
- `UpdateWorkflowStateRequestSchema` validates name (`trim`, non-empty, max 255).
- Default seed names are extracted into `src/constants/workflowStates.ts` so they have a single source of truth (used by `setDefaultWorkflowStates` and future reset-to-default flows).
- Frontend section matches Figma `6088:7818` — `52px` row, `28px` text-field, `25px` Cancel/Save buttons. Edit pencil reveals on hover; on touch devices (`@media (hover: none)`) it stays visible.
- After a successful rename, the updated workflow states are dispatched to Redux (`setWorkflowStates`) so the TemplateBoard on the same page re-renders without a refetch.
- Added an optional `height` prop to `PrimaryBtn` (mirroring `SecondaryBtn`) for consistent sizing.
- Updated Auto-archive + Status Customization to use `theme.color.borders.border` instead of the lighter `borderDisabled` so card borders match the Templates section visually.

## Test plan

- [x] Fresh workspace seeds with the default "To Do" / "In Progress" / "Done" names.
- [x] Renaming a state on `/configure-tasks-app` persists across reload.
- [x] Renamed labels appear on the main task board column headers, list view, task detail workflow selector, and activity log.
- [x] TemplateBoard on the configure page reflects renames immediately (Redux dispatch path).
- [x] Public API (`GET /api/v1/tasks/:id` with an API key) still returns the canonical `status` value, not the renamed label.
- [x] CU attempting `PATCH /api/workflow-states/:id` is denied via `PoliciesService`.
- [x] Cross-workspace ID PATCH returns 404.
- [x] Empty / whitespace-only / unchanged name values disable the Save button or reject the request.
- [x] Mobile: the edit pencil is always visible, hover background isn't required.
- [x] Row height stays constant (52px) across display ↔ edit transitions for all three rows.

### DIFF
--- a/src/app/configure-tasks-app/actions.ts
+++ b/src/app/configure-tasks-app/actions.ts
@@ -2,6 +2,7 @@
 import { apiUrl } from '@/config'
 import { CreateTemplateRequest, UpdateTemplateRequest } from '@/types/dto/templates.dto'
 import { UpdateWorkspaceSettingsDTO } from '@/types/dto/workspaceSettings.dto'
+import { UpdateWorkflowStateRequest } from '@/types/dto/workflowStates.dto'
 
 export const createNewTemplate = async (token: string, payload: CreateTemplateRequest) => {
   const resp = await fetch(`${apiUrl}/api/tasks/templates?token=${token}`, {
@@ -43,4 +44,16 @@ export async function updateWorkspaceSettings(token: string, payload: UpdateWork
     throw new Error(`Failed to update workspace settings: ${resp.status}`)
   }
   return await resp.json()
+}
+
+export async function updateWorkflowState(token: string, id: string, payload: UpdateWorkflowStateRequest) {
+  const resp = await fetch(`${apiUrl}/api/workflow-states/${id}?token=${token}`, {
+    method: 'PATCH',
+    body: JSON.stringify(payload),
+  })
+  if (!resp.ok) {
+    throw new Error(`Failed to update workflow state: ${resp.status}`)
+  }
+  const { updatedWorkflowState } = await resp.json()
+  return updatedWorkflowState
 }
--- a/src/app/configure-tasks-app/ui/StatusCustomizationSection.tsx
+++ b/src/app/configure-tasks-app/ui/StatusCustomizationSection.tsx
@@ -120,7 +120,7 @@ export const StatusCustomizationSection = ({ initialWorkflowStates, token }: Sta
                 px: '12px',
                 py: '12px',
                 minHeight: '52px',
-                borderTop: (theme) => (index === 0 ? 'none' : `1px solid ${theme.color.borders.border}`),
+                boxShadow: (theme) => (index === 0 ? 'none' : `inset 0 1px 0 0 ${theme.color.borders.border}`),
                 '&:hover': isEditing
                   ? undefined
                   : {
--- a/src/app/configure-tasks-app/ui/StatusCustomizationSection.tsx
+++ b/src/app/configure-tasks-app/ui/StatusCustomizationSection.tsx
@@ -1,0 +1,219 @@
+'use client'
+
+import { useEffect, useRef, useState } from 'react'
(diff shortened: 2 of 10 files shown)


priosshrsth (dismissed):
@arpandhakal lgtm

src/constants/workflowStates.ts
```diff
@@ -0,0 +1,9 @@
+import { StateType } from '@prisma/client'
+
+export const DEFAULT_WORKFLOW_STATE_NAMES: Record<StateType, string> = {
+  unstarted: 'To Do',
+  started: 'In Progress',
+  completed: 'Done',
+  backlog: 'Backlog',
+  cancelled: 'Cancelled',
```
priosshrsth, line 8: @arpandhakal Will it be hard to change the key to todo, inProgress, done, etc. I think these are more intuitive. But just my opinion. So I am ok with this as well.

src/app/configure-tasks-app/ui/StatusCustomizationSection.tsx
```diff
@@ -0,0 +1,219 @@
...
+import { WorkflowStateResponse } from '@/types/dto/workflowStates.dto'
+import { PrimaryBtn } from '@/components/buttons/PrimaryBtn'
+import { SecondaryBtn } from '@/components/buttons/SecondaryBtn'
+import { updateWorkflowState } from '@/app/configure-tasks-app/actions'
+import store from '@/redux/store'
+import { setWorkflowStates } from '@/redux/features/taskBoardSlice'
+
+const STATUS_ICON_BY_TYPE: Record<StateType, 'ToDo' | 'InProgress' | 'SuccessSolid'> = {
+  unstarted: 'ToDo',
+  backlog: 'ToDo',
```
SandipBajracharya, line 16: Do we even have `backlog` and `cancelled` status in use?

src/app/configure-tasks-app/actions.ts
```diff
@@ -44,3 +45,15 @@ export async function updateWorkspaceSettings(token: string, payload: UpdateWork
   }
   return await resp.json()
 }
+
+export async function updateWorkflowState(token: string, id: string, payload: UpdateWorkflowStateRequest) {
```
SandipBajracharya, line 49: We could have directly made api call from client instead of creating server action and making api call. Since the same structure is followed in multiple places, it should be okay for now I guess. We can work on this in improvement maybe?

src/app/configure-tasks-app/ui/StatusCustomizationSection.tsx
arpandhakal: We don't have them in use as of now. But we had them in workflowstatus type since long. We might wanna keep these just in case.

src/app/configure-tasks-app/actions.ts
arpandhakal: This is a pretty simple patch api request without any realtime, revalidation, complex mutation requirements. So I think this is okay.

SandipBajracharya (approved):
lgtm

PR #1217, opened 2026-05-11
OUT-3713 | Tasks app outage from subtasks update payload

arpandhakal:
## Changes

- [x] When subtasks are updated in board page or details page, subsequent realtime changes if captured through `postgres_changes` from our realtime handler. The payload from realtime does not contain `workflowstate` object because it is a raw table payload without any join operation. But the subtask sorting handler strictly required the `workflowstate` object. Made this optional which acts as a defense mechanism not to crash the app. 
- [x] While setting tasks on `accessibleTask` store from realtime payload, created a map of `workflowstate` object to pass to the sorting handler. This acts as a solution to fix our problem. 
- [x] added a quick improvement on optimistic updates. 

 
## Testing Criteria

- [LOOM](https://www.loom.com/share/fee19ac46a15456195cf556d30aefddc?focus_title=1&muted=1&from_recorder=1)

Also fixed the sorting optimistic update on board view.

### DIFF
--- a/src/app/detail/ui/TaskCardList.tsx
+++ b/src/app/detail/ui/TaskCardList.tsx
@@ -197,14 +197,21 @@ export const TaskCardList = ({
     if (variant === 'task') {
       store.dispatch(updateWorkflowStateIdByTaskId({ taskId: task.id, targetWorkflowStateId: value.id }))
     }
-    if (mode === UserRole.Client && !previewMode) {
-      clientUpdateTask(z.string().parse(token), task.id, value.id, skipSubtaskCascade)
+    const runUpdate = async () => {
+      if (mode === UserRole.Client && !previewMode) {
+        await clientUpdateTask(z.string().parse(token), task.id, value.id, skipSubtaskCascade)
+      } else {
+        await updateTask({
+          token: z.string().parse(token),
+          taskId: task.id,
+          payload: { workflowStateId: value.id, skipSubtaskCascade },
+        })
+      }
+    }
+    if (handleUpdate) {
+      handleUpdate(task.id, { workflowStateId: value.id, workflowState: value }, runUpdate)
     } else {
-      updateTask({
-        token: z.string().parse(token),
-        taskId: task.id,
-        payload: { workflowStateId: value.id, skipSubtaskCascade },
-      })
+      void runUpdate()
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/app/detail/ui/TaskCardList.tsx
```diff
@@ -196,15 +197,24 @@ export const TaskCardList = ({
...
+    if (handleUpdate) {
+      handleUpdate(task.id, { workflowStateId: value.id, workflowState: value }, runUpdate)
     } else {
-      updateTask({
-        token: z.string().parse(token),
-        taskId: task.id,
-        payload: { workflowStateId: value.id, skipSubtaskCascade },
-      })
+      void runUpdate()
     }
```
priosshrsth, line 218:
```suggestion
    if (handleUpdate) {
      handleUpdate(task.id, { workflowStateId: value.id, workflowState: value }, runUpdate)
} else  if(mode === UserRole.Client && !previewMode){
await clientUpdateTask(z.string().parse(token), task.id, value.id, skipSubtaskCascade)
      } else {
        await updateTask({
          token: z.string().parse(token),
          taskId: task.id,
          payload: { workflowStateId: value.id, skipSubtaskCascade },
        })
      }
}
```

The code is wrong. But something like this @arpandhakal ?

src/redux/features/taskBoardSlice.tsx
```diff
@@ -231,7 +231,12 @@ const taskBoardSlice = createSlice({
     },
 
     setAccessibleTasks: (state, action: { payload: TaskResponse[] }) => {
-      state.accessibleTasks = action.payload
+      const workflowStateById = new Map(state.workflowStates.map((s) => [s.id, s]))
+      state.accessibleTasks = action.payload.map((task) => {
+        if (task.workflowState || !task.workflowStateId) return task
```
priosshrsth, line 238: @arpandhakal What does this mean? Could you provide the context behind it?

priosshrsth (changes requested):

src/redux/features/taskBoardSlice.tsx
arpandhakal:
```
    const workflowStateById = new Map(state.workflowStates.map((s) => [s.id, s]))
      state.accessibleTasks = action.payload.map((task) => {
        if (task.workflowState || !task.workflowStateId) return task
        const hydrated = workflowStateById.get(task.workflowStateId)
        return hydrated ? { ...task, workflowState: hydrated } : task
      })
      ```
      
      The `!task.workflowStateId` check is there just to ensure typescript that it exists before searching `task.workflowStateId` in the map. This is just a convention that we follow and has no repercussions whatsoever. If there is already a `task.workflowState` object in a task response, then we completely skip the hydrating process.

priosshrsth: Oh ok. So the idea is to if it has worksflow state or it does not have workflowStateId, then return the original task. I guess that makes sense.

priosshrsth (approved):
@arpandhakal You might want to look into comments from greptile. Looks good to me personally.

PR #1232, opened 2026-05-15
OUT-3741 | TypeError: Cannot read properties of undefined (reading 'map')

arpandhakal:
## Changes

- [x] added task bread crumbs item fallback when taskpath is undefined.


priosshrsth (approved):
lgtm

PR #1239, opened 2026-05-19
OUT-3693 | crm > tasks view is refreshing while users are typing comments

arpandhakal:
## Changes

- [x] In `src/lib/realtime.ts`, add `isInCompanyPreviewScope(task)` mirroring the `isIuCompanyPreview` branch of `getClientOrCompanyAssigneeFilter` (tasksShared.service.ts) — accepts tasks assigned to a client of the preview company, or associated with such a client.
- [x] Apply the carve-out in the four client-role realtime branches:
  - `isSubtaskAccessible`
  - `handleRealtimeTaskInsert`
  - `handleRealtimeTaskUpdate` → `isReassignedOutOfClientScope` (the path that was redirecting the IU back to the board mid-typing)
  - `handleRealtimeTaskUpdate` → `isReassignedIntoClientScope`

### Why this fixes OUT-3693

When an IU opens CRM > company details > Tasks tab, the iframe token carries `internalUserId` + `companyId` (company preview). `RealTime.tsx` derives `userRole = client` whenever `companyId` is present, so the realtime handler runs every check through the client branches. Those branches only accept tasks where `task.clientId === tokenPayload.clientId`, but in company preview `tokenPayload.clientId` is undefined. For any task assigned to a client of the previewed company:

1. INSERTs are silently dropped (symptom: new client-of-company tasks don't show up live).
2. The very next UPDATE on an open task hits `isReassignedOutOfClientScope` (it sees `clientId !== undefined`), triggers `redirectToBoard`, and removes the task from the board — taking any in-progress comment text with it.

The server-side fetch already handles this correctly via the OUT-2898 carve-out in `tasksShared.service.ts` (which is why the tasks come back on refresh). This PR teaches realtime the same carve-out.

## Testing Criteria

- [x] As an IU, open CRM > a company > Tasks tab (company preview mode). Have a second tab open as an IU with full access, and edit a task that is assigned to a client of that company. The preview tab should keep the task open and update its fields live instead of bouncing back to the board.
- [x] Same setup: type a long comment on a task assigned to a client of the previewed company; have the other tab trigger any UPDATE on the task (e.g. change the status). Comment text should be preserved (no redirect).
- [x] In company preview, create a new task assigned to a client of that company from the other tab. It should appear on the preview tab's board in realtime (no refresh needed).
- [x] Regression: as a client user (real CU portal, no preview), confirm that tasks assigned to other clients of the same company still do NOT appear in realtime and the existing redirect-on-out-of-scope behaviour is unchanged.
- [x] Regression: as an IU with no preview (full access or client-access-limited), confirm board/realtime behaviour is unchanged.
- [x] Regression: client preview mode (IU + clientId in token) — confirm task visibility and active-task updates are unchanged.

## Notes

- The fix relies on `assignee` redux state to derive `companyClientIds` for the previewed company. `AssigneeFetcher` already loads the workspace user list for IU previews via `/api/users`, so no extra fetch is needed.
- `RealTime.tsx`'s `userRole = client` for any token with `companyId`/`clientId` is intentional and left untouched per discussion — the carve-out lives entirely inside the handler.

- Touches `src/lib/realtime.ts` only.
- All four call sites are additive boolean clauses; existing client/CU and full-access IU realtime paths are unaffected.
- Worth a regression pass on: CU portal board live updates, IU board (full access + ClientAccessLimited) live updates, and client-preview mode flows in CRM.

### DIFF
--- a/src/lib/realtime.ts
+++ b/src/lib/realtime.ts
@@ -67,6 +67,43 @@ export class RealtimeHandler {
   }
 
   /**
+   * IU-in-company-preview carve-out. Mirrors the `isIuCompanyPreview` branch in
+   * `getClientOrCompanyAssigneeFilter` (tasksShared.service.ts) so realtime treats
+   * "TEAM TASKS" (assigned to a client of the preview company) and team-associated
+   * tasks as in-scope, even though userRole is `client` when companyId is present.
+   */
+  private isInCompanyPreviewScope(newTask: RealTimeTaskResponse): boolean {
+    if (getPreviewMode(this.tokenPayload) !== 'company') return false
+    const companyId = this.tokenPayload.companyId
+    if (!companyId) return false
+
+    const companyClientIds = this.getCompanyClientIdsFromAssignee(companyId)
+
+    if (newTask.companyId === companyId && newTask.clientId && companyClientIds.includes(newTask.clientId)) {
+      return true
+    }
+
+    if (Array.isArray(newTask.associations)) {
+      for (const association of newTask.associations) {
+        if (!association || typeof association !== 'object') continue
+        const assocCompanyId = (association as { companyId?: string }).companyId
+        const assocClientId = (association as { clientId?: string | null }).clientId
+        if (assocCompanyId !== companyId) continue
+        if (assocClientId && companyClientIds.includes(assocClientId)) {
+          return true
+        }
+      }
+    }
+
+    return false
+  }
+
+  private getCompanyClientIdsFromAssignee(companyId: string): string[] {
+    const { assignee } = selectTaskBoard(store.getState())
+    return assignee.filter((u) => u.type === 'clients' && u.companyId === companyId).map((u) => u.id)
+  }
+
+  /**
    * Filters out tasks this user type does not have access to
    */
   private isSubtaskAccessible(newTask: RealTimeTaskResponse): boolean {
@@ -98,6 +135,7 @@ export class RealtimeHandler {
 
       if (
         !this.isAssociatedOrShared(newTask) &&
+        !this.isInCompanyPreviewScope(newTask) &&
         !(
           (newTask.clientId == this.tokenPayload.clientId && newTask.companyId == this.tokenPayload.companyId) ||
           (newTask.clientId == null && newTask.companyId == this.tokenPayload.companyId)
@@ -266,6 +304,7 @@ export class RealtimeHandler {
       // - task's companyId does not match current user's active companyId
       if (
         !this.isAssociatedOrShared(newTask) &&
+        !this.isInCompanyPreviewScope(newTask) &&
         (!newTask.assigneeId ||
           !!newTask.internalUserId ||
           (newTask.clientId && newTask.clientId !== this.tokenPayload.clientId) ||
@@ -319,6 +358,7 @@ export class RealtimeHandler {
     const isReassignedOutOfClientScope =
       this.userRole === AssigneeType.client &&
       !this.isAssociatedOrShared(updatedTask) &&
+      !this.isInCompanyPreviewScope(updatedTask) &&
       (!updatedTask.clientId
         ? updatedTask.companyId !== this.tokenPayload.companyId
         : updatedTask.companyId !== this.tokenPayload.companyId || updatedTask.clientId !== this.tokenPayload.clientId)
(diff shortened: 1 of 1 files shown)


src/lib/realtime.ts
```diff
@@ -66,6 +66,43 @@ export class RealtimeHandler {
...
+   * `getClientOrCompanyAssigneeFilter` (tasksShared.service.ts) so realtime treats
+   * "TEAM TASKS" (assigned to a client of the preview company) and team-associated
+   * tasks as in-scope, even though userRole is `client` when companyId is present.
+   */
+  private isInCompanyPreviewScope(newTask: RealTimeTaskResponse): boolean {
+    if (getPreviewMode(this.tokenPayload) !== 'company') return false
+    const companyId = this.tokenPayload.companyId
+    if (!companyId) return false
+
+    const companyClientIds = this.getCompanyClientIdsFromAssignee(companyId)
```
priosshrsth:
```suggestion
    const companyClientIds = new Set(this.getCompanyClientIdsFromAssignee(companyId))
```

src/lib/realtime.ts
```diff
@@ -66,6 +66,43 @@ export class RealtimeHandler {
...
+      return true
+    }
+
+    if (Array.isArray(newTask.associations)) {
+      for (const association of newTask.associations) {
+        if (!association || typeof association !== 'object') continue
+        const assocCompanyId = (association as { companyId?: string }).companyId
+        const assocClientId = (association as { clientId?: string | null }).clientId
+        if (assocCompanyId !== companyId) continue
+        if (assocClientId && companyClientIds.includes(assocClientId)) {
```
priosshrsth: I would suggest to using set if possible.

src/lib/realtime.ts
```diff
@@ -66,6 +66,43 @@ export class RealtimeHandler {
...
+          return true
+        }
```
priosshrsth: Could we have simplified this? why is associations not typed already?

Maybe return (association as unknown)?.companyId === companyId && companyClientIds.has((association as unknown)?.companyId) is enough. with return `Array.hasSome` approach

src/lib/realtime.ts
```diff
@@ -66,6 +66,43 @@ export class RealtimeHandler {
...
+      }
+    }
+
+    return false
+  }
+
+  private getCompanyClientIdsFromAssignee(companyId: string): string[] {
+    const { assignee } = selectTaskBoard(store.getState())
+    return assignee.filter((u) => u.type === 'clients' && u.companyId === companyId).map((u) => u.id)
```
priosshrsth: I'd suggest `reduce` or `flatMap` for this.

priosshrsth (changes requested):

priosshrsth (dismissed):
lgtm

src/lib/realtime.ts
```diff
@@ -66,6 +66,34 @@ export class RealtimeHandler {
...
+
+    if (newTask.companyId === companyId && newTask.clientId && companyClientIds.has(newTask.clientId)) {
+      return true
+    }
+
+    return !!newTask.associations?.some(
+      (association) =>
+        association?.companyId === companyId && !!association?.clientId && companyClientIds.has(association.clientId),
+    )
+  }
```
priosshrsth, line 90:
```suggestion
    return !!newTask.associations?.some?.(
      (association) =>
        association?.companyId === companyId && !!association?.clientId && companyClientIds.has(association.clientId),
    )
  }
```

Based on the existing codebases I think it's safe to assume that associations might not always be an array. However looks like if association is array. I think association is always defined otherwise `isAssociatedOrShared` would have crashed the app. Can you check? And again if we are so unsure if it cannot always be Asooications[], I suggest updating the type to reflect that.

So, RealTimeTaskResponse.associations should be (unknown | Associtation)[] | undefined.

priosshrsth (changes requested):

src/lib/realtime.ts
priosshrsth: @arpandhakal The reasons is the type suggest associations would be array or undefined. But we seem to be doing Array.isArray(associatons) even after checking that it is defined. So clearly our type is telling a lie.

priosshrsth (approved):
lgtm

PR #1257, opened 2026-05-25
OUT-3737 | Send flow: email-only delivery via NotificationService

arpandhakal:
## Changes

- [x] Add `src/jobs/notifications/send-reminder-email.ts` — `sendReminderEmail()` helper that dispatches a single reminder email via `copilot.createNotification` with `deliveryTargets.email` only.
- [x] Add `src/jobs/notifications/send-reminder-email.test.ts` — covers payload shape, return value, company-recipient header branch, null `recipientCompanyId`, and error propagation.

Implements [OUT-3737](https://linear.app/assemblycom/issue/OUT-3737).

### Why not reuse `NotificationService.create`

That method assumes a `NotificationTaskActions` action and writes to `ClientNotification` — which tracks read-state for in-product notifications that reminders don't create. Reminder dedupe state lives in `TaskReminderSent` (OUT-3734), inserted by the caller (OUT-3730 cron) on success as the idempotency primitive.

### What the helper deliberately does NOT do

- No write to `ClientNotification` / `InternalUserNotification`.
- No write to `TaskReminderSent` (caller's responsibility on success).
- No internal Copilot client init — caller passes in a per-workspace `CopilotAPI` so the cron can amortize one client across many tasks per workspace, per the ticket's "do not rely on request-scoped `this.copilot`".

## Testing Criteria

- [x] `yarn test src/jobs/notifications/send-reminder-email.test.ts` — 5/5 passing.
- [x] `yarn lint:check` — 0 errors on new files.
- [x] `yarn prettier:check` — clean.
- [ ] End-to-end verification deferred to OUT-3730 (cron that calls this) + OUT-3732 (staging dry-run).

## Notes

- Base branch is `feature/email-reminders`, not `main`. Bundles into PR #1256.
- Depends on OUT-3734 (`TaskReminderType` enum), OUT-3735 (`getReminderEmailDetails`) — both already merged into `feature/email-reminders`.
- Unblocks OUT-3730 (Trigger.dev scheduled task) which will import `sendReminderEmail` and own the per-workspace fan-out, company member resolution via `getCompanyClients`, and ledger inserts.
- `senderId = task.createdById` per the ticket. Flag in review if reminders should use a different sender identity (e.g. a system user).

- **Net new code only** — no existing notification flow (assign, comment, reassign, etc.) is touched. `NotificationService.create` is untouched, as is `getEmailDetails`. No DB schema change. No Trigger.dev registration.
- The helper is not wired up anywhere yet — it will only run once OUT-3730 lands and calls it. Until then there is zero production behavior change from this PR.

### DIFF
--- a/src/jobs/notifications/send-reminder-email.test.ts
+++ b/src/jobs/notifications/send-reminder-email.test.ts
@@ -1,0 +1,118 @@
+import { WorkspaceResponse } from '@/types/common'
+import { CopilotAPI } from '@/utils/CopilotAPI'
+import { TaskReminderType } from '@prisma/client'
+import { sendReminderEmail } from './send-reminder-email'
+
+const workspace: WorkspaceResponse = {
+  id: 'ws_1',
+  brandName: 'Acme',
+  labels: {
+    individualTerm: 'client',
+    individualTermPlural: 'clients',
+    groupTerm: 'company',
+    groupTermPlural: 'companies',
+  },
+}
+
+const task = { id: 'task_1', title: 'Submit timesheet', createdById: 'iu_1' }
+
+const buildCopilotMock = (createNotification: jest.Mock) => ({ createNotification }) as unknown as CopilotAPI
+
+describe('sendReminderEmail', () => {
+  it('returns the Copilot notification id', async () => {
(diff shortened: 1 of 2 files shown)


priosshrsth (approved):
lgtm

PR #1259, opened 2026-05-25
allow attachment in public taski api

priosshrsth:
## Summary

Adds a public API endpoint for uploading attachments that can be embedded in tasks created via the public task API.

- `POST /api/attachments/public` — multipart upload. File is uploaded to the workspace-root path (same location as the in-app pre-task-save flow). A `ScrapMedia` row is registered so the file is reaped if the URL is never used. Returns `{ fileName, fileSize, mimeType, downloadUrl }`. **No Attachment row is created here.**
- Caller embeds `downloadUrl` in a task `description` using native Tapwrite markup (`` for images, `<div data-type="attachment">…<attachment-view />` for everything else) — same DOM the in-app editor produces.
- `POST /api/tasks/public` runs the existing task-creation sweep, which reads Supabase metadata, moves the file to the task-scoped path, and **creates the Attachment row** with `taskId` set. This is the same code path the in-app flow has always used.
- The stale workspace-root `ScrapMedia` row is harmless and gets reaped by the existing scrap-media cron on its next pass.

## What changed in shared code

`tasksShared.service.ts` has a single substantive change: `replace` → `replaceAll` in the body URL rewrite. This fixes a pre-existing latent bug — non-image Tapwrite markup embeds the same URL twice (outer `data-src` and inner `<attachment-view src>`), and `replace()` only rewrites the first occurrence, leaving the inner link 404-ing after the file move. Necessary for non-image public uploads (e.g., PDFs); incidental fix for the same case in the in-app flow.

Everything else is in public-only files.

## Testing

- Manual: upload via public endpoint → create task with native markup using the returned `downloadUrl` → confirm the file ends up at the task-scoped path, attachment renders in-app, downloadable URL works.
- Manual cleanup: upload via public endpoint and never reference the URL → confirm the scrap-media cron deletes the file after the expiry window.


arpandhakal (approved):

SandipBajracharya (approved):
lgtm

PR #1261, opened 2026-05-27
OUT-3738 | Observability: structured logging + Sentry for reminder cron

arpandhakal:
## What

Observability for the reminder cron ([OUT-3738](https://linear.app/assemblycom/issue/OUT-3738)). Targets `feature/email-reminders` since the reminder stack isn't on `staging` yet.

- **Wire Sentry into the Trigger.dev runtime** (`src/jobs/sentry.ts`). Jobs run in a standalone Node process, separate from the Next.js server, so `sentry.server.config.ts` (loaded via `instrumentation.ts`) never executes there — without this, `captureException` from a job was a silent no-op. Reuses the already-installed `@sentry/nextjs` (no new deps); one-time init guarded on DSN presence, `defaultIntegrations: false` to stay lean (matches Trigger.dev's documented setup).
- **`send-task-reminders`** — wrap `getEligibleReminders` so an eligibility-query failure is captured (tags `job`, `phase: eligibility`), logged, and rethrown (run still fails visibly). End-of-run log reshaped into one greppable structured summary: `{ eligibleWorkspaces, totalEligibleTasks, enqueued, skipped, runAt }`.
- **`dispatch-reminder-email`** — capture the *terminal* send failure in `onFailure` (after retries exhaust) with tags `taskId`, `recipientId`, `reminderType`, `workspaceId`. ON CONFLICT dedupe skips are never captured.

## Deliberate divergences from the ticket

1. **Summary fields.** The ticket lists `totalEmailsSent/Failed/Attempted`, but this cron fans out — it only *enqueues* `dispatch-reminder-email` jobs and never sends inline, so send outcomes aren't knowable at the orchestrator. They live in the dispatcher's Trigger.dev run metrics + its onFailure Sentry events. The summary reports what the orchestrator actually knows (`enqueued`/`skipped`).
2. **"Mirror auto-archive."** Auto-archive has no Sentry at all, so the Sentry wiring is net-new; only the per-workspace log *format* was mirrored.

## ⚠️ Deploy requirement

For captures to transmit in staging/prod, `SENTRY_DSN` (or `NEXT_PUBLIC_SENTRY_DSN`) must be set on the **Trigger.dev environment** (same place as `COPILOT_ENV`). Without it, `Sentry.init` no-ops and the acceptance check ("force a Copilot 500 → Sentry event") won't fire.

## Testing

- `send-task-reminders.test.ts`: eligibility failure captures + rethrows; ON CONFLICT skip path asserts no capture.
- `dispatch-reminder-email.test.ts`: onFailure capture carries the exact tags; `run()` failure does **not** capture (waits for onFailure).
- All suites pass · `tsc --noEmit` clean · eslint clean.
- Acceptance: forcing a Copilot 500 in staging → tagged Sentry event — pending manual staging verification (needs the DSN env var above).

### DIFF
--- a/src/jobs/notifications/dispatch-reminder-email.test.ts
+++ b/src/jobs/notifications/dispatch-reminder-email.test.ts
@@ -3,6 +3,7 @@ import { TaskReminderType } from '@prisma/client'
 const mockSendReminderEmail = jest.fn()
 const mockTaskReminderSentDelete = jest.fn()
 const mockCopilotApiCtor = jest.fn()
+const mockCaptureException = jest.fn()
 
 jest.mock('@trigger.dev/sdk/v3', () => ({
   task: ({ run }: { run: (payload: unknown) => unknown }) => ({ run }),
@@ -11,6 +12,10 @@ jest.mock('@trigger.dev/sdk/v3', () => ({
 }))
 
 jest.mock('@/config', () => ({ copilotAPIKey: 'test-api-key' }))
+
+jest.mock('@/jobs/sentry', () => ({
+  Sentry: { captureException: (...args: unknown[]) => mockCaptureException(...args) },
+}))
 
 jest.mock('@/lib/db', () => ({
   __esModule: true,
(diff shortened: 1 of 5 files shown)


src/jobs/notifications/dispatch-reminder-email.test.ts
priosshrsth, line 1: imo we don't think we need test for sentry integration.

priosshrsth (approved):
lgtm

PR #1264, opened 2026-05-27
OUT-3731 | Tests: eligibility SQL unit tests + idempotency integration test

arpandhakal:
## What

Real-Postgres tests for the reminder cron ([OUT-3731](https://linear.app/assemblycom/issue/OUT-3731)). **Stacked on `OUT-3738-observability`** (PR #1261) — the idempotency test asserts the Sentry capture added there — so review/merge that first.

The ticket asked for "real DB, no mocks", but its cited reference actually mocks the DB and **no real-DB test infra existed**. So this PR introduces that harness (testcontainers) and the two suites on top of it.

## New: integration test harness
- `jest.integration.config.ts` + `test/integration/*` + a `test:integration` script.
- `globalSetup` boots an ephemeral `postgres:16-alpine`, runs `prisma migrate deploy` (the real migration history, so the schema matches prod), and publishes its URL; `globalTeardown` stops it.
- `setup-env` forces every worker's `DATABASE_URL` onto the container (overriding the dev `.env`), and the test client is **hard-guarded** to refuse any non-local DB — a misconfig can't truncate a real database.
- Kept separate from `yarn test` via `testPathIgnorePatterns`, so the DB-less unit suite still runs in CI. (Note: CI needs a Docker-capable job to run `test:integration` — not added here.)

## Tests
- **`eligibility.integration.test.ts`** (8): runs the actual SQL — all six windows hit on their exact day, ±1-day boundary misses, deleted/archived/completed exclusions, company → single company-level row, and the three subtask carve-outs (+ dead-parent variant). Dates anchored to the DB's `CURRENT_DATE` to avoid UTC-midnight flake.
- **`reminder-idempotency.integration.test.ts`** (3): real DB, Copilot/Trigger/Sentry as doubles (`batchTrigger` fans out to the dispatcher inline so a forced failure exercises the real `onFailure`). One send → one ledger row; immediate re-run → 0 new (unique constraint); forced Copilot failure → ledger row cleared + tagged Sentry event.

## ⚠️ Production bug found & fixed
The real DB surfaced that **ledger compensation never actually deleted**. The global `softDelete`/`softDeleteMany` Prisma extension on `DBClient` rewrites `.delete()/.deleteMany()` into `deletedAt` updates for **every** model, but `TaskReminderSents` has no `deletedAt` column — so:
- `dispatchReminderEmailOnFailure` and the cron's `dispatchChunk` compensation silently failed,
- the `@@unique([taskId, recipientId, reminderType])` constraint would then **permanently block any re-send** of a terminally-failed reminder.

Both paths now hard-delete via `$executeRaw` to bypass the extension; the two unit tests that asserted the old `.delete`/`.deleteMany` calls were updated.

## Testing
- `test:integration` → 11/11 pass · `yarn test` (notifications) → 25/25, integration excluded · `tsc --noEmit` clean · eslint clean.


priosshrsth (approved):
lgtm

PR #1282, opened 2026-06-02
OUT-3750 | SyntaxError: Unexpected token 'A', "An error o"... is not valid JSON

arpandhakal:
## What & why

Fixes [OUT-3750](https://linear.app/assemblycom/issue/OUT-3750) / Sentry **TASKS-88**.

The `(home)` page loaders, the `_fetchers/*` components, and the comment server actions fetch the app's **own** API over HTTP (`apiUrl` resolves to the deployment's own URL) and then call `res.json()` directly.

When the inner serverless function fails at the **Vercel platform level** (timeout / crash / dropped connection — *not* a handled app error, which `withErrorHandler` always returns as JSON), Vercel returns a non-JSON body — `"An error occurred with this application."`, `"Internal Server Error"`, or an empty response. `JSON.parse` then throws `SyntaxError: Unexpected token ...` / `Unexpected end of JSON input`, and with no error boundary the whole render 500s.

The 4 Sentry events confirm this across `GET /`, `GET /detail/...`, and `POST /detail/...`, with three distinct non-JSON bodies.

## Change

Route these raw `res.json()` reads through the existing `fetchWithErrorHandler`, which checks `res.ok`/status **before** parsing and retries transient failures with backoff:

- `(home)/page.tsx` — `getAllWorkflowStates`, `getViewSettings`
- `_fetchers/TemplatesFetcher.tsx`, `WorkflowStateFetcher.tsx`, `AllTasksFetcher.tsx`
- `detail/[task_id]/[user_type]/actions.ts` — `postComment`, `updateComment`

The comment create/update use `retries=0` — they are **non-idempotent** mutations, and retrying a POST/PATCH whose response was lost could duplicate/re-apply the write. They still get the safe parse + clean error, just no retry.

## Scope / follow-ups

- This is the **mitigation** (`fetchWithErrorHandler` + retry), not the root-cause fix. The durable fix is removing the HTTP self-fetch loopback and calling the service layer directly (the detail `loaders.ts` already does this). Tracked separately.
- Three other SSR pages have the same loopback pattern but weren't in the TASKS-88 events and are intentionally left out of this PR: `configure-tasks-app/page.tsx`, `client/page.tsx`, `manage-templates/[template_id]/page.tsx`.

## Testing

- `tsc --noEmit` clean
- `eslint` clean on changed files

### DIFF
--- a/src/app/(home)/page.tsx
+++ b/src/app/(home)/page.tsx
@@ -27,11 +27,10 @@ import { RealTimeTemplates } from '@/hoc/RealtimeTemplates'
 export const maxDuration = 300
 
 export async function getAllWorkflowStates(token: string): Promise<WorkflowStateResponse[]> {
-  const res = await fetch(`${apiUrl}/api/workflow-states?token=${token}`, {
-    next: { tags: ['getAllWorkflowStates'] },
-  })
-
-  const data = await res.json()
+  const data = await fetchWithErrorHandler<{ workflowStates: WorkflowStateResponse[] }>(
+    `${apiUrl}/api/workflow-states?token=${token}`,
+    { next: { tags: ['getAllWorkflowStates'] } },
+  )
   return data.workflowStates
 }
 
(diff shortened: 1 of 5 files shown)


priosshrsth (approved):
lgtm

PR #1292, opened 2026-06-05
OUT-3836: return 404 not-found when a task in a path is absent or inaccessible

SandipBajracharya:
## Changes

- potential fix to prevent app crash when task is not found


priosshrsth:
@SandipBajracharya main is still our main branch. So we should not be merging directly to production branch. You should first merge this to main and then to production.

PR #1297, opened 2026-06-08
use parentId to determine level if in case path is empty

priosshrsth:
## Changes

- [x] If path is empty then use parentId to get level while getting sub tasks count

Tested by using the deployment in task beta and loading the task app and the detail page.

Context:
1. ltree is updated after doing attachment uploads and other promises, Meaning task is created and if attachment takes tim eto upload, someone could go to the detail page and find the path to be null.

## Fixes
1. If path is null we resort back to parentId.

## Caveats:
Note: If path is null then breadcrumb is not displayed now. I think this is fine for now rather than breaking the detail page. 

Screenshot 2026-06-08 at 14 50 15

https://github.com/user-attachments/assets/f5e0e1da-dfeb-4cb3-a4f5-1ea6034f5d21

### DIFF
--- a/src/app/api/tasks/subtasks.service.ts
+++ b/src/app/api/tasks/subtasks.service.ts
@@ -29,10 +29,11 @@ export class SubtaskService extends BaseService {
     const taskId = z.string().uuid().parse(id)
     const data = (
       await this.db.$queryRaw<{ level: number; parentId?: string }[] | null>`
-      SELECT parentId, nlevel("path") as level FROM "Tasks"
+      SELECT "parentId"::uuid, nlevel("path") as level FROM "Tasks"
       WHERE id = ${taskId}::uuid AND "workspaceId" = ${this.user.workspaceId}
     `
     )?.[0]
+
     if (!data) {
       throw new APIError(httpStatus.NOT_FOUND, 'Unable to get the task data while querying subtasks count.')
     }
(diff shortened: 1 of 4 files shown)


src/app/api/tasks/subtasks.service.ts
```diff
@@ -27,16 +27,24 @@ interface Assignable {
...
     }
-    return level
+
+    const { level, parentId } = data
+
+    if (level) {
+      return level
+    }
+
+    return parentId ? 2 : 1
```
SandipBajracharya, line 48: It would be more helpful if we add a comment here describing numbers 2 and 1. It can be confusing for fresh eyes.

SandipBajracharya (dismissed):
LGTM. Please address Greptile comment.

src/app/api/tasks/tasks.service.ts
```diff
@@ -721,6 +721,8 @@ export class TasksService extends TasksSharedService {
 
     const parentIds = getIdsFromLtreePath(taskWithPath.path)
```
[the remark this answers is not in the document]
priosshrsth: @greptileai Already handled this.

src/app/api/tasks/subtasks.service.ts
priosshrsth: @SandipBajracharya Addressed

SandipBajracharya (approved):
LGTM

PR #1303, opened 2026-06-09
OUT-3825 | Grouped content composer (pure function)

arpandhakal:
## What

Adds `composeGroupedEmail(events)` — the render-agnostic step that turns one recipient's buffered window of `GroupedEmailEvent`s into the grouped-email structure. Pure function: no DB, no Copilot, no rendering, no brand prefix. The render adapter (OUT-3826) consumes its output.

Linear: [OUT-3825](https://linear.app/assemblycom/issue/OUT-3825)

> Stacked on #1301 (OUT-3821 schema) — base is `OUT-3821`. Retarget to `feature/grouped-emails` once #1301 merges.

## Shape

```ts
composeGroupedEmail(events) -> {
  totalEventCount: number
  sections: { eventType, count, taskNames: string[], overflowCount: number }[]
}
```

- **Sections** in fixed order `ASSIGNED → SHARED → COMMENT → COMPLETED`; empty sections omitted. (`COMPLETED` stays in the order but is dormant for CU in M2 — it'll light up for IU in M3 with no composer change.)
- **Per section:** events sorted by `createdAt` then title; up to 3 distinct task names listed; `overflowCount = max(0, distinctTasks - 3)` drives the renderer's "+N other tasks".
- **`totalEventCount`** = every buffered event (the N in "You have N new task updates"); section `count`s sum to it.

## Design note: recipient vs. event grouping

The composer operates on **one recipient's window** — recipient-level grouping happens upstream via the `windowKey` claim (OUT-3823/3824). It deliberately never reads recipient IDs; sections-by-type is only the layout *inside* that single per-recipient email.

## Design note: multiple comments on one task

Each comment is a separate buffered event, so it counts individually toward `totalEventCount` and the COMMENT section `count` (e.g. "3 comments were added"), but the task name is listed **once** in the bullets. That's why section `count` (events) can exceed the number of task names shown — intentional, and covered by a test.

## Testing

- 9 unit tests, **100% coverage** — section ordering, empty-section omission, all-empty → `{ totalEventCount: 0, sections: [] }` (caller's skip-send signal), the N count, count-sums-to-N, 3-cap/overflow math, createdAt-then-title sort, and the multi-comment-one-task case.
- `tsc --noEmit` ✅ · eslint ✅ · prettier ✅

### DIFF
--- a/src/app/api/notification/groupedEmail.composer.test.ts
+++ b/src/app/api/notification/groupedEmail.composer.test.ts
@@ -1,0 +1,130 @@
+import { GroupedEmailEventType } from '@prisma/client'
+import { composeGroupedEmail, GroupedEmailEventInput, MAX_TASK_NAMES_PER_SECTION } from './groupedEmail.composer'
+
+let seq = 0
+
+const event = (overrides: Partial<GroupedEmailEventInput> = {}): GroupedEmailEventInput => {
+  seq += 1
+  return {
+    eventType: GroupedEmailEventType.ASSIGNED,
+    taskId: `task_${seq}`,
+    taskTitleSnapshot: `Task ${seq}`,
+    createdAt: new Date(`2026-06-09T10:00:${String(seq).padStart(2, '0')}.000Z`),
+    ...overrides,
+  }
+}
+
+const sectionFor = (events: GroupedEmailEventInput[], type: GroupedEmailEventType) =>
+  composeGroupedEmail(events).sections.find((section) => section.eventType === type)
+
+describe('composeGroupedEmail', () => {
+  it('orders sections ASSIGNED → SHARED → COMMENT → COMPLETED regardless of input order', () => {
+    const events = [
+      event({ eventType: GroupedEmailEventType.COMMENT }),
+      event({ eventType: GroupedEmailEventType.COMPLETED }),
+      event({ eventType: GroupedEmailEventType.ASSIGNED }),
+      event({ eventType: GroupedEmailEventType.SHARED }),
+    ]
+
+    const result = composeGroupedEmail(events)
+
+    expect(result.sections.map((section) => section.eventType)).toEqual([
+      GroupedEmailEventType.ASSIGNED,
+      GroupedEmailEventType.SHARED,
+      GroupedEmailEventType.COMMENT,
+      GroupedEmailEventType.COMPLETED,
+    ])
+  })
+
+  it('omits sections that have no events', () => {
+    const events = [
+      event({ eventType: GroupedEmailEventType.ASSIGNED }),
+      event({ eventType: GroupedEmailEventType.COMMENT }),
+    ]
+
+    expect(composeGroupedEmail(events).sections.map((section) => section.eventType)).toEqual([
+      GroupedEmailEventType.ASSIGNED,
+      GroupedEmailEventType.COMMENT,
+    ])
+  })
+
+  it('returns no sections and a zero count when there are no events', () => {
+    expect(composeGroupedEmail([])).toEqual({ totalEventCount: 0, sections: [] })
+  })
+
+  it('counts every buffered event in totalEventCount', () => {
+    const events = [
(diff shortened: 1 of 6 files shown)


src/app/api/notification/groupedEmail.composer.ts
```diff
@@ -0,0 +1,64 @@
...
+const distinctTaskNames = (events: GroupedEmailEventInput[]): string[] => {
+  const seen = new Set<string>()
+  const names: string[] = []
+  for (const event of events) {
+    if (seen.has(event.taskId)) continue
+    seen.add(event.taskId)
+    names.push(event.taskTitleSnapshot)
+  }
+  return names
+}
```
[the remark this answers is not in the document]
arpandhakal: @greptileai done

src/app/api/notification/groupedEmail.composer.test.ts
```diff
@@ -0,0 +1,130 @@
+import { GroupedEmailEventType } from '@prisma/client'
+import { composeGroupedEmail, GroupedEmailEventInput, MAX_TASK_NAMES_PER_SECTION } from './groupedEmail.composer'
+
+let seq = 0
```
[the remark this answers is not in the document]
arpandhakal: @greptileai done

src/app/api/notification/groupedEmail.composer.test.ts
```diff
@@ -0,0 +1,153 @@
...
```
priosshrsth: I don't think it's a good idea to have it here. Maybe have it inside describe?

src/app/api/notification/groupedEmail.composer.test.ts
```diff
@@ -0,0 +1,153 @@
...
+
+const event = (overrides: Partial<GroupedEmailEventInput> = {}): GroupedEmailEventInput => {
```
priosshrsth: This can go inside descibe too.

src/app/api/notification/groupedEmail.composer.ts
```diff
@@ -0,0 +1,63 @@
+import { GroupedEmailEvent, GroupedEmailEventType } from '@prisma/client'
+
+export const MAX_TASK_NAMES_PER_SECTION = 3
+
+const SECTION_ORDER: GroupedEmailEventType[] = [
```
priosshrsth, line 5: Maybe RHS could be `Object.values(GroupedEmailEventType)`

priosshrsth (changes requested):

src/app/api/notification/groupedEmail.composer.test.ts
arpandhakal: Good call — moved `seq` and its `beforeEach` reset inside the `describe` block (8d6c31b).

src/app/api/notification/groupedEmail.composer.test.ts
arpandhakal: Done — `event` (and `sectionFor`) moved inside `describe` too.

src/app/api/notification/groupedEmail.composer.ts
arpandhakal: I'd lean toward keeping the explicit array here. `Object.values(GroupedEmailEventType)` works today only because the Prisma enum happens to be declared in the same order we want the email sections — but those are two separate concerns. The enum order is a schema detail; `SECTION_ORDER` is a deliberate presentation decision (the PRD's section order). Deriving one from the other means a future enum reorder/addition (e.g. the M3 IU work or a new event type) would silently change the email layout with no failing test.

It also pairs with OUT-3826's renderer, where `sectionHeading` is a `Record<GroupedEmailEventType, …>` — so a new enum value is a compile error there, forcing a conscious decision rather than auto-flowing into the email. The explicit list keeps that same "opt-in deliberately" posture. Happy to revisit if we settle on an enum-derived-ordering convention elsewhere — I didn't spot one in the codebase. WDYT?

priosshrsth (approved):
lgtm

PR #1321, opened 2026-06-15
OUT-3823 | Interception: divert CU email into the grouped buffer

arpandhakal:
## What & why

Wires the **front door** of the grouped-email pipeline (M2). Today every CU task event (Assigned / Shared / Comment) sends one immediate email → spam on bulk-assign. This intercepts those emails centrally in `NotificationService` and writes them to the `GroupedEmailEvent` buffer instead, so the 5-minute dispatcher (OUT-3824) can send **one grouped summary per recipient**. In-product notifications still fire immediately.

Linear: [OUT-3823](https://linear.app/assemblycom/issue/OUT-3823)

## Approach

All CU email paths converge on `NotificationService.create()` and `createBulkNotification()`, so interception is central — not at the ~6 call sites.

- New `groupedEventTypeFor(action)` maps `Assigned*`→`ASSIGNED`, `Shared*`→`SHARED`, `Commented`→`COMMENT` (else `null`).
- New `bufferGroupedEmailEvent(...)`: finds/opens a grouping window, inserts the event (`createMany` + `skipDuplicates` → DB unique index dedups repeat events), and enqueues the 5-min flush timer **only on a new window**.
- `create()` and `createBulkNotification()`: when a target action carries a CU email, buffer it and strip `email` from the dispatch. For comment-to-CU (email-only, in-product disabled) the Copilot call is skipped entirely.

### Window identity = (clientId, companyId)
A client can belong to multiple companies (`companyIds[]`), and the rest of the system keys CU notifications on **(clientId, companyId)**. The window lookup and `windowKey` are scoped to the pair so a multi-company client isn't merged. The dispatcher needs no change — each `windowKey` now contains exactly one pair.

## Deviation from the ticket
Ticket step 1 (`isGroupedEmailEnabled` flag) is **stale** — OUT-3822 was cancelled (no gate). Interception always diverts; isolation is preview env + a test workspace. "Flag OFF: byte-for-byte" → **non-target actions and IU recipients are unchanged.**

## Out of scope
Real-DB idempotency & concurrent window-open races → **OUT-3827** (integration tier). Unit tests cover the logic deterministically.

## Tests
New `notification.service.test.ts` (7 cases): buffer + window open, window reuse, **(client, company) scoping**, non-target byte-for-byte, no-email skip, comment bulk skips dispatch, no-email bulk. `yarn tsc` / lint / prettier clean; full notification + jobs suites green (74).

### DIFF
--- a/src/app/api/notification/notification.service.ts
+++ b/src/app/api/notification/notification.service.ts
@@ -161,6 +161,8 @@ export class NotificationService extends BaseService {
       const clientNotifications = []
       const iuNotifications = []
 
+      const association = AssociationsSchema.parse(task.associations)?.[0]
+
       // NOTE: The reason we are skipping using NotificationService#create and implementing notification dispatch + save manually is because
       // we can just do one `createMany` DB call instead of one per notification, saving a ton of DB calls
       for (let recipientId of recipientIds) {
@@ -180,7 +182,7 @@ export class NotificationService extends BaseService {
             await this.bufferGroupedEmailEvent({
               task,
               recipientClientId: recipientId,
-              recipientCompanyId: task.companyId ?? null,
+              recipientCompanyId: task.companyId ?? association?.companyId ?? null,
               eventType: groupedType,
               commentId: opts?.commentId,
             })
--- a/src/app/api/notification/notification.service.ts
+++ b/src/app/api/notification/notification.service.ts
@@ -13,7 +13,9 @@ import APIError from '@api/core/exceptions/api'
 import { BaseService } from '@api/core/services/base.service'
 import { NotificationTaskActions } from '@api/core/types/tasks'
 import { getEmailDetails, getInProductNotificationDetails } from '@api/notification/notification.helpers'
-import { AssigneeType, ClientNotification, Task } from '@prisma/client'
+import { AssigneeType, ClientNotification, GroupedEmailEventType, Task } from '@prisma/client'
+import { randomUUID } from 'crypto'
+import { enqueueGroupedEmailFlush } from '@/jobs/notifications/flush-grouped-email'
 import Bottleneck from 'bottleneck'
 import httpStatus from 'http-status'
 import { z } from 'zod'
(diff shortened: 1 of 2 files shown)


src/app/api/notification/notification.service.ts
```diff
@@ -159,12 +175,24 @@ export class NotificationService extends BaseService {
             continue
           }
 
+          const groupedType = this.groupedEventTypeFor(action)
+          if (email && groupedType) {
+            await this.bufferGroupedEmailEvent({
+              task,
+              recipientClientId: recipientId,
+              recipientCompanyId: task.companyId ?? null,
```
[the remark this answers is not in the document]
arpandhakal: done. review again @greptileai?

src/app/api/notification/notification.service.ts
```diff
@@ -53,10 +55,24 @@ export class NotificationService extends BaseService {
...
+          recipientClientId: recipientId,
+          recipientCompanyId: task.companyId ?? association?.companyId ?? null,
+          eventType: groupedType,
+          commentId: opts.commentId,
+        })
+        email = undefined
+      }
+      if (!inProduct && !email) return
+
       const notificationDetails = this.buildNotificationDetails(
```
priosshrsth: I think this could be simplified with early returns.

eg. if opts.disabledEmal return.

const email = groupedType && recipientId ?  undefined : getEmailDetails(workspace, actionUser, task, { commentId: opts?.commentId })[action]

etc..

let's rethink about the flow.

priosshrsth (changes requested):

src/app/api/notification/notification.service.ts
arpandhakal: The function cannot get shorter by early returning, because the work isn't just email work. The only simplification is cosmetic.

priosshrsth: I don't undertand. in line 74, we are retruning early. I think that line itself can be moved earlier, can't we?

The whole block of line 63 does not run if email is falsy anyway.

priosshrsth (approved):
lgtm

PR #1335, opened 2026-06-23
OUT-3789 | Filter chip key and value are vertically misaligned

arpandhakal:
## Summary
- Replaced `Typography variant="sm"` with a plain `Box` for the filter value text in `FilterChip.tsx`
- The `Typography` variant was applying its own `line-height` from the design system theme, causing the name text to sit at a different baseline than the sibling "key:" `Box`
- Both text pieces now inherit the same font styles from the parent Stack, aligning them consistently

## Test plan
- [x] Apply a filter (Assignee, Related to, Creator) and confirm the chip key and value are vertically aligned
- [x] Confirm long names are still truncated with ellipsis
- [x] Confirm the chip renders correctly in all three filter types


priosshrsth (approved):
@arpandhakal lgtm. But do ensure that font weight is not being set by parent to be other than normal/400. It is weird to have explicitly set it to 400 unless required.

PR #1347, opened 2026-06-25
OUT-3908 | Unshared client task comments appear in email notifications

arpandhakal:
## Summary
- A task associated with a client (but with `isShared: false`) was incorrectly triggering comment email notifications to that client
- Root cause: `getNotificationParties` for `CommentToCU` was adding associated clients to the recipient list without checking `task.isShared`
- Fix: added `&& task.isShared` guard to the associations block — consistent with the same check in `task-notifications.service.ts` for task-create/update flows

## Test plan
- [x] Add a comment on a task that is associated with a client but **not** shared (`isShared: false`) — client should receive **no** email notification
- [x] Add a comment on a task that is associated with a client **and** shared (`isShared: true`) — client should receive an email notification as before
- [x] Verify grouped email also does not include comments from unshared associated tasks

Fixes OUT-3908


priosshrsth (approved):
lgtm

PR #1353, opened 2026-06-29
OUT-3920 | Add the +N other tasks on the bullet point

arpandhakal:
## Summary
- Moves the `+N other tasks` overflow text from a bare `<em>...<br>` outside the `<ul>` into a `<li>` inside it, for both the normal grouped email and the grouped reminder email
- Both renderers now produce identical HTML structure; the overflow item renders as a bullet point in line with the task list

## Test plan
- [x] Both `groupedEmail.renderer.test.ts` and `groupedReminderEmail.renderer.test.ts` pass (28 tests)
- [x] Verify a grouped email with overflow visually shows `+N other tasks` as a bullet point, not as orphaned italic text below the list


priosshrsth (approved):
lgtm. And I do think this is a better approach than before.

PR #1362, opened 2026-06-30
OUT-3925, OUT-3926 | Extend grouped email system to support IU recipients

arpandhakal:
## Summary

- Wires `recipientIuId` into `bufferGroupedEmailEvent` with a separate window query and IU-specific window key format (`{iuId}:iu:{uuid}`) to avoid collisions with CU keys
- Teaches `flushGroupedEmail` to group and flush IU buffer rows, sending to `recipientInternalUserId` in the Copilot notification payload
- Adds `isIuEmailEnabled()` stub backed by `IU_EMAIL_ALWAYS_ENABLED` env var so IU emails can be tested locally before OUT-3929 (platform preference flag) ships
- Removes the hard `disableEmail` block for IU assignees in `sendUserTaskNotification`, gating it on `isIuEmailEnabled()` instead
- Fixes `buildNotificationDetails` to accept an explicit `isIuRecipient` flag so the payload correctly sets `recipientInternalUserId` (the old inference from absence-of-email breaks once IUs have email)

## Test plan

- [x] Set `IU_EMAIL_ALWAYS_ENABLED=true` in `.env.local`
- [x] Assign a task to an IU — confirm a `GroupedEmailEvents` row is written with `recipientIuId` set and `recipientClientId` null
- [x] Wait 5 minutes (or trigger `flushGroupedEmail` manually) — confirm the email is sent to the IU via `recipientInternalUserId`
- [x] Assign multiple tasks to the same IU within 5 minutes — confirm a single grouped summary email is sent
- [x] Assign a task to a CU — confirm existing CU grouped email behaviour is unchanged
- [x] With `IU_EMAIL_ALWAYS_ENABLED` unset — confirm no IU emails are sent (prod-safe)

### DIFF
--- a/src/app/api/notification/isIuEmailEnabled.ts
+++ b/src/app/api/notification/isIuEmailEnabled.ts
@@ -1,0 +1,3 @@
+import { iuEmailAlwaysEnabled } from '@/config'
+
+export const isIuEmailEnabled = (): boolean => iuEmailAlwaysEnabled
--- a/src/app/api/notification/notification.service.test.ts
+++ b/src/app/api/notification/notification.service.test.ts
@@ -8,6 +8,7 @@ const mockFindFirst = jest.fn()
 const mockFindMany = jest.fn()
 const mockClientNotifCreate = jest.fn()
 const mockClientNotifCreateMany = jest.fn()
+const mockInternalUserNotifCreate = jest.fn()
 const mockQueryRaw = jest.fn()
 const mockGroupedCreateMany = jest.fn()
 
@@ -30,6 +31,7 @@ jest.mock('@/lib/db', () => ({
         createMany: (...args: unknown[]) => mockClientNotifCreateMany(...args),
       },
       groupedEmailEvent: { createMany: (...args: unknown[]) => mockGroupedCreateMany(...args) },
+      internalUserNotification: { create: (...args: unknown[]) => mockInternalUserNotifCreate(...args) },
       $queryRaw: (...args: unknown[]) => mockQueryRaw(...args),
     }),
   },
@@ -90,6 +92,7 @@ beforeEach(() => {
   mockGroupedCreateMany.mockResolvedValue({ count: 1 })
   mockClientNotifCreate.mockResolvedValue({})
   mockClientNotifCreateMany.mockResolvedValue({ count: 1 })
+  mockInternalUserNotifCreate.mockResolvedValue({})
   mockCreateNotification.mockResolvedValue({
     id: 'notif_1',
     createdAt: '2026-06-15T10:00:00.000Z',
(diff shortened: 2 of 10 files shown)


src/app/api/notification/notification.service.test.ts
```diff
@@ -318,6 +318,6 @@ describe('guard: CU wiring boundaries', () => {
   it('never writes recipientIuId in a CU grouped event row', async () => {
     await buildService().create(NotificationTaskActions.Assigned, makeTask())
     const row = mockGroupedCreateMany.mock.calls[0][0].data[0]
-    expect(row.recipientIuId).toBeUndefined()
+    expect(row.recipientIuId).toBeNull()
   })
 })
```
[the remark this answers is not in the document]
arpandhakal: done

src/app/api/notification/notification.service.ts
```diff
@@ -63,28 +67,36 @@ export class NotificationService extends BaseService {
...
 
-      // Non-null only when this CU email should be diverted into the grouped buffer.
       const groupedType = email && recipientId ? this.groupedEventTypeFor(action) : null
       if (groupedType) {
         const association = AssociationsSchema.parse(task.associations)?.[0]
         await this.bufferGroupedEmailEvent({
           task,
-          recipientClientId: recipientId,
-          recipientCompanyId: task.companyId ?? association?.companyId ?? null,
+          recipientClientId: isIuRecipient ? null : recipientId,
```
SandipBajracharya:
```suggestion
          ...(isIuRecipient && {recipientClientId: recipientId}),
          ...(isIuRecipient && {recipientCompanyId: (task.companyId ?? association?.companyId ?? undefined)}),
          ..
```

SandipBajracharya: With this change, then we can remove null type from the `bufferGroupedEmailEvent()` function

src/app/api/notification/notification.service.ts
```diff
@@ -601,35 +611,49 @@ export class NotificationService extends BaseService {
 
   private async bufferGroupedEmailEvent(args: {
     task: Task
-    recipientClientId: string
-    recipientCompanyId: string | null
+    recipientClientId?: string | null
```
SandipBajracharya:
```suggestion
    recipientClientId?: string
    ..
```

src/config/index.ts
```diff
@@ -53,6 +53,9 @@ export const assemblyApiDomain = z.string().url().parse(process.env.NEXT_PUBLIC_
 // Substring stripped from the task title when building the reminder email subject for
 // subject-override workspaces, and the value it's replaced with. Configured via env so the
 // workspace-specific phrasing isn't hardcoded (OUT-3919).
+// Bypasses the platform preference check while OUT-3929 is pending. Set true in dev/staging.
+export const iuEmailAlwaysEnabled = process.env.IU_EMAIL_ALWAYS_ENABLED === 'true'
```
SandipBajracharya, line 57: Is this flag for temporary purpose? Are we going to remove this in the future?

src/app/api/notification/notification.service.ts
arpandhakal: Done — switched to conditional spread and dropped the null type from bufferGroupedEmailEvent().

src/app/api/notification/notification.service.ts
arpandhakal: Done — recipient*Id params are now optional string (no null).

src/config/index.ts
arpandhakal: Yes, temporary — it bypasses the platform preference check while OUT-3929 is pending, and is set true only in dev/staging. It'll be removed once OUT-3929 lands.

src/app/api/notification/isIuEmailEnabled.ts
```diff
@@ -0,0 +1,3 @@
+import { iuEmailAlwaysEnabled } from '@/config'
+
+export const isIuEmailEnabled = (): boolean => iuEmailAlwaysEnabled
```
priosshrsth, line 3: I don't understand the significance of this. Why do we even have this function instead of using the config.

priosshrsth: Ok I think I remember this particular feature being worked on by core assembly team. In that case, should this be a promise and accept workspaceId. This can be done later as well when we actually have an api. Either way it's fine.

src/app/api/notification/notification.service.ts
```diff
@@ -63,28 +67,36 @@ export class NotificationService extends BaseService {
...
       const groupedType = email && recipientId ? this.groupedEventTypeFor(action) : null
       if (groupedType) {
         const association = AssociationsSchema.parse(task.associations)?.[0]
         await this.bufferGroupedEmailEvent({
           task,
-          recipientClientId: recipientId,
-          recipientCompanyId: task.companyId ?? association?.companyId ?? null,
+          ...(!isIuRecipient && { recipientClientId: recipientId }),
+          ...(!isIuRecipient && { recipientCompanyId: task.companyId ?? association?.companyId ?? undefined }),
+          ...(isIuRecipient && { recipientIuId: recipientId }),
```
priosshrsth: This feels unnecessarily complicated.
why not just send `recipientId` and `companyId`. Maybe just add new option `isClient?: boolean`?

Also variable names:
`isIuRecipient => isRecipientIu` : since we are check if recipients iu or cu and not the other way around.

src/app/api/notification/notification.service.ts
```diff
@@ -55,6 +55,10 @@ export class NotificationService extends BaseService {
         action,
       )
 
+      const isIuRecipient =
```
priosshrsth: Does it handle the case when task is handled to CU and shared to IU?

priosshrsth:
```suggestion
      const isAssignedToIu =
```

Just provides the actual insight of the flag. We can look into it and determine if it is true then the task is assigned to IU.

src/app/api/notification/notification.service.ts
```diff
@@ -601,35 +611,49 @@ export class NotificationService extends BaseService {
...
+          LIMIT 1`
+      : await this.db.$queryRaw<{ windowKey: string }[]>`
+          SELECT "windowKey" FROM "GroupedEmailEvents"
+          WHERE "workspaceId" = ${task.workspaceId}
+            AND "recipientClientId" = ${recipientClientId}::uuid
+            AND "recipientCompanyId" IS NOT DISTINCT FROM ${recipientCompanyId}::uuid
+            AND "sentAt" IS NULL
+            AND "createdAt" > now() - interval '5 minutes'
+          ORDER BY "createdAt" DESC
+          LIMIT 1`
```
priosshrsth: Can we DRY this?

priosshrsth (changes requested):

src/app/api/notification/notification.service.ts
arpandhakal: done

src/app/api/notification/notification.service.ts
arpandhakal: done

arpandhakal: Yes, it handles the case when task is handled to IU and shared to CU*

src/app/api/notification/notification.service.ts
arpandhakal: optimized.

priosshrsth (approved):
lgtm

PR #1378, opened 2026-07-08
OUT-3929 | Gate IU notifications via platform notification settings

arpandhakal:
## What & why

Implements [OUT-3929](https://linear.app/assemblycom/issue/OUT-3929). The Assembly platform lets IUs toggle **Product/Email per notification type** on `/settings/notifications` and enforces those prefs **at send time** — a notification is gated per-IU only when it carries a `notificationSettingId`, and a send carries exactly one (one category). This wires Tasks into that mechanism and removes the interim `IU_EMAIL_ALWAYS_ENABLED` env flag.

> Fresh start off `feature/iu-email`; supersedes the earlier single-setting attempt (old PR #1377).

## The core constraint

A **combined grouped email mixes categories**, so it can only be gated all-or-nothing against one setting id — you can't drop just the "assigned" slice for an IU who disabled assigned while keeping "comment". Doing that in one email requires **reading a specific IU's preferences**, and no such API exists in the SDK/docs today (prefs are only applied at send time). So this PR gates what it can now and leaves a seam for later.

## How it works

- **Declared settings → ids.** `CopilotAPI.getNotificationSettings()` resolves the install by `appId` and fetches `installs/{id}/notification-settings`. `resolveIuNotificationSettingId({ category })` maps a category → declared setting id **by label** (per-workspace cache, 5-min TTL, config only). Returns `undefined` when undeclared / on fetch error.
- **In-product:** every IU send carries the category id → platform gates the in-product surface per IU (works today).
- **Email (grouped):** IU emails are always buffered (cross-category windows, unchanged grouping). At flush, the summary carries `notificationSettingId` **only when the whole window is a single category** (then the platform gates it per IU); a single-event replay carries its own id. **Mixed-category summaries send without an id** (not per-IU gated yet — see below).
- **Filter seam:** `filterEventsForIuPreferences` in `flush-grouped-email.ts` is a pass-through (`TODO(OUT-3929)`) where per-IU category filtering will go once Assembly exposes a preference-read endpoint — then a mixed window can be trimmed to allowed categories and an all-disabled recipient skipped.
- **Replies** are now **buffered as COMMENT events** (were direct sends), so they group and gate like top-level comments; the IU in-product still fires immediately with the id.
- **Kill switch removed:** `IU_EMAIL_ALWAYS_ENABLED` is gone; the platform is the gate.

## Category → label mapping (declare these in App Setup, exact strings)

| Category | Label |
|---|---|
| `ASSIGNED` | `New task assigned` |
| `COMMENT` | `New comment on a task` |
| `COMPLETED` | `Task completed` |

## ⚠️ Notes / limitations

- **Blocked dependency:** true per-IU gating of a *mixed-category* combined email needs a per-IU preference **read** endpoint (being confirmed with Assembly). Until then, mixed-category summaries aren't gated per-IU; single-category windows, in-product, and replays are.
- Until a category's setting is declared in the dashboard, its sends carry no id (in-product ungated, email grouped but ungated) — safe, just not yet gated.
- **Behavior change:** CU reply emails are now grouped (up to a 5-min delay) instead of sent immediately — consistent with how top-level CU comments already behave.

## Greptile

- **P1 (ungated direct reply email):** resolved — replies now go through the buffer, no direct-send path.
- **P2 (raw token in log):** dropped `this.token` from the `_getNotificationSettings` log line.

## Testing

- `yarn tsc`, `yarn lint:check`, `yarn prettier` clean.
- Unit: **177 passing** — resolver (label mapping, undeclared, fail-safe, cache), notification.service (always-buffer, id on in-product + buffered email, cross-category windows), flush (single-category window attaches id; mixed window omits it).
- The pre-existing `authenticate.test.ts` p-retry ESM failure is unrelated and untouched.
- Manual (flagged staging workspace, settings declared): assignment/comment/completion + reply to an IU within 5 min → one combined email; toggle a category off → single-category windows for it stop, in-product stops, mixed windows still send (interim), no app-side errors.

### DIFF
--- a/src/app/api/notification/notification.service.ts
+++ b/src/app/api/notification/notification.service.ts
@@ -601,6 +601,7 @@ export class NotificationService extends BaseService {
             )
             .map((iu) => iu.id)
         }
+        break
       default:
         const userInfo = await this.copilot.me()
         senderId = z.string().parse(userInfo?.id)
--- a/src/app/api/notification/notification.service.ts
+++ b/src/app/api/notification/notification.service.ts
@@ -29,12 +29,12 @@ export class NotificationService extends BaseService {
     action: NotificationTaskActions,
     task: Task,
     opts: {
-      disableEmail: boolean
+      disableEmail?: boolean
       disableInProduct?: boolean
       commentId?: string
       senderCompanyId?: string
       emailOverride?: EmailNotificationDetails
-    } = { disableEmail: false },
+    } = {},
   ) {
     try {
       const isAssignedToIu =
@@ -74,7 +74,7 @@ export class NotificationService extends BaseService {
       const email = baseEmail ? mergeEmailOverride({ base: baseEmail, override: opts.emailOverride }) : baseEmail
 
       const category = this.groupedEventTypeFor(action)
-      // Gating disabled until Copilot exposes a per-IU preference read endpoint — ship IUs ungated.
+      // TODO(OUT-3929): re-enable per-IU gating once Copilot exposes a preference-read endpoint — ship IUs ungated for now.
       const notificationSettingId = undefined
       // const notificationSettingId = isRecipientIu && category ? await resolveIuNotificationSettingId({ copilot: this.copilot, workspaceId: task.workspaceId, category }) : undefined
 
@@ -115,12 +115,7 @@ export class NotificationService extends BaseService {
       if (!inProduct && !notificationDetails.deliveryTargets?.email) return
       console.info('NotificationService#create | Creating single notification:', notificationDetails)
 
-      let notification: NotificationCreatedResponse | null
-      try {
-        notification = await this.copilot.createNotification(notificationDetails)
-      } catch (e: unknown) {
-        notification = await this.handleIfSenderCompanyIdError(e, notificationDetails)
-      }
+      const notification = await this.dispatchNotification(notificationDetails)
 
       console.info('NotificationService#create | Created single notification:', notification)
       // Suppressed by the recipient IU's preference — nothing was created, so there's nothing to save.
@@ -206,7 +201,7 @@ export class NotificationService extends BaseService {
       const association = AssociationsSchema.parse(task.associations)?.[0]
       const category = this.groupedEventTypeFor(action)
       const isRecipientIu = opts.isRecipientIu
-      // Gating disabled until Copilot exposes a per-IU preference read endpoint — ship IUs ungated.
+      // TODO(OUT-3929): re-enable per-IU gating once Copilot exposes a preference-read endpoint — ship IUs ungated for now.
       const notificationSettingId = undefined
       // const notificationSettingId = isRecipientIu && category ? await resolveIuNotificationSettingId({ copilot: this.copilot, workspaceId: task.workspaceId, category }) : undefined
       // Non-null only when these emails should be diverted into the grouped buffer.
@@ -261,12 +256,7 @@ export class NotificationService extends BaseService {
           if (groupedType) notificationDetails.deliveryTargets = { inProduct }
 
           console.info('NotificationService#bulkCreate | Creating single notification:', notificationDetails)
-          let notification: NotificationCreatedResponse | null
-          try {
-            notification = await this.copilot.createNotification(notificationDetails)
-          } catch (e: unknown) {
-            notification = await this.handleIfSenderCompanyIdError(e, notificationDetails)
-          }
+          const notification = await this.dispatchNotification(notificationDetails)
 
           console.info('NotificationService#bulkCreate | Created single notification:', notification)
           if (!notification) {
@@ -704,6 +694,16 @@ export class NotificationService extends BaseService {
 
     if (isNewWindow) {
       await enqueueGroupedEmailFlush({ workspaceId: task.workspaceId, windowKey })
+    }
+  }
+
+  private async dispatchNotification(
+    notificationDetails: NotificationRequestBody,
+  ): Promise<NotificationCreatedResponse | null> {
+    try {
+      return await this.copilot.createNotification(notificationDetails)
+    } catch (e: unknown) {
+      return await this.handleIfSenderCompanyIdError(e, notificationDetails)
     }
   }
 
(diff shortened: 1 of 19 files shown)


src/app/api/tasks/task-notifications.service.ts
```diff
@@ -289,13 +288,13 @@ export class TaskNotificationsService extends BaseService {
       // Don't do this in parallel since this can cause rate-limits, each of them has their own bottlenecks for avoiding ratelimits
       shouldCreateNotification &&
         (await this.notificationService.create(NotificationTaskActions.CompletedForCompanyByIU, updatedTask, {
-          disableEmail: !isIuEmailEnabled(),
+          disableEmail: false,
```
priosshrsth: Do we even need this anymore? I mean if we are in the task already, I think we are meant to send the email, aren't we?

src/jobs/notifications/flush-grouped-email.ts
```diff
@@ -87,6 +87,21 @@ const senderFromEvents = (events: WindowEvent[]): EventSender | undefined => {
...
+const filterEventsForIuPreferences = (events: WindowEvent[]): WindowEvent[] => events
+
+// The platform can only gate a send that carries one setting id. A grouped summary gets an id only
+// when every live event shares the same one (i.e. a single-category window); a mixed window sends
+// without an id and is not gated per-IU until the filter seam above is implemented.
+const singleCategorySettingId = (events: WindowEvent[]): string | undefined => {
+  const ids = events.map((e) => e.individualEmail?.notificationSettingId)
+  return ids.every((id) => id && id === ids[0]) ? (ids[0] ?? undefined) : undefined
+}
+
```
priosshrsth, line 104: Do we need any changes in this? My understanding is if email is not to be sent, we won't even add it to our row.

priosshrsth (changes requested):

src/jobs/notifications/flush-grouped-email.ts
arpandhakal: we will be adding to the row, but we will just filter out the rows while compiling all the events into a single email. This is done because it will be one time calling the notification settings api for IU. If we are filtering while adding it to the row, we have to call the api multiple times, for each events.

src/app/api/tasks/task-notifications.service.ts
arpandhakal: yes. that argument is unnecessary. removing.

src/app/api/notification/notification.service.ts
```diff
@@ -102,19 +109,22 @@ export class NotificationService extends BaseService {
...
       if (!inProduct && !notificationDetails.deliveryTargets?.email) return
       console.info('NotificationService#create | Creating single notification:', notificationDetails)
 
-      let notification: NotificationCreatedResponse
+      let notification: NotificationCreatedResponse | null
       try {
         notification = await this.copilot.createNotification(notificationDetails)
       } catch (e: unknown) {
         notification = await this.handleIfSenderCompanyIdError(e, notificationDetails)
       }
```
priosshrsth: I think we can use and intermidate function to replace let with const here and simplify it to one liner here.

src/app/api/notification/notification.service.ts
```diff
@@ -241,11 +256,12 @@ export class NotificationService extends BaseService {
             { inProduct, email },
             opts?.senderCompanyId,
             isRecipientIu,
+            notificationSettingId,
           )
           if (groupedType) notificationDetails.deliveryTargets = { inProduct }
 
           console.info('NotificationService#bulkCreate | Creating single notification:', notificationDetails)
-          let notification: NotificationCreatedResponse
+          let notification: NotificationCreatedResponse | null
```
priosshrsth: Same as above. Try using const instead of let. It will open new and better way to do things for sure.

src/app/api/notification/notification.service.ts
```diff
@@ -194,9 +204,13 @@ export class NotificationService extends BaseService {
       const iuNotifications = []
 
       const association = AssociationsSchema.parse(task.associations)?.[0]
-      // Non-null only when these emails should be diverted into the grouped buffer.
-      const groupedType = email ? this.groupedEventTypeFor(action) : null
+      const category = this.groupedEventTypeFor(action)
       const isRecipientIu = opts.isRecipientIu
+      // Gating disabled until Copilot exposes a per-IU preference read endpoint — ship IUs ungated.
+      const notificationSettingId = undefined
```
priosshrsth: @arpandhakal Can you add TODO comment here insteadl? TODO's will be picked up by IDE so will be helpful when actually implementing the feature.

src/app/api/tasks/task-notifications.service.ts
priosshrsth: @arpandhakal Not seeing the changes.

src/types/common.ts
```diff
@@ -220,6 +223,20 @@ export const NotificationRequestBodySchema = z
...
+export const NotificationSettingSchema = z.object({
+  id: z.string(),
+  label: z.string(),
+  surfaces: z.array(z.enum(['product', 'email'])),
+  default: z.object({ product: z.boolean().optional(), email: z.boolean().optional() }).optional(),
+})
+export type NotificationSetting = z.infer<typeof NotificationSettingSchema>
+
+export const NotificationSettingsResponseSchema = z.object({
+  notifications: z.array(NotificationSettingSchema).default([]),
```
priosshrsth, line 236: @arpandhakal Can you try this in workspace where IU likely haven't touched notification settings. Want to ensure that app will not crash if they return null here. Assembly api seems to be unreliable in this pattern.

priosshrsth (changes requested):

src/types/common.ts
arpandhakal: right now we are not using `resolveIuNotificationSettingId` which uses this schema. Since we are waiting for the assembly team to expose an endpoint for the notification preferences, the function related to this schema is disabled as of now. But, I pretty sure we dont crash the app for worksapces that haven't touched notification settings even when we enable it.

priosshrsth (approved):
lgtm

PR #1399, opened 2026-08-12
OUT-4029 | Scope task/label deletion by id + workspace (fix cross-workspace collateral deletes)

SandipBajracharya:
## What & why

`SubtaskService.softDeleteAllSubtasks` (and `deleteAllAssigneeTasks`) deleted DB rows keyed on the task **`label` string** with **no `workspaceId` filter**. `Labels` has no `workspaceId` column and prefixes derive from brand/company names, so strings like `THE10-001` collide across workspaces — deleting a task in one workspace soft-deleted tasks (and Labels rows) in **other** workspaces.

Prod impact (read-only queries, 2026-08-06): **2,390 tasks soft-deleted as collateral across 263 workspaces**; worst offender `THE10-001` spanned 121 workspaces.

Linear: [OUT-4029](https://linear.app/assemblycom/issue/OUT-4029)

## Changes

**Core fix**
- `softDeleteAllSubtasks`: select `id` (not `label`) from the ltree query and `deleteMany` by `id` + `workspaceId`.
- `deleteAllAssigneeTasks`: drop the unscoped `label.deleteMany`.
- Remove `LabelMappingService.deleteLabel` and all call sites. `Labels` is a next-number registry — orphan rows are harmless, and deleting them rolls the counter back and causes duplicate label reissue. This also makes the OUT-4027 null-guard (#1397) moot.

**Tests**
- Unit: both delete methods delete by id/assignee scoped to `workspaceId` and never touch `Labels`.
- Integration (Testcontainers Postgres): two workspaces sharing `THE10-001`; deleting one workspace's tree soft-deletes only its own subtree and leaves the other untouched. Verified to **fail** against the old string-keyed code and **pass** with the fix.

**Test infra (incidental)**
- `next.config.js`: `transpilePackages: ['p-retry', 'is-network-error']` so `next/jest` can load these pure-ESM deps (fixes a pre-existing `authenticate.test.ts` load failure). Also carries a ngrok dev-origin line.
- `authenticate.test.ts`: corrected a stale public-route `source` assertion (`public` → `platform`) that matched an intentional earlier change (`eb76729f`), previously hidden while the suite couldn't load.

## Not included / follow-ups
- **Restoring the 2,390 collaterally-deleted tasks** — they're soft-deleted and recoverable; needs a `src/cmd/` script + product sign-off (per the ticket's open question).
- The `Labels` table now accumulates orphan rows on delete/reassign (agreed tradeoff) — worth a follow-up if the table grows large.

## Verification
- `yarn tsc` clean, `yarn lint:check` clean.
- Unit tests green; integration test green (and proven to catch the regression).


priosshrsth (commented):
@SandipBajracharya Let's not create label too.

priosshrsth (dismissed):
lgtm

priosshrsth (approved):

PR #1434, opened 2026-08-24
OUT-4093 | Stop reading localStorage in the assignee cache

priosshrsth:
Fixes the Sentry `SecurityError: Failed to read the 'localStorage' property from 'Window'` ([OUT-4093](https://linear.app/assemblycom/issue/OUT-4093/securityerror-failed-to-read-the-localstorage-property-from-window), TASKS-9X) by deleting the code that read `localStorage` rather than guarding it. Net -25 lines in `forageStorage.ts`.

`migrateAssignees()` touched `window.localStorage`, which throws on property access when Chrome blocks third-party storage in our iframe. It shipped in OUT-2348 eleven months ago tagged "remove this after a while" and only moved a cache, so unmigrated users just re-fetch over the network.

What to look at:

- `forageStorage.ts` — the `hasStorageAccess()`/`requestStorageAccess()` gate is gone. It reports access to unpartitioned *cookies*, but the cache is IndexedDB; MDN documents it returning `false` where third-party access is actually fine, which fed into a `requestStorageAccess()` that is auto-denied without a user gesture (there is none in `AssigneeCacheGetter`'s mount effect). It disabled the cache for the users it was meant to protect. The existing try/catch already degrades to a network fetch.
- `forageStorage.test.ts` — stubs `globalThis.window` because `jest-environment-jsdom` isn't installed. The `toHaveBeenCalledWith` assertions are load-bearing: without them the tests pass vacuously via the `typeof window === 'undefined'` early return.

Verified: `yarn tsc` clean; `grep -rn localStorage src/` empty, so no path can still throw this; 3 new tests pass. The 10 failures in `withErrorHandler`/`authenticate` reproduce on a clean `main` and are unrelated.

Not verified: the cache actually populating in a real Chrome iframe with third-party cookies blocked. That upside rests on documentation, not observation — the crash fix does not.

Supersedes #1407, which guarded the migration instead of removing it.

### DIFF
--- a/src/app/_cache/AssigneeCacheGetter.tsx
+++ b/src/app/_cache/AssigneeCacheGetter.tsx
@@ -3,7 +3,7 @@
 import { setAssigneeList } from '@/redux/features/taskBoardSlice'
 import store from '@/redux/store'
 import { useEffect } from 'react'
-import { getAssignees, migrateAssignees } from '@/app/_cache/forageStorage'
+import { getAssignees } from '@/app/_cache/forageStorage'
 
 interface ClientAssigneeCacheGetterProps {
   lookupKey: string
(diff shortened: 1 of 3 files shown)


SandipBajracharya (approved):
LGTM. Just verifying: if we return empty array then the app fetches from the assignees from backend right?

PR #1439, opened 2026-08-26
OUT-4109 | Handle SWR mutate rejections instead of letting them reach Sentry

priosshrsth:
`Error: An error occurred while fetching the data.` is `src/utils/fetcher.ts` throwing inside an SWR `mutate()` whose promise nothing handles, so it lands in Sentry via `onunhandledrejection` rather than the `try/catch` that's already sitting right there. [OUT-4109](https://linear.app/assemblycom/issue/OUT-4109/error-an-error-occurred-while-fetching-the-data) — same root cause as OUT-4107.

Replaces #1437.

What to look at:

- Two rejection paths per component, not one: the optimistic `mutate(...)` (now awaited inside the existing `try/catch`) and the fire-and-forget `debounceMutate(cacheKey)` revalidation (now `.catch`-ed).
- `handleSubTaskCreation` / `handleSubtemplateCreation` had to become `async` for the `await`; both are passed as `(payload) => void` props, which `Promise<void>` satisfies.

Verified: `yarn tsc`, `yarn lint:check` (0 errors), `yarn prettier:check`. Not verified: no manual run against a failing API — this only changes who catches the rejection, not the optimistic rollback behaviour.

https://claude.ai/code/session_01Y4tF84tzJW7Bo1DehsDX22


SandipBajracharya (approved):
lgtm

